PR #11, opened 2022-07-27
Add `--dry-run` and `--replace` flags run command

adambabik:
`--dry-run` prints out what and where is executed:
```sh
$ go run main.go --chdir=./examples run --dry-run echo-hello                                                           
/bin/zsh in ./examples: echo "Hello, rdme!"
```

`--replace` allows to use sed to replace parts of the command:
```sh
$ go run main.go --chdir=examples run echo-hello
Hello, rdme!

$ go run main.go --chdir=examples run --replace="s/rdme/Gopher/" echo-hello
Hello, Gopher!
```

`--replace` uses [sed implementation in Go](https://github.com/rwtodd/Go.Sed).

Both flags can be combined so that if you want to run a snippet that requires some modifications, you can use `--dry-run` to verify that your `--replace` does what you expect.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/cli/cli/v2 v2.14.2
 	github.com/google/go-github/v45 v45.2.0
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
+	github.com/rwtodd/Go.Sed v0.0.0-20210816025313-55464686f9ef
 	github.com/yuin/goldmark v1.4.12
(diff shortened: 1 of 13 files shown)


sourishkrout (approved):
👍 LGTM. What prompted the idea @adambabik?

adambabik:
@sourishkrout I needed to generate a new migration and we have a snippet in readme with hardcoded message. I wanted to customize it directly in rdme. Alternatively I could do `rdme print new-migration | sed ... | eval`.

PR #21, opened 2022-08-26
Add parameter substitution on run 

hamza-tam:
This PR adds the support of parameters substitution when using the run command. 

As an example a command written in the README as: 
```sh {name=echo}
$ echo <YOUR_NAME>
```
Could be called by running 
```sh
$ rdme run echo username
```
The placeholder `YOUR_NAME` would then be replaced by `username` in the final command.

PS: it's my first time working with Golang ^^

### DIFF
--- a/internal/parser/parse_snippets_test.go
+++ b/internal/parser/parse_snippets_test.go
@@ -66,3 +66,20 @@ $ curl -X PATCH -H "Content-Type: application/json" localhost:8080/feedback/a02b
 	p5Snippet, _ := snippets5.Lookup("patch-feedback")
 	assert.Equal(t, []string{`curl -X PATCH -H "Content-Type: application/json" localhost:8080/feedback/a02b6b5f-46c4-40ff-8160-ff7d55b8ca6f/ -d '{"message": "Modified!"}'`}, p5Snippet.Lines())
 }
+
+func TestParameterSubstition_Snippets(t *testing.T) {
+	// Given
+	p6 := New([]byte(`
+Snippet without proper annotations:
+` + "```sh { name=run }" + `
+go run <FILENAME>
+` + "```"))
+	// When
+	snippets6 := p6.Snippets()
+	assert.Len(t, snippets6, 1)
+	snippets6[0].FillInParameters([]string{"main.go"})
+	// Then
+	assert.EqualValues(t, []string{"run"}, snippets6.Names())
+	assert.EqualValues(t, "go run main.go", snippets6[0].Content())
+}
+
--- a/internal/parser/parse_snippets_test.go
+++ b/internal/parser/parse_snippets_test.go
@@ -1,4 +1,3 @@
-// Parser testing methods
 package parser
 
 import (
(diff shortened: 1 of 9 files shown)


adambabik:
Hi @hamza-tam. Thank you for your contribution! The idea looks great, however, rdme already supports that using `--replace` flag. Following your example, it would work like this:

```
rdme run echo --replace="s/<YOUR_NAME>/Test/"
```

`--replace` uses [sed syntax](https://www.digitalocean.com/community/tutorials/the-basics-of-using-the-sed-stream-editor-to-manipulate-text-in-linux#replacing-and-referencing-matched-text). The advantage is that you can use an arbitrary placeholder format.

I also recommend using shell variables as placeholders, for example: 

```
echo $YOUR_NAME
```

Then you can execute it like this:

```
YOUR_NAME=MyName rdme run echo
```

hamza-tam:
Hello @adambabik, Thank you for your feedback !

 The replace command is indeed very handy. In the spirit of the tool, I think adding the feature to input parameters directly too would be very useful. `rdme`'s goal is to remove the need to copy commands manually. While the `--replace` solves the issue of replacing placeholders with actual values, I think it is more suited for edge cases with weird parameter formatting.

Let's take as an example this command from the gh Cli documentation:
```sh
$ gh auth login --hostname <hostname>
```

you could either run this command using `rdme run gh-auth --replace="s/<hostname>/hostname/"` which requires some mouse gymnastics, and kind of removes the need for the tool, Or directly type `rdme run gh-auh hostname`.

Of course other parameter patterns could be added to support more syntaxes.

adambabik:
@hamza-tam You're right about tool's spirit and important aspects of UX and simplicity. Having predictable arguments substitution would be useful indeed as `--replace` is quite complex. 

For the arguments substitution, I would like to see the following:
- If any arguments is missing, the error clearly states it, for instance: `Argument "hostname" is missing`.
- All arguments must be provided; we can skip defaults in any form.
- It should be possible to skip resolving arguments with `--resolve-args=false`.

internal/parser/parse_snippets_test.go
```diff
@@ -66,3 +67,20 @@ $ curl -X PATCH -H "Content-Type: application/json" localhost:8080/feedback/a02b
 	p5Snippet, _ := snippets5.Lookup("patch-feedback")
 	assert.Equal(t, []string{`curl -X PATCH -H "Content-Type: application/json" localhost:8080/feedback/a02b6b5f-46c4-40ff-8160-ff7d55b8ca6f/ -d '{"message": "Modified!"}'`}, p5Snippet.Lines())
 }
+
+func TestParameterSubstition_Snippets(t *testing.T) {
+	// Given
+	p6 := New([]byte(`
+Snippet without proper annotations:
+` + "```sh { name=run }" + `
+go run <FILENAME>
```
adambabik, line 76: We should test also multiple params, missing params, multiline scripts with params.

internal/parser/snippet.go
```diff
@@ -94,3 +99,31 @@ func (s Snippets) Names() (result []string) {
...
+	for _, m := range match {
+		parameters = append(parameters, m[0])
+	}
+	return parameters
+}
+
+func (s *Snippet) mapParameterValues(parameters []string, values []string) string {
+	var paramExp regexp.Regexp
+	newCmd := s.Content()
+	for i := range parameters {
```
adambabik, line 115:
```suggestion
	for i := range parameters {
	    newCmd = strings.ReplaceAll(newCmd, parameters[i], values[i])
	}
```

That should work, I don't see a reason we need to use `regexp` package here.

hamza-tam:
> 

Hello @adambabik, Would it be possible to develop the improvements you asked in a separate PR in order to not make the current one much more complex ? I envisioned that the current one would add the basic feature, and it could be improved upon afterwards.

adambabik:
@hamza-tam totally we can. We are in the middle of some bigger changes for a new use-case but I will try to incorporate your PR this week.

sourishkrout:
Hi @hamza-tam, sorry about the long delay closing the loop here. Runme's development (fka `rdme`) has advanced since. It offers full support of ENV var completion: https://docs.runme.dev/configuration/cell-level#set-environment-variables.

This work both in the CLI & Extension. Here's an example:

image

image

Closing this issue as "won't fix" for now. Feel free to reopen or file another issue if you encounter any other issues. Thank you!

PR #33, opened 2022-10-12
Unify data structures in cli and renderer

adambabik:
Highlights:
* Now the CLI and JSON output use the same data structures.
* Removal of the `description` field from the JSON output. It seems unnecessary as it basically duplicates a previous block. In the CLI `description` refers to the previous block.
* Added the `name` field to the JSON output.

Closes https://github.com/stateful/rdme/issues/31


sourishkrout (approved):
👍  LGTM

Issue #36, opened 2022-10-13
Shell issues with comments and backslashes


sourishkrout:
Won't run because of `#` inside of https://github.com/stateful/runme.dev

```sh
rdme run macos
```

Similar issue where `\` aren't being handled properly in one-liner:

```sh
rdme run deno-install
```

PR #37, opened 2022-10-14
Fix a few shell execution bugs

adambabik:
Highlights:
* Limited the length of a snippet name to 32 characters
* Each line of a snippet goes through a lexer to get rid of comments and handle trailing escape characters

Close https://github.com/stateful/runme/issues/36


sourishkrout (approved):
:+1: Works. LGTM.

PR #52, opened 2022-11-02
Add raw shell executable to keep newlines

lalyos:
runme automatically "minifies" code block, by putting multiple commands into a single line separated by `;` 
I run into 2 issues listed below. Which can be fixed with the introduction of a new Executor called `shell-raw`.

## Multiline for cycle formatted incorrectly

Consider the following example, where its already written in a single line:
```sh {name=1-oneliner}
for i in a b c ; do echo === $i; done
```

now the same for loop but in multiple lines:
```sh {name=2-multiline}
$ for i in a b c ; do
  echo === $i
done
```

the second on will fail, with syntax error:
```
line 1: `set -e -o pipefail;for i in a b c ; do;echo === $i;done;'
```
you can see that the `;` after `do` is not needed. I think you can have the same issue with `if/then`.

## Heredoc

The bash heredoc feature is quite often used in documentations to create a file from bash.
Something like this:
```sh-raw {name=4-heredoc}
cat > Dockerfile <<EOF
FROM alpine
RUN apk add curl
EOF
```

### DIFF
--- a/internal/runner/executable.go
+++ b/internal/runner/executable.go
@@ -27,6 +27,11 @@ func New(block *document.CodeBlock, base *Base) (Executable, error) {
 			Cmds: block.Lines(),
 			Base: base,
 		}, nil
+	case "sh-raw":
+		return &ShellRaw{
+			Cmds: block.Lines(),
+			Base: base,
+		}, nil
 	case "go":
 		return &Go{
 			Source: block.Content(),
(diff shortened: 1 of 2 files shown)


adambabik:
@lalyos thank you for the contribution! The problem statement is very clear, appreciate that.

Only one request, please add `sh-raw` to the [`supportedExecutables`](https://github.com/stateful/runme/pull/52/files#diff-9a3d7d29312dfa69c4c3904e0d8f16616cb64d9ebb1b922fc3405f6667c47d7bR45) map and I will merge it.

lalyos:
@adambabik just fixed your request.

adambabik:
@lalyos thank you!

Issue #49, opened 2022-10-27
Problem parsing with lists containing code blocks


christian-bromann:
The following example of markdown is not rendered correctly:

    ## Running
    
    ### In Minikube
    
    Deploy the application to Minikube using the Linkerd2 service mesh.
    
    1. Install the `linkerd` CLI
    
        ```bash
        curl https://run.linkerd.io/install | sh
        ```
    
    1. Install Linkerd2
    
        ```bash
        linkerd install | kubectl apply -f -
        ```

    1. View the dashboard!
    
        ```bash
        linkerd dashboard
        ```
    
    1. Inject, Deploy, and Enjoy
    
        ```bash
        kubectl kustomize kustomize/deployment | \
            linkerd inject - | \
            kubectl apply -f -
        ```
    
    1. Use the app!
    
        ```bash
        minikube -n emojivoto service web-svc
        ```

It should be rendered like this:

### In Minikube

Deploy the application to Minikube using the Linkerd2 service mesh.

1. Install the `linkerd` CLI

    ```bash
    curl https://run.linkerd.io/install | sh
    ```

1. Install Linkerd2

    ```bash
    linkerd install | kubectl apply -f -
    ```

1. View the dashboard!

    ```bash
    linkerd dashboard
    ```

1. Inject, Deploy, and Enjoy

    ```bash
    kubectl kustomize kustomize/deployment | \
        linkerd inject - | \
        kubectl apply -f -
    ```

1. Use the app!

    ```bash
    minikube -n emojivoto service web-svc
    ```

But gets rendered like this:

<img width="486" alt="Screenshot 2022-10-27 at 10 38 34" src="https://user-images.githubusercontent.com/731337/198360319-b03c12ca-3a65-42c8-8b49-876c22828514.png">

adambabik:
@christian-bromann this is very interesting case. Another one, less popular, is a code block within a block quote.

Have you seen https://github.com/microsoft/vscode-markdown-notebook? It seems to be doing exactly what we need and the output looks almost right. The order is not preserved, all list items starts with `1.` and it does not handle the case with a block quote. The algorithm is pretty trivial though.

I am working on something a bit more advanced which will move code blocks up to the top level under the root node while preserve the affected blocks.

adambabik:
[...]

Issue #50, opened 2022-10-29
Missing allow-unknown option in list command


lalyos:
The example readme mentions a `--allow-unknown` option, but the latest release (0.2.4)
doesn't have it. Even couldn't find on master branch.

The related commit:
https://github.com/stateful/runme/commit/59c2c9a83058eefaead628b898a492bc2ce16562#r88258426

lalyos:
btw I love this project, actually I was prototyping on a very similar thing :D

adambabik:
Hi @lalyos! Thank you for the kind words, we are glad you find it useful :) There is much more to come so stay tuned!

Regarding the flag, you're correct. I forgot to include the commit in the previous PR. It will be added very soon.

adambabik:
It was released in https://github.com/stateful/runme/releases/tag/v0.3.0.

PR #53, opened 2022-11-02
Refactor parsing and expose new WASM interface

adambabik:
Highlights:
* Parsing is simplified and refactored to handle edge cases more easily like nested fenced code blocks.
* Rendering process exposes each block separately so that when using the Notebook interface, we have blocks corresponding to cells.

To test the new WASM interface, run:

```
make wasm
cd examples/web
python -m http.server 9000
```

And open the address in a browser. You should see the new output. Check out `examples/web/index.html` to learn more about the new global functions.

## Todo

- [x] Go through these MDs:
   * [DevOps Handbook](https://github.com/freeCodeCamp/freeCodeCamp/blob/084bf7600f2f8fcdcb2bb82abe2f842d457b270a/docs/devops.md)
- [x] Implement the case when editing a block results in its split. For example, a block containing a single paragraph might be edited and converted into two paragraphs which result in two blocks each containing a single paragraph.

Closes https://github.com/stateful/runme/issues/49 https://github.com/stateful/runme/issues/50


christian-bromann:
Another markdown example which we can use to test this fix: https://github.com/vercel/next.js/blob/canary/contributing/core/developing.md

adambabik:
@christian-bromann thanks for this example. It helped me to find a bug which was fixed in https://github.com/stateful/runme/pull/53/commits/af65f91952e567209e836202f399f3998684ebfc.

sourishkrout (changes requested):
1. Consider adding the streamlined WASM interface.
2. Marshaling runs into deadlock with this file https://gist.github.com/sourishkrout/4f6bc408bd2287099fac39e2b99a5f9a - circular reference?

@adambabik please lemme know if you have questions

internal/document/block.go
```diff
@@ -205,67 +249,210 @@ func (b *CodeBlock) Name() string {
...
+				start++
+			}
+			for ch := b.source[start]; ch == '\n'; ch = b.source[start] {
+				start++
+			}
+
+			nextBlock := b.inner.NextSibling()
+			for nextBlock != nil && !hasLine(nextBlock) {
+				nextBlock.NextSibling()
+			}
```
sourishkrout, line 388: Never mind the ciruclar ref comment... this is where the infinite loop happens @adambabik

internal/document/block.go
```diff
@@ -205,67 +249,210 @@ func (b *CodeBlock) Name() string {
...
+				_, start = getContentRange(b.source, previousBlock)
+				start++
+			}
+			for ch := b.source[start]; ch == '\n'; ch = b.source[start] {
+				start++
+			}
+
+			nextBlock := b.inner.NextSibling()
+			for nextBlock != nil && !hasLine(nextBlock) {
+				nextBlock.NextSibling()
```
sourishkrout, line 387:
```suggestion
				nextBlock = nextBlock.NextSibling()
```
i believe this will fix it

internal/document/block.go
```diff
@@ -205,67 +249,210 @@ func (b *CodeBlock) Name() string {
...
+				stop = stop + idx
+			}
+
+			b.content = string(b.source[start:stop])
+			b.start = start
+			b.stop = stop
+		case ast.KindThematicBreak:
+			previousBlock := b.inner.PreviousSibling()
+			for previousBlock != nil && !hasLine(previousBlock) {
+				previousBlock.PreviousSibling()
```
sourishkrout, line 373:
```suggestion
				previousBlock = previousBlock.PreviousSibling()
```
and here

web/main_js.go
```diff
@@ -18,51 +20,93 @@ var (
...
+	result := make(map[string]interface{})
+	_ = json.Unmarshal(b.Bytes(), &result)
+	return result
+}
+
+func getSource(this js.Value, args []js.Value) any {
+	if parsed == nil {
+		return errors.New("call initialize() first")
+	}
+	return parsed.Source()
```
sourishkrout, line 87: if you try calling this from the browser @adambabik, it'll throw an error. Suspecting it does know how what to do what to do with a byte array.

web/main_js.go
```diff
@@ -18,51 +20,93 @@ var (
...
+	if parsed == nil {
+		return errors.New("call initialize() first")
+	}
+
+	idx := args[0].Int()
+	newSource := args[1].String()
+
+	updater := document.NewUpdater(parsed)
+	if err := updater.UpdateBlock(idx, newSource); err != nil {
+		return err
```
sourishkrout, line 100: same with @adambabik, it appears it can't handle the Golang Error type

sourishkrout:
question for you @adambabik: looks like `source: '```\nnode ./scripts/stdin.js\n```'` always includes the triple backticks of fenced code blocks now. will this be universally true for `type=code`?

adambabik:
> question for you @adambabik: looks like `` source: '```\nnode ./scripts/stdin.js\n```' `` always includes the opening/closing triple backticks of fenced code blocks now. will this be universally true for `type=code`?

Not sure what you mean by `type=code`. If you mean also code blocks (not fenced) then no. These are called fence marks and only exists for the fenced code blocks and we only recognize those as executable.

I can get rid of the fence marks if that's problematic. VS Code interface for code cells probably accepts raw code without them. Using `lines` would be worse because they strip out all prefix and trailing white spaces and we want to keep them - this is an existing flaw where code blocks look like a wall of text without indentations.

web/main_js.go
adambabik: Good catch, I didn't verify in the `example.html` this function. It's fixed and it works now in the browser.

sourishkrout:
> Not sure what you mean by `type=code`. If you mean also code blocks (not fenced) then no. These are called fence marks and only exists for the fenced code blocks and we only recognize those as executable.
> 
> I can get rid of the fence marks if that's problematic. VS Code interface for code cells probably accepts raw code without them. Using `lines` would be worse because they strip out all prefix and trailing white spaces and we want to keep them - this is an existing flaw where code blocks look like a wall of text without indentations.

Actually for visualization keeping the backticks is better, since when we have unsupported languages (non-shell-based) we do want to render them as the original markdown.

I think it makes sense to stay as close to the source of truth as possible. They only place it came up was with the language detection model. If I leave the backticks in everything will be guessed as markdown. I just created a purpose specific normalization function.

sourishkrout:
Integration available here: https://github.com/stateful/vscode-runme/pull/52

examples/web/index.html
```diff
@@ -18,11 +18,28 @@
...
       const go = new Go();
-      WebAssembly.instantiateStreaming(fetch("runme.wasm"), go.importObject).then((runme) => {
-        go.run(runme.instance);
-        console.log(GetSnippets(markdown));
-        console.log(GetDocument(markdown));
-        console.log(PrepareScript(["bash -c 'echo \"1\"'"]));
+      WebAssembly.instantiateStreaming(fetch("runme.wasm"), go.importObject).then((obj) => {
+        go.run(obj.instance);
+
+        console.log("return error", Runme.getSource());
```
sourishkrout, line 24: Technically it's odd to "return" errors in JS, but let's not get hung up on it. Lets plan to iterate on errors as part of building out the editing stuff.

sourishkrout (approved):
Ready to merge @adambabik?

examples/web/index.html
adambabik: There is no other way AFAIK. I could return a string but I guess it makes sense - an exported JS function returns a JS `Error` instance. Or I could return `null` :) We will get back to it.

PR #56, opened 2022-11-10
WIP: Add shell command

adambabik:
`runme shell` allows to open a new shell session which inherits all (most?) of the properties of the current and user's shell. For example, it inherits environment variables and it uses user's defined shell using the `SHELL` env.

The main purpose of running `runme shell` is to open a stateful shell session in which `runme run` commands can be executed and their side-effects are preserved. For example, executing a command that exports shell variables within a runme shell should preserve them:

````
➜ ./runme shell
detected prompt: ➜
runme: welcome to runme shell
runme: artifacts will be stored in /var/folders/g9/rrx_zc9n6xq6wf6y28n92r3w0000gn/T/runme-1001-1812127229


➜ ./runme ls --filename test-shell.md
NAME  FIRST COMMAND      # OF COMMANDS  DESCRIPTION
set   export A="Test A"  1              First, set an environment variable.
get   echo $A            1              Then read it.

➜ ./runme run --filename test-shell.md set

# this line is inserted automatically
➜ export A="Test A"

➜ ./runme run --filename test-shell.md get

# this line is inserted automatically
➜ echo $A
Test A
````

## Todo

- Consider changing the UX. `./runme run --filename test-shell.md get` should not print `echo $A`, even though, it's automatically. The better UX could be to just output the result directly. Note that this won't be a problem when a trigger will be from outside, for example, when it will be VS Code.


adambabik:
It will be replaced with https://github.com/stateful/runme/pull/93.

Issue #58, opened 2022-11-16
Run CLI against a node_modules dependency-tree (e.g. vue-js)


sourishkrout:
- [ ] Build a workflow to test against dep tree's README.md's

adambabik:
I recently came across an interesting post and tool to test CLIs: https://bitfieldconsulting.com/golang/cli-testing.

sourishkrout:
> I recently came across an interesting post and tool to test CLIs: https://bitfieldconsulting.com/golang/cli-testing.

Sounds good! Can you track a new issue for this? I don't see those test scenarios as competing. I think we want both.

PR #60, opened 2022-11-16
Readme robustness test workflow

sourishkrout:
The test subject right now is a scaffolding of a Vue app, we're really only interested in the dependencies tree.

Closes https://github.com/stateful/runme/issues/58.


sourishkrout:
Creating PR so I can test the workflow in Github Actions

sourishkrout:
Added quite a few more dependencies, not just vuejs, to broaden exposure.

Makefile
```diff
@@ -28,6 +28,11 @@ test:
 test/update-snapshots:
 	@TZ=UTC UPDATE_SNAPSHOTS=true go test ./...
 
+.PHONY: test/robustness
+test/robustness:
+	@cd integration/subject && npm install --include=dev
+	@RESULTS=`find . -name "README.md" | grep -v "\/\." | xargs dirname | uniq | xargs -n1 ./runme json --chdir`
```
adambabik, line 34: Or this to be more explicit with the intention:

```suggestion
	find . -name "README.md" | grep -v "\/\." | xargs dirname | uniq | xargs -n1 ./runme json --chdir &>/dev/null
```

adambabik (approved):

Makefile
sourishkrout: Good call. Also wouldn't want whatever buffer's behind `RESULTS` to run into an overflow (if that's possible) 🤷 .

PR #74, opened 2022-12-08
internal/document: improve attrs processing

adambabik:
Highlights:
- Remove serializing the `name` attribute if it was not present in the source.
- Add spaces between `{ ... }`.

### DIFF
--- a/internal/document/block.go
+++ b/internal/document/block.go
@@ -60,10 +60,8 @@ func newCodeBlock(
 	source []byte,
 	render Renderer,
 ) (*CodeBlock, error) {
-	name := getName(node, source, nameResolver)
-
 	attributes := getAttributes(node, source)
-	attributes["name"] = name
+	name := getName(node, source, nameResolver)
 
 	value, err := render(node, source)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


sourishkrout:
I might need to introduce an identifier on the extension side that allows me to identify a cell (might just use xid). What do you think about excluding metadata keys starting with `_` (eg `_id`) from serialization? Simplifies managing transient metadata

adambabik:
@sourishkrout of course, we can exclude internal/private attributes. I was thinking either of `_` or `runme.dev/attribute-name`.

sourishkrout:
> @sourishkrout of course, we can exclude internal/private attributes. I was thinking either of `_` or `runme.dev/*`.
> 
> Let me know which way you want to go.

What do you think is cleaner? I'd like to avoid the metadata schema growing like weeds.

adambabik:
@sourishkrout if `_` is unused by VS Code, we can use `_` as it's shorter. If `_` is already used and we want to have more control then `runme.dev/*`.

sourishkrout:
> @sourishkrout if `_` is unused by VS Code, we can use `_` as it's shorter. If `_` is already used and we want to have more control then `runme.dev/*`.

Worked through the requirement for a ephemeral metadata attributes and turns out I can get around it using the native vscode API. However, I think we should probably jut do both `_` and `runme.dev/*` and choose depending on use-case which one is more appropriate.

Also, please go ahead and remove `index` entirely from the metadata. It's no longer required since the extension is relying exclusively on the API now.

sourishkrout (changes requested):
Please see previous comment. Otherwise, ready to merge 👍 .

Issue #75, opened 2022-12-08
runme fmt does not mirror default behavior of other commands


sourishkrout:
It's a bit awkward that `runme fmt` requires the filename to be passed as mandatory argument.

Why can't it work like `runme ls` where it defaults to `--filename README.md`.

Consistency is important.

PR #76, opened 2022-12-09
Unify reading and writing source file

adambabik:
`runme fmt` is exposed now and works like other commands when executed without params.

Closes https://github.com/stateful/runme/issues/75


sourishkrout (approved):
LGTM 👍

Issue #89, opened 2022-12-16
Frontmatter is not being preserved


sourishkrout:
Popular tooling like docusaurus, Jekyll, and Hugo use frontmatter to store metadata. Runme's parser should at the minimum make sure to preserve it.

Related to https://github.com/stateful/runme/issues/88

sourishkrout:
Let's take a look into this soon please @adambabik. I tested with files here: https://github.com/stateful/runme.dev/blob/main/docs/intro.md and it appears we're treating frontmatter like regular markdown.

adambabik:
I did an initial research and Hugo uses the same MD parser as we do but it does not use it for Front Matter. It uses a custom pre-parser which recognizes various formats. It's [documented](https://gohugo.io/content-management/front-matter/).

I will use the same technique.

sourishkrout:
> I did an initial research and Hugo uses the same MD parser as we do but it does not use it for Front Matter. It uses a custom pre-parser which recognizes various formats. It's [documented](https://gohugo.io/content-management/front-matter/).
> 
> I will use the same technique.

Sounds good 👍 . We might want to consider retaining the frontmatter (blob) in a single metadata field across short-term just to make sure it's not being wiped if that's a quicker stop-gap fix. Longer-term it'd be great to have the frontmatter parsed and serializable into its originate format (yaml, json, toml, etc).

PR #92, opened 2022-12-22
document: support front matter

adambabik:
This change adds an ability to retain a front matter section in a document.

The front matter won't be available to edit. It will be stored in Notebook's metadata.

Supported formats are: TOML, YAML, and JSON.

Closes https://github.com/stateful/runme/issues/89


sourishkrout (approved):
LGTM 👍

PR #97, opened 2023-01-10
examples: add demo with gRPC client

adambabik:
This change contains an example of a client written in TypeScript that talks to runme's server through [Connect protocol](https://connect.build/docs/introduction). This is a web-friendly gRPC compatible alternative with great libraries for Go and TS. Note that connect-node which will allow to write Connect servers is still [in progress](https://github.com/bufbuild/connect-web/discussions/315).

In order to test it, you must start `runme server` with `--web` flag.

Then, you can cd to `examples/client-ts`, `npm install`, and `npm start`.

One more highlight is the fact that the client uses code generated by [Buf Remote Packages](https://docs.buf.build/bsr/remote-packages/overview).


sourishkrout:
Got it running.

1. I didn't realize that I needed to to `--address 0.0.0.0:8080` since the test UI is accepting it.
2. My `bash` is not located in the assumed directory. Not super critical but we should probably detect the right location.

adambabik:
> 1. I didn't realize that I needed to to `--address 0.0.0.0:8080` since the test UI is accepting it.

I will fix this so that it defaults to `localhost:8080` when using `--web` and not passing `--address/-a`.

> 2. My `bash` is not located in the assumed directory. Not super critical but we should probably detect the right location.

That's a bit more tricky. We mostly use macOS for development and default bash is old (3.2) and prompt auto-detection does not work with it. I used a newer bash. It is explained in `index.ts` in `createSession()`. But I think for the time being, it might be better to assume the old bash. I will adjust that as well.

PR #101, opened 2023-01-13
kernel: use expect to detect prompt

adambabik:
Highlights:
- switch to use [goexpect](https://github.com/google/goexpect) instead of custom logic
- add more examples with [xterm.js](http://xtermjs.org/); `client-raw-ts` is close to raw asynchronous bi-directional communication between a terminal emulator and a pseudoterminal

In order to run the examples, first compile the new binary with `make` and then start the server using `./runme server --web`. 

Next, you can change directory to `examples/client-ts` and inside run `npm start`. After successful compilation, there should be a link to open in a browser.


examples/client-raw-ts/src/index.ts
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        term.write(data)
+    }
+
+    console.log("finished reading output")
+}
+
+async function createSession() {
+    // Prompt can be auto-detected but it will work only in bash 4.4+.
```
sourishkrout, line 95: So if one doesn't have a bash 4.4+ installed will it not work @adambabik? is this something we should check for?

sourishkrout: image

sourishkrout:
Got this up and running locally. Had to install `GNU bash, version 5.2.15(1)-release` via hombrew.

I noticed that if I don't use the chunked option the UI won't show intermediate results for a sequence of commands. That's expected right?

examples/client-raw-ts/src/index.ts
adambabik: The command to detect a prompt in lower bash versions won't work. 

I want to implement using a Docker container to have these examples and tests working reliably regardless where they are run.

adambabik:
> I noticed that if I don't use the chunked option the UI won't show intermediate results for a sequence of commands. That's expected right?

That is correct. I am not sure if this is going to be useful, but I will keep it for now. It's the simples and most reliable way to run commands. As discussed separately, a bi-directional streaming RPC will be added soon as well.

adambabik:
@sourishkrout I'd like to merge it. Let me know if there are any blockers. The rest of issues we identified with the kernel will be done in a separate PRs.

sourishkrout:
> @sourishkrout I'd like to merge it. Let me know if there are any blockers. The rest of issues we identified with the kernel will be done in a separate PRs.

Sounds good. I'll review the code one more time and will approve the PR accordingly before EoD.

examples/client-raw-ts/src/index.ts
sourishkrout: Gotcha. Just to clarify for the kernel services to work for a user, one would have to have a more recent (4.4+) version of bash installed? I'm mainly asking because it's something we might want to prompt for in the UI if the min requirements are not met. Lemme know @adambabik. Thanks!

sourishkrout (approved):
Ready to rock :+1:. Good work!

examples/client-raw-ts/src/index.ts
adambabik: It's a bigger discussion in my opinion. For the testing purposes, I default to bash and a version 4.4+ is required to use a prompt auto-detection, however, you can specify the prompt manually.

Overall, there are a few factors we will need to take a look in the extension to decide how to start a session:
- Selecting a shell. One usually has many shells installed and a preferred one is in `/etc/passwd`. Alternatively, there might be the `SHELL` env defined.
- Deciding on the prompt. If we use user's shell, it's likely that the prompt is customized and if it's something like [spaceship-prompt](https://github.com/spaceship-prompt/spaceship-prompt) we may have hard time to work with it. I have a very rich prompt at the moment and detection of it is flaky with the current code.

Things we may consider:
- Using a known shell via Docker. I plan to add such an option for testing purposes. This obviously is the most stable and predictable solution. 
- Using a system shell (either a static like `bash` or user's current one) but setting our own reliable prompt. Going further, we may provide our own `.bashrc` (or `.zshrc` and so on) to make it even more reliable.
- Defaulting to user's shell and user's prompt and do everything possible to handle it.

Issue #100, opened 2023-01-13
Add interactive runme TUX mode


sourishkrout:
It would be awesome if when you run the bare "runme" command it would drop you into an interactive list with each command section and you could run it by hitting enter and then use the arrows keys to switch to the next section.

So the CLI interactions would be:

 1. runme
 2. <enter> to run first section
 3. <down-arrow> to select next item
 4. <enter> to run the second section

sourishkrout:
Implemented by https://github.com/stateful/runme/pull/103

sourishkrout:
Just shipped with v0.5.0. Cc @silas, sorry about the delay - we had a company offsite in the middle of it.

silas:
Awesome, just updated, thanks everyone!

PR #103, opened 2023-01-18
Interactive TUX

mxsdev:
Solves #100 

<p align="center">

</p>


mxsdev:
@sourishkrout Oh yeah, sorry, I misunderstood the original ticket. Fixed in [4c33c94](https://github.com/stateful/runme/pull/103/commits/4c33c94bea59a33afd1021438b938bbf0974f222) and I added an `--exit` flag for the case where you want to run a command one-off.

sourishkrout:
Hey @silas, you have a sec to check out whether or not the UX meets your expectations?

Outside of that, @mxsdev, one thing I noticed is that the list entries can exceed your terminal's height with a fully loaded README.md such as https://github.com/stateful/vscode-runme/blob/main/examples/README.md. We could track this issue in a separate ticket, though.

silas:
I check tried the branch locally and it works great!

The one comment I'd add is that if the section fails it drops you out of the tui and prints a kinda ugly error message.

```sh
runme

> echo-ok sh
    echo ok
  echo-fail
    echo fail (...)

ok
runme

  echo-ok
    echo ok
> echo-fail sh
    echo fail (...)

fail
failed to run command "set -e -o pipefail;echo fail;false;\n": exit status
```

For the following readme

~~~markdown
```sh
echo ok
```

```sh
echo fail
false
```
~~~

Might be worth just printing an error message with the exist status code and sending you back into the selection menu.

Might also be nice to add a newline after the run, just to create some space between the run output and the `runme` title.

mxsdev:
Added scrolling (number of shown entries configurable with `--entries) and recovery from error. Also made the spacing nicer between the command's stdout and the runme tui. (and I fixed the linting errors...[for real this time](https://github.com/mxsdev/runme/actions/runs/3960604421) 😅 )

<p align="center">

</p>

I didn't give any scroll indicators, so the ui is a little minimal in this respect. Definitely let me know if you think we should add more to the ui.

@silas The ugly error message should be tracked in a separate issue, since this is the same behavior as in executing `runme run <cmd>`.

silas:
@mxsdev looks great! Also nice spacing change, love the polish.

I created issue #106 about improving the error message.

sourishkrout:
Great stuff @mxsdev! Good call on spinning out #106. @adambabik could you do a review please?

internal/cmd/root.go
```diff
@@ -16,6 +16,10 @@ var (
 )
 
 func Root() *cobra.Command {
+	exitAfterRun := false
```
adambabik, line 19: Having `tuiCmd` flags here while they belong to `tuiCmd` as well as interleaving `tuiCmd` with the root command might be improved a bit.

In `root.go`, I would add these lines:

```go
tuiCmd := tuiCmd()
// tuiCmd is also a default command
cmd.RunE = tuiCmd.RunE
// remember to copy flags
tuiCmd.Flags().VisitAll(func(f *pflag.Flag) {
	pflags.AddFlag(f)
})
```

And:

```go
cmd.AddCommand(tuiCmd)
```

In this way, we keep `tuiCmd` totally unaware that it is used in any different way than other commands.

adambabik (commented):
It's really good stuff 👍 It works very well!

I have two major comments about the code:
- `tuiModel` can be improved (removal of unnecessary pointers) so that it's more aligned with how bubbletea encourages to write programs.
- `tuiCmd()` can look like any other command and you can avoid interleaving some of its fragments with the root command.

I am happy to talk in more detail on Discord if you wish.

Thanks for the PR!

internal/cmd/tui.go
```diff
@@ -0,0 +1,276 @@
...
+					_, err := fmt.Print("\n")
+					if err != nil {
+						return err
+					}
+				}
+
+				isInitialRun = false
+
+				prog := tea.NewProgram(model)
+				if _, err := prog.Run(); err != nil {
```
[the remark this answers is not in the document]
mxsdev: The reason I had all those pointers initially was because I didn't realize `prog.Run()` returns the model, haha. Thanks for pointing it out!

internal/cmd/root.go
mxsdev: The flag copying code errors because it re-defines the flags set by `setDefaultFlags`. In the most recent PR I used the following code:

```go
	tuiCmd.Flags().VisitAll(func(f *pflag.Flag) {
		if flag := cmd.Flags().Lookup(f.Name); flag == nil {
			cmd.Flags().AddFlag(f)
		}
	})
```

The alternative would be to never run `setDefaultFlags` on the root command and let them cascade from the tui command. I felt that it was better to be explicit than implicit here, even if at the expense of verbosity. Let me know if you want me to do it the other way!

mxsdev:
@adambabik Should all be implemented as you said, let me know if I missed anything!

internal/cmd/run.go
```diff
@@ -41,37 +43,45 @@ func runCmd() *cobra.Command {
...
-			ctx, cancel := ctxWithSigCancel(cmd.Context())
-			defer cancel()
+	return &cmd
+}
 
-			if dryRun {
-				executable.DryRun(ctx, cmd.ErrOrStderr())
-				return nil
-			}
+func runBlockCmd(block *document.CodeBlock, cmd *cobra.Command, opts *runCmdOpts) error {
```
adambabik, line 58: To align the interface with `newExecutable()` you can change it to:

```suggestion
func runBlock(cmd *cobra.Command, block *document.CodeBlock, opts *runCmdOpts) error {
```

internal/cmd/tui.go
```diff
@@ -0,0 +1,263 @@
...
+
+				newModel, err := prog.Run()
+				if err != nil {
+					return err
+				}
+
+				model = newModel.(tuiModel)
+				result := model.result
+
+				if result.run != nil {
```
adambabik, line 228: You can use a technique called the early return to avoid unnecessary nesting:

```go
if result.run != nil {
    break
}

if err := runBlockCmd(result.run, cmd, nil); err != nil {
   ...
}
```

adambabik (approved):
Thank you! I added a few nitpick comments, but it's good to be merged :)

adambabik:
@mxsdev I have one more ask - can you resolve conflicts, please?

mxsdev:
Resolved conflicts, but I realized there is an error when you have a README.md file with no scripts. 

I think in this case we should probably just show a message like "No scripts to run in README.md!" And then close the program. Do you want to open a separate issue for this? Or just solve it here?

sourishkrout:
> Resolved conflicts, but I realized there is an error when you have a README.md file with no scripts.
> 
> I think in this case we should probably just show a message like "No scripts to run in README.md!" And then close the program. Do you want to open a separate issue for this? Or just solve it here?

I'd say let's get this merged and work on the issue above in a separate PR. We can hold off cutting a stable release until that has landed.

amac4015:
iam have to say maybe a little more i dont know trainning or someting is in order since i dont really have tpo muck knowalge on whats the scope of the progect

amac4015:
or not its cool good work guys

PR #104, opened 2023-01-18
Opt to use node-grpc client for codegen

sourishkrout:
This will generate types and implementation code that are compliant with `node-grpc`.

Chose not to commit the generated code yet. I will let you do that post merge, @adambabik.

### DIFF
--- a/buf.gen.yaml
+++ b/buf.gen.yaml
@@ -20,3 +20,8 @@ plugins:
   - plugin: buf.build/bufbuild/es
     out: internal/gen/proto/ts
     opt: target=ts
+
+  # Remote plugin. You can switch to a local one by using "timostamm-protobuf-ts".
+  - plugin: buf.build/community/timostamm-protobuf-ts
+    out: internal/gen/proto/ts
+    opt: client_grpc1,optimize_code_size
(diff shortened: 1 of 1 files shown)


sourishkrout:
As per convo/thread on Slack: this is no longer required @adambabik. Feel free to close unmerged.

sourishkrout:
> As per convo/thread on Slack: this is no longer required @adambabik. Feel free to close unmerged.

Actually we might want to commit it anyways. It's handy to do local codegen. Changed the opts to be in line with Buf's registry.

adambabik:
@sourishkrout just to confirm, the default options from the Buf Remote Plugins do work for our case, is that correct?

sourishkrout:
> @sourishkrout just to confirm, the default options from the Buf Remote Plugins do work for our case, is that correct?

Yep, I took them from the generated files distributed through their registry's headers, e.g.:

```js
/* eslint-disable */
// @generated by protobuf-ts 2.8.2 with parameter output_javascript,optimize_code_size,long_type_string,add_pb_suffix,ts_nocheck,eslint_disable
// @generated from protobuf file "runme/parser/v1/parser.proto" (package "runme.parser.v1", syntax proto3)
// tslint:disable
// @ts-nocheck
```

PR #110, opened 2023-01-25
Fix TUI error when no scripts in readme with simple error

### LINKED ISSUE #109: TUI fails when README.md has no scripts
When README.md has no scripts, the tui is subject to error when pressing enter, and will show the current entry as `1/0`. 

I think in this case, it's easiest to error and close the program. Alternatively, we could show "No Scripts" in the TUI. The latter might be better if we intend to implement a "refresh" functionality at some point.

mxsdev:
Fixes #109 

This takes the simplest solution, which just prints an error message to prevent the TUI from running.

### DIFF
--- a/internal/cmd/tui.go
+++ b/internal/cmd/tui.go
@@ -8,6 +8,7 @@ import (
 
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/mgutz/ansi"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/stateful/runme/internal/document"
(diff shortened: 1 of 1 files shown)


internal/cmd/tui.go
```diff
@@ -196,6 +197,10 @@ func tuiCmd() *cobra.Command {
 				return err
 			}
 
+			if len(blocks) == 0 {
+				return errors.New("no scripts in README.md")
```
adambabik, line 201: As a file might have a different name, let's do this:

```suggestion
				return errors.Errorf("no scripts in %s", fFileName)
```

PR #111, opened 2023-01-26
internal/cmd: cleanup TUI cmd

adambabik:
cc @mxsdev


mxsdev:
LGTM, only thing is whether or not we want to provide a way for users to disable scrolling (show all entries). Originally this was done by providing a number <= 0 to `--entries`, but with this change that will revert to the default number of entries. I can definitely understand not wanting to allow that behavior though.

adambabik:
Ah, I see the point of using `goMath.MaxInt32` now. I will revert that actually. It's indeed odd to disable scrolling, but on the other hand, why not allow it. `entries` will get a default value if not specified anyway.

Issue #108, opened 2023-01-25
Improve error message when no README.md is present


mxsdev:
Currently, if you try to run something like `runme run script` in a folder without a `README.md`, you get something like the following:

```
failed to read /Users/maxstoumen/Projects/runme/README.md: open /Users/maxstoumen/Projects/runme/README.md: no such file or directory
exit status 1
```

Would be nice to have a prettier message here.

mxsdev:
Solved in #116

PR #116, opened 2023-02-03
Improve no readme error message

mxsdev:
Solves #108 

Error is now formatted like:

```
readme file does not exist: ./cmd/README.md
```

### DIFF
--- a/internal/cmd/common.go
+++ b/internal/cmd/common.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/pkg/errors"
@@ -26,6 +27,13 @@ func readMarkdownFile(args []string) ([]byte, error) {
 	if arg == "" {
 		f, err := os.DirFS(fChdir).Open(fFileName)
 		if err != nil {
+			if pathErr, ok := err.(*os.PathError); ok {
+				switch pathErr.Err {
+				case syscall.ENOENT, os.ErrNotExist:
+					return nil, errors.Errorf("readme file does not exist: %s", pathErr.Path)
+				}
+			}
+
 			return nil, errors.Wrapf(err, "failed to read %s", filepath.Join(fChdir, fFileName))
 		}
 		defer func() { _ = f.Close() }()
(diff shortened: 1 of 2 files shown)


internal/cmd/common.go
```diff
@@ -26,6 +27,13 @@ func readMarkdownFile(args []string) ([]byte, error) {
 	if arg == "" {
 		f, err := os.DirFS(fChdir).Open(fFileName)
 		if err != nil {
+			if pathErr, ok := err.(*os.PathError); ok {
+				switch pathErr.Err {
+				case syscall.ENOENT, os.ErrNotExist:
```
adambabik, line 32: `ENOENT` and `ErrNotExist` are [the same errors](https://github.com/golang/go/blob/25f5d9d4a253cb880bff909ebbdc05c8941c4a48/src/syscall/syscall_unix.go#L123-L124). It's very unusual to compare errors returned by stdlib with `syscall` errors.

My suggestion is as follows:

```go
                if err != nil {
-                       return nil, errors.Wrapf(err, "failed to read %s", filepath.Join(fChdir, fFileName))
+                       var pathError *os.PathError
+                       if errors.As(err, &pathError) {
+                               return nil, errors.Errorf("failed to %s markdown file %s: %s", pathError.Op, pathError.Path, pathError.Err.Error())
+                       } else {
+                               return nil, errors.Wrapf(err, "failed to read %s", filepath.Join(fChdir, fFileName))
+                       }
                }
```

Which will produce something like:

```
./runme --filename nonexistent.md ls
failed to open markdown file /path/to/nonexistent.md: no such file or directory
```

adambabik:
Also, we can add an integration test:

```
--- a/testdata/script/basic.txtar
+++ b/testdata/script/basic.txtar
@@ -6,6 +6,10 @@ exec runme ls --allow-unknown
 cmp stdout golden-list-allow-unknown.txt
 ! stderr .
 
+! exec runme ls --filename nonexistent.md
+stderr 'failed to open markdown file .*/nonexistent.md: no such file or directory'
+! stdout .
```

You can execute it like this:
```
go test -timeout 10s -run ^TestRunme$ .
```

mxsdev:
@adambabik Done 👍

internal/cmd/common.go
```diff
@@ -26,6 +26,10 @@ func readMarkdownFile(args []string) ([]byte, error) {
 	if arg == "" {
 		f, err := os.DirFS(fChdir).Open(fFileName)
 		if err != nil {
+			if pathError := (*os.PathError)(nil); errors.As(err, &pathError) {
```
adambabik, line 29: That's clever :) Idiomatic go would be with `var err *os.PathError` in a separate line.

adambabik (approved):

PR #119, opened 2023-02-06
internal/runner: add gRPC service

adambabik:
This PR implements another service called "runner". It is similar to the "kernel" service. The difference is that "kernel" tries to detect when each command starts and ends and based on that allows submission of new commands. "runner" does not that, it does not try to be smart. It still allows to run a shell with TTY (which is a long running program that needs some input to terminate) but relies on a caller to provide commands and input data when needed.

Note: "runner" does not implement the Connect protocol. It's only a plain gRPC service.

Check out `internal/runner/service_test.go` to learn how to use "runner"'s RPC methods. `internal/runner/command_test.go` can also be useful.


internal/api/runme/runner/v1/runner.proto
```diff
@@ -0,0 +1,122 @@
...
+
+message GetSessionRequest {
+    string id = 1;
+}
+
+message GetSessionResponse {
+    Session session = 1;
+}
+
+message ListSessionsRequest {}
```
mxsdev, line 43: Is there any reason we need this command? That is, is there any circumstance in which a process would want to access a session it did not create?

Imagine an attacker gets access to the service. They could then use this command to find existing sessions and print out their environment variables.

internal/runner/command.go
```diff
@@ -0,0 +1,340 @@
...
+			}
+		}
+
+		// stdout is read from pty. stderr is unused because
+		// it can't be distinguished from pty.
+		cmd.Stdout = rbuffer.NewRingBuffer(4096)
+		cmd.Stderr = rbuffer.NewRingBuffer(0)
+	} else {
+		cmd.Stdin = &safeBuffer{buf: bytes.NewBuffer(cfg.Input)}
+		cmd.Stdout = rbuffer.NewRingBuffer(4096)
```
mxsdev, line 158: Why use a ring buffer here? I realize you're concerned about stress on the GC, but I feel like this is an extreme solution. I could imagine, for example, a large JSON payload easily exceeding `4kb`. Or, if many lines are sent as one `stdout` message, that could also easily exceed the limit.

I don't know a lot about go's GC, but is it going to stress the GC severely if we use a dynamic queue instead?

internal/api/runme/runner/v1/runner.proto
```diff
@@ -0,0 +1,122 @@
...
+
+message ListSessionsResponse {
+    repeated Session sessions = 1;
+}
+
+message DeleteSessionRequest {
```
mxsdev, line 49: From my tests, I noticed that sessions are not automatically removed when its creator disconnects.

I realize it may not be trivial to implement such a thing. However, in principle, we are leaking environment variables unnecessarily if a client disconnects unexpectedly. It may be desirable from a security standpoint to try and prevent that.

mxsdev:
Alright, finally got to reviewing this. I did a bunch of local tests and everything checks out from a functionality standpoint. I made some notes on some potential concerns - let me know what you think there.

Environment diffing looks great by the way, really love the way you tackled that problem! Great work overall 🎉 !

internal/api/runme/runner/v1/runner.proto
adambabik: We were thinking about interoperability between the extension and CLI where that could be helpful. Overall, having full CRUD is helpful to run experiments.

Security is a factor, but hiding this method is "security by obscurity" which is not recommended. We might hide it before figuring out something better, though.

internal/runner/command.go
adambabik: It's not really about GC here, because `bytes.Buffer` implementation is pretty smart about it. It's about avoiding the buffer growing indefinitely. I agree that we probably should adjust the numbers, but it can wait until we start testing it. It can be parametrized in the create session request as well.

internal/api/runme/runner/v1/runner.proto
adambabik: Good remarks. I think it's hard to tell at this point whether we want the server or client to have more responsibility. Yes, we can implement automatic session deletion, but that would mean more session restarts which we would like to avoid until we introduce sessions persistency. There're many tradeoffs which will be easier to resolve after we integrate both components and star using them.

PR #121, opened 2023-02-08
Implement `suggest` command

mxsdev:
Implements the `suggest`/`suggest branch` commands, which integrates with the stateful GQL API. Branch is currently working, but will fail on CI.

This is obviously a big PR so here's the highlights:

 - I pruned as many files as I could from `internal/renderer`, but there are still quite a few I cannot remove simply. There are still some unused fields in `model.go`, so let me know if you want me to kill those as well. @adambabik I know you said you thought we could just use `input_model.go` - were you suggesting rewriting all references to `question_model`?
 - @adambabik If you could specifically review 6a811352876cf05be92add8a3a38a383e991359f and 162c075f2c0bdd91c8c91d8cf894b0eb21212531 that would be great, as those are modifications I made to the original source to eliminate linting errors, and it may not be idiomatic or good go code
 - CI will fail on `internal/auth/auth.go:402`, which has a sec issue - `ReadHeaderTimeout` has no value. Not sure how to go forth on this - google says 10s is a good default?
 - I replaced viper with a simpler configuration system in `internal/cmd/common.go`

Some more speculative thoughts:

 - I did not write any extra tests - we may want to add more tests since this is going into prod now
 - There are some TODOs under @adambabik which now may be a good time to do (search the codebase for `TODO(mxs)`)
 - Auth uses same system as the stateful cli, using `DiskStorage` from `internal/auth/disk_storage.go`. We might want to abstract this out if we intend it to be used for more stuff


client/graphql/errors.go
```diff
@@ -0,0 +1,118 @@
...
+func (e *APIError) Error() string {
+	if e.apiError != nil {
+		return e.apiError.Error()
+	}
+
+	if len(e.userErrors) > 0 {
+		var errs []error
+
+		var b bytes.Buffer
+		if _, err := b.WriteString(userErrorString(e.userErrors[0])); err != nil {
```
adambabik, line 63:
```suggestion
		b.WriteString(userErrorString(e.userErrors[0]))
```

is fine because `bytes.Buffer.Write()` and related `WriteString()` etc. do not return non-nil error ever. You can check out [implementation](https://github.com/golang/go/blob/1a09d57de58a90987789ef637083aac21533eeb7/src/bytes/buffer.go#L166-L173).

client/graphql/errors.go
```diff
@@ -0,0 +1,118 @@
...
+			errs = append(errs, err)
+		}
+		for i := 1; i < len(e.userErrors); i++ {
+			if err := b.WriteByte('\n'); err != nil {
```
adambabik, line 67: Similarly here. `_ = b.WriteByte('\n')` is enough.

client/graphql/schema/model/date.go
```diff
@@ -0,0 +1,68 @@
...
+
+func (d DateTime) marshal() ([]byte, error) {
+	return []byte(`"` + time.Time(d).Format(dateTimeScalarFormat) + `"`), nil
+}
+
+func (d DateTime) MarshalJSON() ([]byte, error) {
+	return d.marshal()
+}
+
+func (d DateTime) MarshalGQL(w io.Writer) error {
```
adambabik, line 59: I can't tell that for sure now but I think we can't change signature of this method. At the same time, I am not sure this is needed anymore. It might have been used by a different GQL lib we used to use.

internal/auth/api_server_mock.go
```diff
@@ -0,0 +1,62 @@
...
+	s := &apiServerHandlerMock{ServeMux: mux}
+	mux.HandleFunc(APIAuthEndpoint, func(w http.ResponseWriter, r *http.Request) {
+		if err := authHandler(w, r); err != nil {
+			log.Fatal(err)
+		}
+	})
+	return s
+}
+
+func authHandler(w http.ResponseWriter, r *http.Request) error {
```
adambabik, line 27: It does not make sense to return error from the handler as nothing can be done with it by `mux.HandleFunc`. So you can log errors directly here instead of wrapping `authHandler()` in another func.

internal/renderer/prompt/question.go
```diff
@@ -0,0 +1,82 @@
...
+	case focusMsg:
+		cmds = append(cmds, m.input.Focus())
+	}
+
+	return m, tea.Batch(cmds...)
+}
+
+func (m QuestionModel) View() string {
+	var b strings.Builder
+	if _, err := b.WriteString(m.Text + " [Y/n] " + m.input.View()); err != nil {
```
adambabik, line 74: `strings.Builder` is the same story as `bytes.Buffer`. Errors from `Write*()` can be safely skipped using `_`.

internal/renderer/prompt/input.go
```diff
@@ -0,0 +1,76 @@
...
+	case focusMsg:
+		cmds = append(cmds, m.input.Focus())
+	}
+
+	return m, tea.Batch(cmds...)
+}
+
+func (m InputModel) View() string {
+	var b strings.Builder
+	if _, err := b.WriteString(m.Text + " " + m.input.View()); err != nil {
```
adambabik, line 68: The same here.

internal/renderer/model.go
```diff
@@ -0,0 +1,251 @@
...
+
+	var cmd tea.Cmd
+	m.Child, cmd = m.Child.Update(msg)
+	return m, cmd
+}
+
+func (m Model) View() string {
+	var b strings.Builder
+
+	var writeErr error
```
adambabik, line 104: This is not needed as I explained in another comment.

client/graphql/client.go
```diff
@@ -0,0 +1,117 @@
...
+	graphql.Client
+}
+
+func New(endpoint string, httpClient *http.Client) (*Client, error) {
+	return &Client{
+		Client: graphql.NewClient(endpoint, httpClient),
+	}, nil
+}
+
+func (c *Client) GetStandup(ctx context.Context, date *string, upcoming bool) (result query.Standup, err error) {
```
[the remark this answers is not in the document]
mxsdev: Had to keep `GetUser` and `ContextWithTrackInput` since they were referenced in `suggest`

main.go
```diff
@@ -15,6 +16,10 @@ var (
 )
 
 func root() int {
+	version.BuildDate = BuildDate
```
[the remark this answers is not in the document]
mxsdev: See other comment (for `Version.go`)

mxsdev:
CI is green 🙌 🚀 🎉

@adambabik Everything should be solved except for the Version issue which I'm not sure about. If you're seeing this tomorrow morning feel free to fix that up.

Keep in mind as per Sebastian's words:

> One thing I will do tmrw is to require the suggestions APIs to check for the site-admin flag to be set. Primarily to be in compliance with OpenAI's release guidelines. We passed their initial app review but need to have our PR/announcement be signed off by OpenAI before we are allowed exercise their APIs on end-users' behalf.

So we probably shouldn't merge this until his approval!!

adambabik (commented):
@mxsdev Looks great! 💪 We are pretty close to call it done :)

Last few comments I missed previously:
- Let's move `client` package to `internal/client`. It's a public repo and we don't want to manifest that `client` is safe to use by others. Overall, anything in `internal/` can't be imported by others. Any other package from a public Go module can be.
- `genqlient.yaml` is missing. You can find it [here](https://github.com/stateful/cli/blob/main/genqlient.yaml). I would try to put this file into `internal/client/graphql` and the same goes for [a file with the `go:generate genqlient` directive](https://github.com/stateful/cli/blob/10bae62538946f06b90a5b1a4699ce1fa2ddfed5/generate.go#L3) - without it, the GQL client is not regenerated at all.
- Content of `./internal/client/graphql/schema/README.md` should be updated. Please try to execute all steps to make sure it works.
- Due to how our GraphQL server formats datetimes, `date.go` must be the same as https://github.com/stateful/cli/blob/main/client/graphql/schema/model/date.go. I deleted unused methods only but especially those related to [`json.Marshaler`](https://pkg.go.dev/encoding/json#Marshaler) and `json.Unmarshaler` interfaces must stay.
- The `renderer` package is related to markdown. Let's move all reusable bubbletea-related files to `internal/tui`. I finally deleted `internal/suggestions` completely and moved all generic types to `internal/tui` and it worked.

I tested a bunch of these changes in https://github.com/stateful/runme/commit/3b70dce331682eab2c1ee9cbffe49dea92eb0201.

internal/cmd/utils.go
```diff
@@ -0,0 +1,162 @@
...
+	"strings"
+
+	"github.com/henvic/httpretty"
+	"github.com/mattn/go-isatty"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/stateful/runme/client"
+	"github.com/stateful/runme/client/graphql/query"
+
+	// "github.com/stateful/runme/client/rest"
```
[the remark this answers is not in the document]
mxsdev: Looks like you handled this in 3b70dce331682eab2c1ee9cbffe49dea92eb0201

mxsdev:
@adambabik 

Went over your changes and it all looks good - think I got everything, let me know if there's anything I missed!

I was able to run the codegen scripts for the schema no problem. There was an update and I pushed it, hope that's alright (we can easily revert it if not).

sourishkrout:
> Keep in mind as per Sebastian's words, we have to pass an OpenAI review before we can execute queries on users' behalf. So we probably shouldn't merge this until we get his approval!!

Thanks for keeping an eye out for me 👍.  I blocked the OpenAI calls on the API level for now to be on the safe side. No need to hold off merging into `main`. We can do prime-time polish in separate PRs.

adambabik (approved):
I think we can merge it to master. Any other changes or improvements can be done in separate PRs. A discussion about those is in https://github.com/stateful/runme/issues/123.

Awesome work @mxsdev!

PR #125, opened 2023-02-13
internal/cmd: clean up texts and visible cmds

adambabik:
Highlights:
- Commands' short description starts with an upper case and does not end with a point. The same goes for flags' description.
- Removed unnecessary TODO comments.
- Hid flags and commands that should not be exposed.


mxsdev (approved):
LGTM. Meant to remove those TODO's before merge, my bad there

PR #128, opened 2023-02-15
internal/kernel: remove kernel and move shell session to runner

adambabik:
The `kernel` package is not really needed anymore. All responsibilities were moved to `runner`.


adambabik:
@mxsdev @sourishkrout do you rely on the kernel service at all or you fully switched to the runner?

mxsdev:
@adambabik fully switched to the runner. There is no code using the kernel intended to be merged to main over there.

sourishkrout:
> @adambabik fully switched to the runner. There is no code using the kernel intended to be merged to main over there.
> 
> @sourishkrout says we might want to circle back and add the kernel, but that's speculative at the moment.

👍 cool to nuke kernel-related code on both ends. If we wind up reviving the idea of it we'll look into the repo's history.

Issue #131, opened 2023-02-15
internal/runner: add exit flag to `ExecuteRequest`


mxsdev:
Programs not run in TTY mode currently have no way to request a program exit. There should be a new flag in `ExecuteRequest` which will perform program exit.

It's probably a good foresight to make this field an enum, in case there are different kinds of "stop" requests - for example, I know node.js `child_process` has a distinction between regular "stop" and "force stop." Lmk what you think @adambabik

Issue #130, opened 2023-02-15
internal/runner: send `ExitCode` on program exit in TTY mode


mxsdev:
As discussed with @adambabik, programs run in TTY mode have no way of knowing when the program exits. 

Iirc, you were saying we should poll the program and send an `ExitCode` in an `ExecuteResponse` when the program ends in this case.

Issue #129, opened 2023-02-15
internal/cmd/server: Logger is development logger in prod


mxsdev:
The problem is this line in `internal/cmd/server.ts`:

```ts
logger, err := zap.NewDevelopment()
```

Development logger cannot be interpreted by the client-side since it is not JSON.

Might be nice as well to have a special field in the "ready" message so that the client-side can easily determine when the server is up and listening @adambabik

PR #134, opened 2023-02-16
Apply various runner service fixes

adambabik:
Highlights:
- `runme server` now has the `--dev` flag to enable a development mode. Currently, it affects only logs which will have a more friendly formatting and a debug level. Without the flag, the logs are JSON structured logs and have an info level. Closes https://github.com/stateful/runme/issues/129.
- `ExecuteRequest` has a new field called `stop` that is an enum. It allows terminating the running command. Closes https://github.com/stateful/runme/issues/131.
- Logic waiting for the command to finish got improved. It should handle cases when a command is run with TTY (aka interactive mode) and finishes on its own. The RPC method will exit instead of hanging until the client closes the send direction. Closes https://github.com/stateful/runme/issues/130.


mxsdev:
@adambabik 

After testing client-side here's what I have:

**Stop**

 - For non-TTY, since the client has to end its input stream early, there is no way to send a stop request in this case. 
 - For TTY, works great!

**TTY End**

 - Works fine, but after the session ends, my logs look like this:
```
2023-02-16T09:36:16.070-0600	INFO	runner/service.go:126	running Execute in runnerService
2023-02-16T09:36:17.075-0600	INFO	runner/command.go:273	read all from stdin
2023-02-16T09:36:17.076-0600	INFO	runner/service.go:300	finished command execution	{"exitCode": 0}
2023-02-16T09:36:17.076-0600	INFO	runner/service.go:208	failed to receive requests; stopping program	{"error": "rpc error: code = Canceled desc = context canceled"}
2023-02-16T09:36:17.076-0600	INFO	runner/command.go:296	failed to close pty; continuing
2023-02-16T09:36:17.076-0600	INFO	runner/command.go:302	failed to terminate process group; trying Process.Signal()	{"error": "no such process"}
2023-02-16T09:36:17.076-0600	INFO	runner/command.go:304	failed to signal process; trying Process.Kill()	{"error": "os: process already finished"}
2023-02-16T09:36:17.076-0600	INFO	runner/service.go:210	failed to stop program	{"error": "os: process already finished", "errorVerbose": "os: process already finished\ngithub.com/stateful/runme/internal/runner.(*command).stop\n\t/Users/mxs/Projects/runme/internal/runner/command.go:305\ngithub.com/stateful/runme/internal/runner.(*command).Kill\n\t/Users/mxs/Projects/runme/internal/runner/command.go:282\ngithub.com/stateful/runme/internal/runner.(*runnerService).Execute.func3\n\t/Users/mxs/Projects/runme/internal/runner/service.go:209\nruntime.goexit\n\t/opt/homebrew/Cellar/go/1.19.5/libexec/src/runtime/asm_arm64.s:1172"}
```
I think we're running into the issue I was mentioning before - the server ends the stream, which causes `srv.Recv` to get an error since it has to stop blocking. Looks like somehow there are then two requests to end the process happening. 

**Other**

Getting the following after trying to `INTERRUPT` a program:

```
2023-02-16T09:31:02.870-0600	INFO	runner/service.go:281	command finalizer failed	{"error": "open /var/folders/cs/k8y186rn3rv2gm63f46pnmfr0000gn/T/713880327/.env_end: no such file or directory", "errorVerbose": "open /var/folders/cs/k8y186rn3rv2gm63f46pnmfr0000gn/T/713880327/.env_end: no such file or directory\ngithub.com/stateful/runme/internal/runner.(*command).readEnvFromFile\n\t/Users/mxs/Projects/runme/internal/runner/command.go:314\ngithub.com/stateful/runme/internal/runner.(*command).collectEnvs\n\t/Users/mxs/Projects/runme/internal/runner/command.go:336\ngithub.com/stateful/runme/internal/runner.(*command).Finalize\n\t/Users/mxs/Projects/runme/internal/runner/command.go:359\ngithub.com/stateful/runme/internal/runner.(*runnerService).Execute\n\t/Users/mxs/Projects/runme/internal/runner/service.go:280\ngithub.com/stateful/runme/internal/gen/proto/go/runme/runner/v1._RunnerService_Execute_Handler\n\t/Users/mxs/Projects/runme/internal/gen/proto/go/runme/runner/v1/runner_grpc.pb.go:238\ngoogle.golang.org/grpc.(*Server).processStreamingRPC\n\t/Users/mxs/go/pkg/mod/google.golang.org/grpc@v1.53.0/server.go:1620\ngoogle.golang.org/grpc.(*Server).handleStream\n\t/Users/mxs/go/pkg/mod/google.golang.org/grpc@v1.53.0/server.go:1708\ngoogle.golang.org/grpc.(*Server).serveStreams.func1.2\n\t/Users/mxs/go/pkg/mod/google.golang.org/grpc@v1.53.0/server.go:965\nruntime.goexit\n\t/opt/homebrew/Cellar/go/1.19.5/libexec/src/runtime/asm_arm64.s:1172"}
```

This error only happens when running `INTERRUPT` (with tty), but not when running `KILL`.

adambabik:
> For non-TTY, since the client has to end its input stream early, there is no way to send a stop request in this case.

That's true. That's why I pushed for cancelling the connection from the client which would be equivalent to killing the process, but that's out of equation.

After introducing `ExecuteRequest.stop`, the client should never close the send direction. I need to fix the code to make it work without this assumption.

Also, based on our Slack convo, when the handler exits and there is an active `srv.Recv()`, it will get the `rpc canceled` error. I need to fix that as well as I thought `rpc canceled` is possible to receive only if the client sends it.

> For TTY, it works fine, but I have the following error after trying to INTERRUPT a program:

This is a bit odd because the logic is the same and both signals `INT` and `KILL` should work the same in the worst case scenario. Regarding the `command finalizer failed` log, I can recognize that a process was interrupted and avoid trying to collect envs as it's doomed anyway if the process is interrupted (unless we improve the logic behind collecting envs but it's not going to happen very soon).

> Works fine, but after the session ends, my logs look like this:

> I think we're running into the issue I was mentioning before - the server ends the stream, which causes srv.Recv to get an error since it has to stop blocking. Looks like somehow there are then two requests to end the process happening.

Yes, my assumption was incorrect as described above in this comment.

Stay tuned for fixes! :)

mxsdev:
@adambabik 

Good stuff!

> This is a bit odd because the logic is the same and both signals INT and KILL should work the same in the worst case scenario. Regarding the command finalizer failed log, I can recognize that a process was interrupted and avoid trying to collect envs as it's doomed anyway if the process is interrupted (unless we improve the logic behind collecting envs but it's not going to happen very soon).

Yeah, I did some digging and it would appear that the following line in `command.go` is basically not being executed by the shell on interrupt:

```
		_, _ = script.WriteString("env > " + filepath.Join(envStorePath, envEndFileName) + "\n")
```

Which makes sense, because the program will never get to that point in the script if interrupted. Not sure why it still executes on `kill` though (maybe this was a user error?)

At any rate, I definitely think that we should not try to collect envs in the case of an interrupt. Arguably we probably don't want to collect envs unless the exit code is `0`. Or we can just do a check to make sure that the env end file exists, and if not, not try to collect envs (and maybe log).

adambabik:
> I definitely think that we should not try to collect envs in the case of an interrupt. Arguably we probably don't want to collect envs unless the exit code is 0.

That's my strategy for now. I think collecting envs can be done much smarter, for example using `trap`. Then, we could collect them even if the program failed or was interrupted.

internal/runner/service.go
```diff
@@ -182,13 +183,22 @@ func (r *runnerService) Execute(srv runnerv1.RunnerService_ExecuteServer) error
...
-		defer func() {
-			_ = stdinWriter.Close()
-		}()
+		defer func() { _ = stdinWriter.Close() }()
 
 		if len(req.InputData) > 0 {
-			if _, err := stdinWriter.Write(req.InputData); err != nil {
+			_, err := stdinWriter.Write(req.InputData)
+
+			// When TTY is false, the service assumes that there will be no more input data and
```
adambabik, line 193: @mxsdev please verify if this assumption is correct. I need to close the `stdin` writer at some point and as we agreed that the client won't close the send direction, I need some other indication.

adambabik:
> Which makes sense, because the program will never get to that point in the script if interrupted. Not sure why it still executes on `kill` though (maybe i just misread the logs idk)

I can't reproduce, at least with https://github.com/stateful/runme/pull/134/commits/0bb412dcf1ef3f0f57a81361d3a91921aa48a342. The result is the same regardless of the signal I send:

```
$ buf curl --protocol grpc --http2-prior-knowledge http://localhost:7890/runme.runner.v1.RunnerService/Execute -d @-
{"program_name": "/usr/local/bin/bash", "commands": ["cat - | tr a-z A-Z"], "tty": true}
{"stop": 1}
{"exitCode":130}

$ buf curl --protocol grpc --http2-prior-knowledge http://localhost:7890/runme.runner.v1.RunnerService/Execute -d @-
{"program_name": "/usr/local/bin/bash", "commands": ["cat - | tr a-z A-Z"], "tty": true}
{"stop":2}
{"exitCode":137}
```

Logs from the servers:

```
2023-02-16T19:10:18.170+0100    INFO    runner/service.go:126   running Execute in runnerService
2023-02-16T19:10:18.171+0100    DEBUG   runner/service.go:176   command config  {"cfg": {"ProgramName":"/usr/local/bin/bash","Args":null,"Directory":"","Session":{"ID":"cfn732k3c37gk8tks74g","Metadata":null},"Tty":true,"Stdin":{},"Stdout":{},"Stderr":{},"IsShell":true,"Commands":["cat - | tr a-z A-Z"],"Script":"","Logger":{}}}
2023-02-16T19:10:35.912+0100    INFO    runner/service.go:233   requested the program to stop
2023-02-16T19:10:35.914+0100    INFO    runner/service.go:292   command finished        {"exitCode": 130}
2023-02-16T19:10:35.915+0100    INFO    runner/service.go:305   command was finalized successfully
2023-02-16T19:10:35.915+0100    INFO    runner/service.go:321   sending the final response with exit code
2023-02-16T19:10:51.111+0100    INFO    runner/service.go:126   running Execute in runnerService
2023-02-16T19:10:51.111+0100    DEBUG   runner/service.go:176   command config  {"cfg": {"ProgramName":"/usr/local/bin/bash","Args":null,"Directory":"","Session":{"ID":"cfn73as3c37gk8tks750","Metadata":null},"Tty":true,"Stdin":{},"Stdout":{},"Stderr":{},"IsShell":true,"Commands":["cat - | tr a-z A-Z"],"Script":"","Logger":{}}}
2023-02-16T19:10:57.440+0100    INFO    runner/service.go:233   requested the program to stop
2023-02-16T19:10:57.440+0100    INFO    runner/service.go:292   command finished        {"exitCode": 137}
2023-02-16T19:10:57.441+0100    INFO    runner/service.go:305   command was finalized successfully
2023-02-16T19:10:57.441+0100    INFO    runner/service.go:321   sending the final response with exit code
```

mxsdev:
@adambabik 

Just tested on client, it's all fixed, except that the issue of non-PTY programs not exiting is still present - see my response on your comment, the code to close stdin never gets called since `InputData` is empty on initial request.

> I can't reproduce, at least with https://github.com/stateful/runme/commit/0bb412dcf1ef3f0f57a81361d3a91921aa48a342

Hmm, not sure what's going on here then.

mxsdev:
@adambabik Works now 👍

mxsdev (approved):
LGTM

internal/runner/service.go
adambabik: Ugh, last second fixes. I changed that in https://github.com/stateful/runme/pull/134/commits/045466cad9f443d221f20f9635bfbe2e381cbfec.

> Other than that, yeah this is fine, but it will disable to ability to send stdin at all for non-PTY for any clients. Hopefully that's alright.

I can imagine a scenario when one wants to execute some command and provide some input through STDIN without starting a TTY. I tested it that way before. I will figure out something tomorrow.

mxsdev: Gotcha. Maybe in a follow up pr?

Issue #136, opened 2023-02-16
Generate keypair and sign self-signed cert at startup for mTLS


sourishkrout:
Preliminary requirements:

- Generate keypair at first startup (let's recycle them)
- Set private key to `chmod 600`
- Sign self-signed cert (all relying parties use the same) at every startup with 7-30d validity (signing should be fast SHA256)
- Think about cert rotation strategy - however, gRPC likely won't request TLS handshakes for standing connections

While we do want to leverage mTLS we really want to mimic a UDS socket's security characteristics. If Windows natively supported them we would just use them instead of mTLS & TCP loopback.

Research: https://www.notion.so/statefulhq/gRPC-API-Security-Notes-687b0fe4a9944653baed40424902f38c

Must have to ship the Runner, however, not required for the Parser (stateless).

sourishkrout:
One aspect, that crossed my mind, is file permissions on Windows. We should make sure fs permissions work as expected since this whole security concept is rooted in it, @mxsdev. Not immediately important. Just dropping a note so we can come back to it.

PR #137, opened 2023-02-16
internal/runner: support TLS

mxsdev:
Regarding #136 

Mostly going for poc/research at the moment.


mxsdev:
@sourishkrout For testing, use the following commands:

```sh
# For server
go run . server --address localhost:7890 --dev --tls /tmp/runme/cert
# For client
go run ./examples/grpc-client --file <file> --addr <addr> --tls /tmp/runme/cert
```

Should be working over both unix socket and tcp. Client should not be able to tls-enabled server w/o tls flag. Both of these things worked in my basic tests 👍

PR #139, opened 2023-02-17
internal/runner: handle interactive commands in Shell

adambabik:
Assume we have the following README command:

````md
```sh {name=xxx interactive=true}
tr a-z x
```
````

It should be possible to run it with `runme run xxx`. Unfortunately, the current implementation does not work correctly - it does start the command and waits for the input, but it does not produce any output. This PR fixes it by enabling TTY if the code block contains annotation `interactive=true`.

Other changes include small refactoring of structs implementing `runner.Executable` interface as well as more logs and tests for `runner.runnerService.Execute()`.

### DIFF
--- a/internal/runner/service.go
+++ b/internal/runner/service.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 
 	"github.com/pkg/errors"
+	"github.com/rs/xid"
 	runnerv1 "github.com/stateful/runme/internal/gen/proto/go/runme/runner/v1"
 	"github.com/stateful/runme/internal/rbuffer"
 	"go.uber.org/zap"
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: internal/runner/service.go [hunk 1]
```diff
@@ -123,19 +124,23 @@ func (r *runnerService) findSession(id string) *Session {
 }
 
 func (r *runnerService) Execute(srv runnerv1.RunnerService_ExecuteServer) error {
-	r.logger.Info("running Execute in runnerService")
+	logger := r.logger.With(zap.String("_id", xid.New().String()))
```
mxsdev, line 127: Why include an `_id` field here? The field will not propagate through the runner so it will only be for top-level log calls in this function. Is there any reason we want those to be distinguished?

mxsdev (changes requested):
Overall LGTM. To be honest I have maybe a few architectural questions but I'll ask on Wednesday's off-boarding session.

internal/runner/service.go
adambabik: Exactly, it makes it easier to group logs per single execution.

PR #140, opened 2023-02-17
internal/cmd: set max msg size for gRPC server

adambabik:
Highlights:
- explicitly define the max message size
- add a test case for the large input
- add an example with Go client

### DIFF
--- a/examples/grpc-client/main.go
+++ b/examples/grpc-client/main.go
@@ -1,0 +1,133 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"io"
+	"log"
+	"os"
+
+	"github.com/pkg/errors"
+	runnerv1 "github.com/stateful/runme/internal/gen/proto/go/runme/runner/v1"
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+var (
+	addr       = flag.String("addr", "127.0.0.1:7890", "the address to connect to")
+	file       = flag.String("file", "", "file with content to upper case")
+	resultFile = flag.String("write-result", "-", "path to a result file (default: stdout)")
+)
+
+func main() {
+	flag.Parse()
+
+	if err := run(); err != nil {
+		log.Fatalf("error: %v", err)
+	}
+}
+
+func run() error {
+	conn, err := grpc.Dial(*addr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return errors.Wrap(err, "failed to connect")
+	}
+	defer conn.Close()
+
+	client := runnerv1.NewRunnerServiceClient(conn)
+
+	stream, err := client.Execute(context.Background())
+	if err != nil {
+		return errors.Wrap(err, "failed to call Execute()")
+	}
+
+	var g errgroup.Group
+
+	g.Go(func() error {
+		source, err := os.Open(*file)
+		if err != nil {
+			return errors.Wrap(err, "failed to open source file")
+		}
+		defer func() { _ = source.Close() }()
+
+		buf := make([]byte, 32*1024)
+
+		for readNext := true; readNext; {
+			n, err := source.Read(buf)
+			if err != nil {
+				if !errors.Is(err, io.EOF) {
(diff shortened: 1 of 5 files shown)


internal/runner/service_test.go
```diff
@@ -218,6 +218,61 @@ func Test_runnerService(t *testing.T) {
 		assert.EqualValues(t, 0, result.Responses[1].ExitCode.Value)
 	})
 
+	// The longest accepted line must not have more than 1024 bytes,
+	// including the new line character at the end. Any line longer results in ^G (BELL).
+	// It is possible to send more data, but it must be divided in 1024-byte chunks
+	// separated by the new line character (\n).
+	// TODO(adamb): sort out the root cause of this limitation.
```
adambabik, line 225: @mxsdev @sourishkrout this is a quite strange limitation I found on my local machine. On Linux, I can write 4096. This limit can be found on [the termios man page](https://man7.org/linux/man-pages/man3/termios.3.html):

> The maximum line length is 4096 chars (including the
         terminating newline character); lines longer than 4096 chars
         are truncated.  After 4095 characters, input processing (e.g.,
         ISIG and ECHO* processing) continues, but any input data after
         4095 characters up to (but not including) any terminating
         newline is discarded.  This ensures that the terminal can
         always receive more input until at least one line can be read.

I haven't found yet which settings is responsible for lower limit on macOS 12.6.

adambabik: I found [it](https://github.com/apple/darwin-xnu/blob/main/bsd/sys/syslimits.h#L98)! It's mentioned in [the manual](https://man.freebsd.org/cgi/man.cgi?query=termios&sektion=4) as `{MAX_INPUT}`.

examples/grpc-client/main.go
```diff
@@ -0,0 +1,133 @@
...
+func run() error {
+	conn, err := grpc.Dial(*addr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return errors.Wrap(err, "failed to connect")
+	}
+	defer conn.Close()
+
+	client := runnerv1.NewRunnerServiceClient(conn)
+
+	stream, err := client.Execute(context.Background())
```
mxsdev, line 40:
```suggestion
	g, ctx := errgroup.WithContext(context.Background())

	stream, err := client.Execute(ctx)
```

Stream needs to close early if there's an error, for example, reading the file

internal/runner/service_test.go
```diff
@@ -218,6 +219,65 @@ func Test_runnerService(t *testing.T) {
 		assert.EqualValues(t, 0, result.Responses[1].ExitCode.Value)
 	})
 
+	// The longest accepted line must not have more than 1024 bytes on macOS,
+	// including the new line character at the end. Any line longer results in ^G (BELL).
+	// It is possible to send more data, but it must be divided in 1024-byte chunks
```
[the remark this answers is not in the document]
adambabik: > So, does this mean that the client-side should split the stdin into 1024 byte chunks separated by new line? Is there any reason we couldn't handle that on the server upon receiving an InputData request?

It's definitely an option, but it may not give the expected result.

Overall, commands like `runme run my-cmd < input.json` are affected and we should advise to put `< input.json` inside the `my-cmd`.

I'll update this comment because it's 1024 on macOS, but on Linux it's 4096 by default. Check out https://github.com/stateful/runme/pull/140#discussion_r1112140881.

internal/cmd/server.go
```diff
@@ -123,8 +123,10 @@ The kernel is used to run long running processes like shells and interacting wit
 
 			logger.Info("started listening", zap.String("addr", lis.Addr().String()))
 
-			var opts []grpc.ServerOption
-			server := grpc.NewServer(opts...)
+			server := grpc.NewServer(
+				grpc.MaxRecvMsgSize(runner.MaxMsgSize),
+				grpc.MaxSendMsgSize(runner.MaxMsgSize),
```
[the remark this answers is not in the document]
adambabik: Yes, just to be safe. However, we already limit the size of messages sent in `readLoop`. The buffer there is 1 MB so the message will be ~1 MB at the top.

mxsdev (approved):

Issue #144, opened 2023-02-21
internal/runner: multiline inline executions get cut off


mxsdev:
The following test (in `service_test.go/Test_runnerService`) will not pass - the resulting `$LICENSE` will get cut off after the first new line. This is a test script in the example `README` file.

```golang
	t.Run("ExecuteLicenseCat", func(t *testing.T) {
		t.Parallel()

		stream, err := client.Execute(context.Background())
		require.NoError(t, err)

		execResult := make(chan executeResult)
		go getExecuteResult(stream, execResult)

		err = stream.Send(&runnerv1.ExecuteRequest{
			ProgramName: "bash",
			Commands: []string{
				"export LICENSE=$(cat ../LICENSE)",
				"echo \"LICENSE: $LICENSE\"",
			},
		})
		assert.NoError(t, err)

		result := <-execResult

		assert.Len(t, result.Responses, 3)
		assert.Greater(t, len(result.Responses[1].StdoutData), 20)
	})
```

adambabik:
I can reproduce it. The problem is with reusing buffers in `readLoop()`. I thought that using two alternating buffers is enough but no message can be altered/reused after calling `SendMsg()` and there is no indicator when that's possible except for GC.

PR #145, opened 2023-02-22
internal/runner: remove bufs reuse in readLoop

adambabik:
gRPC server must not modify message after calling SendMsg. More: https://github.com/grpc/grpc-go/issues/5857.

Closes https://github.com/stateful/runme/issues/144


mxsdev (approved):
We went over this on call yesterday. No problems from me, and I'll re-add that `ExecuteCurl` test as part of #149 before merge 👍

PR #146, opened 2023-02-22
internal/cmd: allow exec commands using gRPC server

adambabik:
This PR adds an option for TUI to run commands through a gRPC server.

First, start the gRPC server: `./runme server -a localhost:7890`. Then pass the address of the server: `./runme tui --server localhost:7890`.

It also introduces `internal/runner/{local,remote}` which are high-level abstractions to run commands either locally or using RPC.

Closes https://github.com/stateful/runme/issues/135

### DIFF
--- a/internal/cmd/tui.go
+++ b/internal/cmd/tui.go
@@ -11,9 +11,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/stateful/runme/internal/document"
 	rmath "github.com/stateful/runme/internal/math"
-	"github.com/stateful/runme/internal/runner"
-	"github.com/stateful/runme/internal/runner/local"
-	"github.com/stateful/runme/internal/runner/remote"
+	"github.com/stateful/runme/internal/runner/client"
 	"github.com/stateful/runme/internal/version"
 )
 
@@ -42,24 +40,39 @@ func tuiCmd() *cobra.Command {
 				visibleEntries = math.MaxInt32
 			}
 
-			var (
-				remoteRunner *remote.Runner
-				localRunner  *local.Runner
-			)
+			var runner client.Runner
+
+			defer func() { _ = runner.Cleanup(cmd.Context()) }()
+
+			opts := []client.RunnerOption{
+				client.WithDir(fChdir),
+				client.WithStdin(cmd.InOrStdin()),
+				client.WithStdout(cmd.OutOrStdout()),
+				client.WithStderr(cmd.ErrOrStderr()),
+			}
 
 			if serverAddr != "" {
-				remoteRunner, err = remote.New(
+				remoteRunner, err := client.NewRemoteRunner(
 					cmd.Context(),
 					serverAddr,
-					remote.WithDir(fChdir),
-					remote.WithStdin(cmd.InOrStdin()),
-					remote.WithStdout(cmd.OutOrStdout()),
-					remote.WithStderr(cmd.ErrOrStderr()),
+					opts...,
 				)
+
 				if err != nil {
 					return errors.Wrap(err, "failed to create remote runner")
 				}
-				defer func() { _ = remoteRunner.Cleanup(cmd.Context()) }()
+
+				runner = remoteRunner
+			} else {
+				localRunner, err := client.NewLocalRunner(
+					opts...,
+				)
+
+				if err != nil {
+					return errors.Wrap(err, "failed to create local runner")
+				}
+
+				runner = localRunner
 			}
 
 			model := tuiModel{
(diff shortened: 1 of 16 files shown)


internal/runner/remote/runner_remote.go
```diff
@@ -0,0 +1,180 @@
...
+
+func (r *Runner) setupSession(ctx context.Context) error {
+	resp, err := r.client.CreateSession(ctx, &runnerv1.CreateSessionRequest{})
+	if err != nil {
+		return errors.Wrap(err, "failed to create session")
+	}
+	r.sessionID = resp.Session.Id
+	return nil
+}
+
```
mxsdev, line 147:
```suggestion
func (r *Runner) deleteSession(ctx context.Context) error {
	if r.sessionID != "" {
		_, err := r.client.DeleteSession(ctx, &runnerv1.DeleteSessionRequest{Id: r.sessionID})
		if err != nil {
			return errors.Wrap(err, "failed to delete session")
		}
		r.sessionID = ""
		return nil
	}
}
```

I think we should add a delete session request, and run it before the `tui` command exits.

internal/runner/remote/runner_remote.go
```diff
@@ -0,0 +1,180 @@
...
+			if msg.ExitCode.Value > 0 {
+				return &ExitError{Code: uint(msg.ExitCode.Value)}
+			}
+			return nil
+		}
+	}
+}
+
+func (r *Runner) setupSession(ctx context.Context) error {
+	resp, err := r.client.CreateSession(ctx, &runnerv1.CreateSessionRequest{})
```
mxsdev, line 140: You have to populate CreateSessionRequest's `Envs` with the current environment, or else it might end up dropping certain variables. We had this same issue with the vscode client.  

As an example of how this can go wrong, `cd` into `vscode-runme/examples` and try to run the node script after the hello world script:

image

internal/runner/remote/runner_remote.go
```diff
@@ -0,0 +1,180 @@
...
+	return nil
+}
+
+func (r *Runner) RunBlock(ctx context.Context, block *document.CodeBlock) error {
+	stream, err := r.client.Execute(ctx)
+	if err != nil {
+		return err
+	}
+
+	tty := block.Interactive()
```
mxsdev, line 154: This needs to be true by default to match the behavior of `vscode-runme`.

For example, in `vscode-runme/examples`, there is a script to run a node script that accepts stdin, but it doesn't have an interactive field set.

mxsdev (changes requested):
Found some bugs - mostly stuff similar to what I found doing client-side implementation in `vscode-runme`.

Also discovered #153 for local runners.

internal/cmd/tui.go
```diff
@@ -12,10 +12,130 @@ import (
...
+				remoteRunner *remote.Runner
+				localRunner  *local.Runner
+			)
+
+			if serverAddr != "" {
+				remoteRunner, err = remote.New(
+					ctx,
+					serverAddr,
+					remote.WithDir(fChdir),
+					remote.WithStdin(cmd.InOrStdin()),
```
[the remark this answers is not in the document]
adambabik: I fixed it to some degree but not fully. 

The problem is that we need to read from stdin in the case of interactive blocks. Reading is a blocking operation and there is no clear way how to interrupt it when the command finishes. I am talking about `remote.Runner.sendLoop()` and the `stdin.Read()` call inside of it.

The current situation is that sometimes, when the execution of command finishes, there is a delay when the list of commands is visible again.

There are many Linux-specific ways to handle that, for example this [SO thread](https://stackoverflow.com/questions/717572/how-do-you-do-non-blocking-console-i-o-on-linux-in-c), but the implementation is not trivial. My recommendation is to create an issue and fix it separately.

internal/runner/remote/runner_remote.go
adambabik: Good point. I added a cleanup step that removes the session when the remote runner is not needed anymore.

internal/cmd/tui.go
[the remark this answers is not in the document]
adambabik: I created an issue: https://github.com/stateful/runme/issues/156.

mxsdev:
@sourishkrout 

Ready for another review!

sourishkrout:
> tui --server localhost:7890

Getting a panic stack trace results from the `--runner` not being set. Could we have a nicer error if the server API is not available?

sourishkrout (approved):
Please see that one comment. Otherwise, great work 👍 . Played through various scenarios and works great.

mxsdev:
@sourishkrout 

Fixed the segfault that's causing this, now the error message is:

```
failed to create remote runner: failed to create session: rpc error: code = Unimplemented desc = unknown service runme.runner.v1.RunnerService
exit status 1
```

Let me know if you're ok with that error message. I can probably make it nicer to handle this specific case, but it would be inconsistent throughout the codebase.

sourishkrout:
> @sourishkrout
> 
> Fixed the segfault that's causing this, now the error message is:
> 
> ```
> failed to create remote runner: failed to create session: rpc error: code = Unimplemented desc = unknown service runme.runner.v1.RunnerService
> exit status 1
> ```
> 
> Let me know if you're ok with that error message. I can probably make it nicer to handle this specific case, but it would be inconsistent throughout the codebase.

This works 👍 . Thank you!

PR #150, opened 2023-02-23
internal/rbuffer: fix overflow hanging

### LINKED ISSUE #149: internal/rbuffer: Ring buffer reads indefinitely when overloaded
The ring buffer, when overloaded, can sometimes end up reading indefinitely. ~~Ostensibly the read counter is not being incremented properly.~~

(Upon further investigation, read counter is being incremented, but `isFull` parameter is never reset by `read` method)

The following test case (in `ring_buffer_test.go`) fails, but should pass: 

```golang
	t.Run("ExceedingInputTerminates", func(t *testing.T) {
		data := append(
			bytes.Repeat([]byte{'1'}, 1024),
			bytes.Repeat([]byte{'2'}, 25*1024)...,
		)

		buf := NewRingBuffer(10 * 1024)
		n, err := buf.Write(data)
		assert.Nil(t, err)
		assert.Equal(t, buf.size, n)

		expected := bytes.Repeat([]byte{'2'}, buf.size)
		iterationCount := 0

		for {
			iterationCount++

			got := make([]byte, len(expected))

			n, err = buf.read(got)
			assert.Nil(t, err)

			assert.Equal(t, expected, got)

			if n == 0 {
				break
			}

			if iterationCount > 1 {
				break
			}
		}

		assert.Equal(t, 1, iterationCount)
	})
```

mxsdev:
I think this should fix #149.

Basically, the `b.isFull` variable is not able to be cancelled by the `read` method, only by the `write` method. I think probably that the buffer should not be set as full anymore once its drained.

### DIFF
--- a/internal/rbuffer/ring_buffer.go
+++ b/internal/rbuffer/ring_buffer.go
@@ -96,6 +96,7 @@ func (b *RingBuffer) read(p []byte) (n int, err error) {
 		copy(p[c1:], b.buf[0:c2])
 	}
 	b.r = (b.r + n) % b.size
+	b.isFull = false
 
 	return n, err
 }
--- a/internal/rbuffer/ring_buffer_test.go
+++ b/internal/rbuffer/ring_buffer_test.go
@@ -55,6 +55,41 @@ func TestRingBuffer(t *testing.T) {
 		assert.Equal(t, buf.size, n)
 		assertRead(t, buf, data[1024-buf.size:])
 	})
+
+	t.Run("ExceedingInputTerminates", func(t *testing.T) {
+		data := append(
+			bytes.Repeat([]byte{'1'}, 1024),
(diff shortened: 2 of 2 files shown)


internal/rbuffer/ring_buffer.go
```diff
@@ -96,6 +96,7 @@ func (b *RingBuffer) read(p []byte) (n int, err error) {
 		copy(p[c1:], b.buf[0:c2])
 	}
 	b.r = (b.r + n) % b.size
+	b.isFull = false
```
mxsdev, line 99: Do we need a check here to make sure that `n == b.size`? @adambabik

adambabik: Yes, that's the root cause I also identified. I think there is one more place to change: https://github.com/stateful/runme/pull/151.

PR #155, opened 2023-02-23
internal/runner: separate env lines by null

### LINKED ISSUE #154: internal/runner: `readEnvFromFile` cannot parse env variables with new lines
The function `readEnvFromFile` naively parses on new-lines, which of course will not support environment variables containing new lines. This is causing truncation for multiline environment variables, for example.

For now I guess we can use the `env -0` flag to separate by null terminators, instead of new lines.

In the future, though, we probably should properly (JSON)-encode environment variables. I imagine this can be wrapped up as part of #148 - probably we could expose a subcommand of `runme` to wrap around something like `godotenv`, which handles marshaling/parsing.

mxsdev:
Fixes #154

### DIFF
--- a/internal/runner/service_test.go
+++ b/internal/runner/service_test.go
@@ -568,7 +568,9 @@ func Test_runnerService(t *testing.T) {
 	t.Run("ExecuteLicenseCat", func(t *testing.T) {
 		t.Parallel()
 
-		session, _ := client.CreateSession(context.Background(), &runnerv1.CreateSessionRequest{})
+		session, err := client.CreateSession(context.Background(), &runnerv1.CreateSessionRequest{})
+		require.NoError(t, err)
+
 		sessionID := session.Session.Id
 
 		stream, err := client.Execute(context.Background())
@@ -582,12 +584,13 @@ func Test_runnerService(t *testing.T) {
 			},
 			SessionId: sessionID,
 		})
-		assert.NoError(t, err)
+		require.NoError(t, err)
 
 		{
 			execResult := make(chan executeResult)
 			go getExecuteResult(stream, execResult)
-			<-execResult
+			result := <-execResult
+			require.EqualValues(t, 0, result.ExitCode)
 		}
 
 		_, _ = client.GetSession(context.Background(), &runnerv1.GetSessionRequest{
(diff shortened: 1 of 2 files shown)


internal/runner/service_test.go
```diff
@@ -568,19 +568,45 @@ func Test_runnerService(t *testing.T) {
 	t.Run("ExecuteLicenseCat", func(t *testing.T) {
 		t.Parallel()
 
+		session, _ := client.CreateSession(context.Background(), &runnerv1.CreateSessionRequest{})
```
adambabik, line 571: No reason not to check the error with `require.NoError`.

internal/runner/service_test.go
```diff
@@ -568,19 +568,45 @@ func Test_runnerService(t *testing.T) {
...
 
+		err = stream.Send(&runnerv1.ExecuteRequest{
+			ProgramName: "bash",
+			Directory:   "../..",
+			Commands: []string{
+				"export LICENSE=$(cat LICENSE)",
+			},
+			SessionId: sessionID,
+		})
+		assert.NoError(t, err)
```
adambabik, line 585: I use `require` when there is no point in moving further if a condition is false. `assert` is used when it's ok to continue.

internal/runner/service_test.go
```diff
@@ -568,19 +568,45 @@ func Test_runnerService(t *testing.T) {
 	t.Run("ExecuteLicenseCat", func(t *testing.T) {
```
adambabik, line 568:
```suggestion
	t.Run("ExecuteMultilineEnvExport", func(t *testing.T) {
```

internal/runner/service_test.go
```diff
@@ -568,19 +568,45 @@ func Test_runnerService(t *testing.T) {
...
+				"export LICENSE=$(cat LICENSE)",
+			},
+			SessionId: sessionID,
+		})
+		assert.NoError(t, err)
+
+		{
+			execResult := make(chan executeResult)
+			go getExecuteResult(stream, execResult)
+			<-execResult
```
adambabik, line 590: It's worth to at least check the exit code with `require.EqualValues(t, 0, result.ExitCode)`.

mxsdev:
@adambabik Good to merge?

adambabik (approved):

PR #164, opened 2023-03-02
Use absolute path for markdown not found error

mxsdev:
On my system, the path returned from path error is relative rather than absolute ¯\\\_(ツ)_/¯ this breaks the tests

Figure absolute is clearer.


mxsdev:
Realized this is because i was using go 1.20. Should note that we'll need to re-open and merge this if we ever update to 1.20.

Issue #163, opened 2023-02-28
internal/cmd/server: Implement health check endpoint


mxsdev:
Would be nice to have a "health check" endpoint - as per @adambabik:

> ... [there is a guideline](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) in the gRPC repo plus [this implementation](https://github.com/grpc-ecosystem/grpc-health-probe) in various languages including [Go](https://pkg.go.dev/google.golang.org/grpc/health/grpc_health_v1?utm_source=godoc).

Would resolve ambiguity as for how clients should determine when server is ready.

mxsdev:
@MusabShakeel576 if you are interested, I think this is a PR you could really help with! 

Basically, if you could implement the health check functionality as described [here](https://github.com/stateful/runme/issues/163#:~:text=there%20is%20a%20guideline), using [this implementation](https://github.com/grpc-ecosystem/grpc-health-probe), that would be helpful - and hopefully not too complicated!

MusabShakeel576:
@mxsdev alright, I'll give it a try and keep you updated with the progress or if I need any help.

PR #165, opened 2023-03-02
internal/cmd/server: health check endpoint

MusabShakeel576:
Adds health check endpoint in gRPC server.

**Issue:**

#163 

**Test:**

- Start the gRPC server `./runme server -a localhost:7890`
- Install [grpc-health-probe](https://github.com/grpc-ecosystem/grpc-health-probe#installation)
- Execute command `grpc_health_probe -addr=localhost:7890`

Response will either be `SERVING` or `NOT SERVING`.

### DIFF
--- a/examples/grpc-client/main.go
+++ b/examples/grpc-client/main.go
@@ -12,7 +12,7 @@ import (
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
-	_ "google.golang.org/grpc/health"
+	healthgrpc "google.golang.org/grpc/health/grpc_health_v1"
 )
 
 var (
@@ -20,13 +20,6 @@ var (
 	file       = flag.String("file", "", "file with content to upper case")
 	resultFile = flag.String("write-result", "-", "path to a result file (default: stdout)")
 )
-
-var serviceConfig = `{
-	"loadBalancingPolicy": "round_robin",
-	"healthCheckConfig": {
-		"serviceName": ""
-	}
-}`
 
 func main() {
 	flag.Parse()
@@ -37,16 +30,24 @@ func main() {
 }
 
 func run() error {
-	options := []grpc.DialOption{
-		grpc.WithTransportCredentials(insecure.NewCredentials()),
-		grpc.WithDefaultServiceConfig(serviceConfig),
-	}
-
-	conn, err := grpc.Dial(*addr, options...)
+	conn, err := grpc.Dial(*addr, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	if err != nil {
 		return errors.Wrap(err, "failed to connect")
 	}
 	defer conn.Close()
+
+	healthClient := healthgrpc.NewHealthClient(conn)
+
+	resp, err := healthClient.Check(context.Background(), &healthgrpc.HealthCheckRequest{})
+	if err != nil {
+		return errors.Wrap(err, "failed to check health")
+	}
+
+	if resp.Status != healthgrpc.HealthCheckResponse_SERVING {
+		return errors.Errorf("service status: %v", resp.Status)
+	}
+
+	log.Printf("service is SERVING")
 
 	client := runnerv1.NewRunnerServiceClient(conn)
 
--- a/examples/grpc-client/main.go
+++ b/examples/grpc-client/main.go
@@ -12,6 +12,7 @@ import (
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
+	_ "google.golang.org/grpc/health"
 )
 
 var (
@@ -19,6 +20,13 @@ var (
 	file       = flag.String("file", "", "file with content to upper case")
 	resultFile = flag.String("write-result", "-", "path to a result file (default: stdout)")
 )
+
+var serviceConfig = `{
+	"loadBalancingPolicy": "round_robin",
+	"healthCheckConfig": {
+		"serviceName": ""
+	}
+}`
 
 func main() {
 	flag.Parse()
@@ -29,7 +37,12 @@ func main() {
 }
 
 func run() error {
-	conn, err := grpc.Dial(*addr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	options := []grpc.DialOption{
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithDefaultServiceConfig(serviceConfig),
+	}
+
+	conn, err := grpc.Dial(*addr, options...)
 	if err != nil {
 		return errors.Wrap(err, "failed to connect")
 	}
--- a/examples/grpc-client/main.go
+++ b/examples/grpc-client/main.go
@@ -47,8 +47,6 @@ func run() error {
 		return errors.Errorf("service status: %v", resp.Status)
 	}
 
-	log.Printf("service is SERVING")
-
 	client := runnerv1.NewRunnerServiceClient(conn)
 
 	g, ctx := errgroup.WithContext(context.Background())
--- a/internal/cmd/server.go
+++ b/internal/cmd/server.go
@@ -30,8 +30,6 @@ func serverCmd() *cobra.Command {
 	const (
 		defaultSocketAddr = "unix:///var/run/runme.sock"
 		defaultLocalAddr  = "localhost:7890"
-		// Empty string represents the health of the system.
-		system = ""
 	)
 
 	var (
@@ -142,7 +140,7 @@ The kernel is used to run long running processes like shells and interacting wit
 			healthcheck := health.NewServer()
 			healthgrpc.RegisterHealthServer(server, healthcheck)
 			// Setting SERVING for the whole system.
-			healthcheck.SetServingStatus(system, healthgrpc.HealthCheckResponse_SERVING)
+			healthcheck.SetServingStatus("", healthgrpc.HealthCheckResponse_SERVING)
 
 			reflection.Register(server)
 			return server.Serve(lis)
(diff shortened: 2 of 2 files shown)


MusabShakeel576:
@mxsdev this PR is a WIP, I didn't Test it myself, nor write the test cases.

I created this PR to get your input on what should I use in place of [parser and runner](https://github.com/stateful/runme/pull/165/files#diff-9d0d77aad9fa41703b1eaea210204e95ee6a03288167890d7ef30bd1a1ba55f8R157-R158), is there any method/function in parser and runner files which returns whether the parser and the runner is running/healthy?

mxsdev:
@MusabShakeel576 

Thanks for working on this! Looks good so far :)

Currently, neither the parser nor the runner block on anything for setup, nor have any globally persistent state. Basically, they are 100% initiated by the client. I think that they will work as long as the parent gRPC server is ready and listening.

So, I guess we can just assume they are operational and flatly return `SERVING` for now, noting that we might need to add more logic to the health check if things get more complex. cc @adambabik 

I will take more of a look/do tests first thing tomorrow. And thanks again for contributing!

MusabShakeel576:
@mxsdev OK, I'll just return `SERVING`, and add the test cases.

adambabik:
> So, I guess we can just assume they are operational and flatly return `SERVING` for now, noting that we might need to add more logic to the health check if things get more complex. cc @adambabik

+1 Returning a healthy status without any additional logic is totally fine and common practice if the service has no dependencies.

MusabShakeel576:
@mxsdev I searched into the repo and didn't find the unit test file for `/internal/cmd/server.go`, isn't there the test file for this and other few files?

Btw I think I also need to update [/examples/grpc-client](https://github.com/stateful/runme/blob/main/examples/grpc-client/main.go) to add the health check, or is it fine?

mxsdev:
@MusabShakeel576 

We currently don't have a unit for `server.go`, since we would need to mock a lot of things (`cobra`, `grpcServer`, etc). I think we'd rather pull an integration test at some point.

Currently we are using `examples/grpc-client` as a kind of integration test. So yeah, if you would be willing to add a check in there for health (return an error if health check is not serving or server is down) that would be great!

MusabShakeel576:
@mxsdev alright.

Sure, I'll try to add the check in `examples/grpc-client`.

MusabShakeel576:
@mxsdev implemented check in `examples/grpc-client`. But there is an error occurring while running this file `error: failed to send initial request: EOF`

I think the error is not causing by my health check code as I tried removing the code and running it, and the error still exists. So maybe the error was there already or maybe I'm using the wrong commands to run it. Following are the commands I'm using:

- To start the example/grpc-client, tried both commands:
  - `go run examples/grpc-client/main.go`
  - `go run examples/grpc-client/main.go --addr=localhost:7890`
- To start the server, tried both:
  - `go run . server -a=localhost:7890 --dev=true --runner=true`
  - `./runme server -a localhost:7890 --runner=true`

Do you know why this error is occurring?

mxsdev:
@MusabShakeel576 

Sorry, I should've mentioned: the `grpc-client` requires a `--file` argument - it should return the contents of the file in upper case!

By the way, as for the code you pushed, I don't think that the `serviceConfig` is necessary, since as I understand this is usually for multiple concurrent connections/addresses. Instead, I think it would be best to import  `"google.golang.org/grpc/health/grpc_health_v1"`, register a new health client, and do a simple check directly, making sure it didn't error and the response has code `SERVING`. 

Let me know if there are any more issues!

MusabShakeel576:
@mxsdev no problem.

`serviceConfig` does this automatically. If you try to change [SERVING](https://github.com/stateful/runme/blob/9ad6d5def132ca2080a226ebd1ef6ad45a75820d/internal/cmd/server.go#L145) to `NOT_SERVING`, it'll throw the error and will be caught at [grpc/client Execute()](https://github.com/stateful/runme/blob/9ad6d5def132ca2080a226ebd1ef6ad45a75820d/examples/grpc-client/main.go#L57). Correct me if I'm wrong.

mxsdev:
@MusabShakeel576 

From my tests, I get the following log:

```
2023/03/03 14:25:12 ERROR: [core] [Channel #1 SubChannel #2] Subchannel health check is unimplemented at server side, thus health check is disabled
```

But this error does not come from `client.Execute()`, it is from an internal logger. So, I think a manual check is necessary to give a non-zero exit code from the `grpc-client`.
 
Lmk if you have different results than I did :)

MusabShakeel576:
@mxsdev hmm, which command do you use to run it. I'm receiving the following result:
```
error: failed to call Execute(): rpc error: code = Unavailable desc = last connection error: connection active but health check failed. status=NOT_SERVING
exit status 1
```

mxsdev:
@MusabShakeel576 

Your error message happens when the status endpoint exists but is not `SERVING`. My error message happens when the server doesn't implement the endpoint at all.

I think I would just prefer to have one explicit check, to deal with any other cases we don't know of, and so that it's extremely clear to a future developer on the project why the client fails due to to health check. Better to be explicit than implicit, no?

No problem keeping the existing `serviceConfig` stuff though 👍

MusabShakeel576:
@mxsdev sure, I'll try to register a new health client.

MusabShakeel576:
@mxsdev implemented the health client, lmk if the message in your terminal/log is looking fine.

examples/grpc-client/main.go
```diff
@@ -37,17 +30,25 @@ func main() {
...
+	resp, err := healthClient.Check(context.Background(), &healthgrpc.HealthCheckRequest{})
+	if err != nil {
+		return errors.Wrap(err, "failed to check health")
+	}
+
+	if resp.Status != healthgrpc.HealthCheckResponse_SERVING {
+		return errors.Errorf("service status: %v", resp.Status)
+	}
+
+	log.Printf("service is SERVING")
```
mxsdev, line 50: No need for log here

internal/cmd/server.go
```diff
@@ -134,6 +138,12 @@ The kernel is used to run long running processes like shells and interacting wit
 			if enableRunner {
 				runnerv1.RegisterRunnerServiceServer(server, runner.NewRunnerService(logger))
 			}
+
+			healthcheck := health.NewServer()
+			healthgrpc.RegisterHealthServer(server, healthcheck)
+			// Setting SERVING for the whole system.
+			healthcheck.SetServingStatus(system, healthgrpc.HealthCheckResponse_SERVING)
```
mxsdev, line 145:
```suggestion
			healthcheck.SetServingStatus("", healthgrpc.HealthCheckResponse_SERVING)
```

Would prefer this inlined if that's alright!

mxsdev (changes requested):
Looks great! I'll test this further and get back to you on Monday 👍

mxsdev:
@MusabShakeel576 

Hey, sorry for getting sidetracked!

I did look into this on Monday and yesterday, the only thing a this point is that we'll need is typescript definitions for the health protos. I was trying to figure out how to get them to bundle as part of the `proto/generate` script from [the official health.proto files](https://github.com/grpc/grpc/blob/master/src/proto/grpc/health/v1/health.proto)

The code is fine, I just need to figure out how to get that to work and then I'm happy to merge 👍

MusabShakeel576:
@mxsdev no worries.

Sure, take your time.

mxsdev:
@adambabik sorry to bother you again, but I'm trying to have the `health.proto` file bundled as part of the buf package. I see you imported this:
 
`"google/protobuf/wrappers.proto"`

But I don't know where that file is being resolved from? If you could just point me in the right direction here that would be massively helpful :)

adambabik:
@mxsdev `google/protobuf/wrappers.proto` is something called a well-known type and `buf` handles it transparently. If you install `protoc` manually it is included in `/path/to/protoc/include` or similar.

`health.proto` source code is located in [grpc/grpc](https://github.com/grpc/grpc/blob/febed5121a82ff05a8bf29003c0935507c45a530/src/proto/grpc/health/v1/health.proto), but it is also present in [Buf Registry](https://buf.build/grpc/grpc/file/main:grpc/health/v1/health.proto). What you want to do is basically generate stubs only for that file. I imagine it should be possible using `buf` but I don't know how. Maybe try asking on their Slack or other community forum? I would love to learn how to do it, too :)

Alternatively (temporarily), you can copy this proto file into `./internal/api/grpc/health/v1/health.proto` and the stubs should be generated as they are for the existing proto files.

mxsdev:
@adambabik Oh my god, I did not realize it was on the buf registry! I tried to look under the `google` org and could not find it.

Thank you !!!!

mxsdev:
@adambabik 

Asked on the slack and there is an `--include-imports` flag on `buf generate`, which bundles all imports. That pretty much solves the problem, though I have no idea if they will still be included as part of the bundle on `buf push`?

mxsdev (approved):
OK, figured it out for real - [see here for client implementation](https://github.com/stateful/vscode-runme/pull/291).

Turns out buf generates typescript definitions not just for projects that have it enabled, but for all projects, including the [grpc healthcheck module](https://buf.build/gen/npm/v1/@buf/grpc_grpc.community_timostamm-protobuf-ts).

So, I'm considering this PR good to merge. Will merge later for sure 👍

adambabik:
> Asked on the slack and there is an --include-imports flag on buf generate, which bundles all imports.

That's nice 👍 Interesting if it's possible to limit the imports using `--exclude-path` 🤔 

> OK, figured it out for real - https://github.com/stateful/vscode-runme/pull/291.

> Turns out buf generates typescript definitions not just for projects that have it enabled, but for all projects, including the [grpc healthcheck module](https://buf.build/gen/npm/v1/@buf/grpc_grpc.community_timostamm-protobuf-ts).

That's even nicer! It's exactly the same what we do for runme's protos, just pointing at the different package. 👏

PR #168, opened 2023-03-06
internal/runner: pty sizing

### LINKED ISSUE #166: internal/runner: Interactive mode failing in some interactive scripts
`deployctl deploy` is not working in interactive mode. What we get instead is an endless stream of xterm sequences.

What I'm thinking here is that there is an expectation to "receive" something in response to these sequences. More investigation is needed.

### LINKED ISSUE #167: internal/runner: set pseudoterminal sizing
Pseudoterminal size needs to be set for some scripts to not fail. 

Should definitely be a field in the `ExecuteRequest` struct so it's user settable.

mxsdev:
Adds a new field to `ExecuteRequest` which allows pseudo-terminal sizing to be set. Should do nothing in non-tty mode.

Fixes #167 
Fixes #166


mxsdev:
Looks like testing PTY doesn't work in CI :/ 

@adambabik any advice on this?

sourishkrout (approved):
👍 LGTM. As discussed, it's ok to temporarily not run the respective tests in C/I.

sourishkrout:
> Looks like testing PTY doesn't work in CI :/ getting:
> 
> ```
> failed to copy from pty to stdout; handled EIO
> ```

@mxsdev please create a follow-up ticket so we plan to take a look later.

mxsdev:
See #169 for follow-up ticket.

internal/runner/service.go
```diff
@@ -272,6 +274,13 @@ func (r *runnerService) Execute(srv runnerv1.RunnerService_ExecuteServer) error
 					return
 				}
 			}
+
+			// only update winsize when field is explicitly set
+			if req.ProtoReflect().Has(
```
adambabik, line 279: Alternatively, you could omit `optional` in the proto definition and validate the `Winsize` message, that is check if all fields are greater than zero (unless that's a valid scenario but I doubt it). If it's `nil` or a zero value, then `pty.Winsize{Cols: 80, Rows: 24}` would be the "normal" size. Some applications actually might expect 80 columns and 24 rows a minimal size (to double check).

internal/runner/command.go
```diff
@@ -75,6 +75,8 @@ type commandConfig struct {
 	Script   string
 
 	Logger *zap.Logger
+
+	Winsize *pty.Winsize
```
adambabik, line 79: Nitpick: it's pty-related so I would group it with the fields starting with `Tty bool`.

internal/runner/service_test.go
```diff
@@ -621,6 +621,96 @@ func Test_runnerService(t *testing.T) {
 		assert.Equal(t, "LICENSE: "+string(expected), string(result.Stdout))
 		assert.EqualValues(t, 0, result.ExitCode)
 	})
+
+	if _, ok := os.LookupEnv("GITHUB_ACTIONS"); !ok {
+		t.Run("ExecuteWinsizeDefault", func(t *testing.T) {
```
adambabik, line 626: This one might be moved to `command_test.go` as it does not test any gRPC service functionality unlike the two below.

adambabik:
> Looks like testing PTY doesn't work in CI :/ getting:
> 
> ```
> failed to copy from pty to stdout; handled EIO
> ```
> 
> @adambabik any advice on this?

I do remember seeing this error in a GH Actions job, but unfortunately, I don't quite remember the root cause. There are a few tests in `command_test.go` which do open a pty and they work fine so it should work overall. Maybe it's ok to open it but `ioctl()` syscalls fail? It would be nice to see if `pty.Setsize()` returns an error; currently it's ignored in `command.setWinsize`.

internal/runner/service.go
mxsdev: The reason I need it as optional is so that I'm not calling `pty.Setsize` when the user doesn't want to change the size. Otherwise the size would be set every time the user sends input, no?

internal/runner/command.go
mxsdev: Oh yeah, that's fair!

internal/runner/service_test.go
mxsdev: Yep that makes sense!

mxsdev:
@adambabik 

Thanks for your input, I will continue to investigate the issue.

I'll also create a follow-up PR to incorporate some of your feedback :)

internal/runner/service.go
adambabik: That's correct unless you cache the size you set at the startup/most recently and do not change it if the new size is the same (if it's a zero value = the default size).

PR #170, opened 2023-03-07
internal/runner: PTY improvements

mxsdev:
- [x] Fix CI for PTY commands
 - [x] Implement `ExecuteWinsizeDefault` as test on `command`
 - [x] Refactor position of `Winsize` to be with other TTY-related config options

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,8 @@
 name: ci
 on:
   push:
+    branches:
+      - main
     paths-ignore:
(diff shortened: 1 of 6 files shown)


mxsdev:
I used the `-T` flag rather than modifying `ci.yml` to ensure consistent behavior across al runners, since (presumably) some terminal settings might have "wider" characters, skewing the results of the command.

mxsdev:
@adambabik Seems reasonable?

Issue #159, opened 2023-02-28
Runme fails in CI


christian-bromann:
When trying to run Runme in CI (e.g. GitHub Action running Ubuntu) the commands fail:

```
Run runme --filename DEVELOPMENT.md run source-env
  runme --filename DEVELOPMENT.md run source-env
  shell: /usr/bin/bash -e {0}
  env:
    ANDROID_HOME: /usr/local/lib/android/sdk
    ANDROID_SDK_ROOT: /usr/local/lib/android/sdk
/bin/sh: 1: set: Illegal option -o pipefail
failed to run command "source-env": exit status 2
Error: Process completed with exit code 1.
```

MusabShakeel576:
Hi @christian-bromann, at which file is this error occurring? [Recent GitHub Action workflows](https://github.com/stateful/runme/actions) looking fine to me.

Btw just by looking at this error, looks like the shell isn't set to `bash` as the error happening in the `sh`, and only bash supports the `pipefail` option, maybe change the shebang line at the beginning of the script to:

```bash
#!/usr/bin/env bash
```

mxsdev:
Should be solved by #171

Long term it would be nice to harden this, see #172

PR #171, opened 2023-03-07
internal/runner: set command options only in supported shells

mxsdev:
Solves #159 

~~Would like write tests for `ShellFromShellPath` function before merge~~


mxsdev:
@sourishkrout You ok with me merging on this? @degrammer had success with it for fixing CI

PR #174, opened 2023-03-08
internal/cmd/run: Integrate with gRPC runner

mxsdev:
Implements the following for the `run` command:

 - `--server` flag, which when provided uses the gRPC runner at the given address
 - `--session` flag, which when provided uses the given session ID. If this flag is provided, the session will not be removed on program exit.

### DIFF
--- a/examples/grpc-client/main.go
+++ b/examples/grpc-client/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"flag"
+	"fmt"
 	"io"
 	"log"
 	"os"
@@ -15,9 +16,11 @@ import (
 )
 
 var (
-	addr       = flag.String("addr", "127.0.0.1:7890", "the address to connect to")
-	file       = flag.String("file", "", "file with content to upper case")
-	resultFile = flag.String("write-result", "-", "path to a result file (default: stdout)")
+	addr          = flag.String("addr", "127.0.0.1:7890", "the address to connect to")
+	file          = flag.String("file", "", "file with content to upper case")
+	resultFile    = flag.String("write-result", "-", "path to a result file (default: stdout)")
+	createSession = flag.Bool("create-session", false, "create a new session")
+	deleteSession = flag.String("delete-session", "", "delete the given session")
 )
 
 func main() {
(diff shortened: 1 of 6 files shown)


examples/grpc-client/main.go
```diff
@@ -37,6 +40,34 @@ func run() error {
 
 	client := runnerv1.NewRunnerServiceClient(conn)
 
+	if *createSession {
+		resp, err := client.CreateSession(context.Background(), &runnerv1.CreateSessionRequest{
+			Envs: os.Environ(),
+		})
+		if err != nil || resp.Session == nil {
```
adambabik, line 47: The convention is that if `err` is `nil` then `resp.Session` is never `nil` so typically you don't need to check it.

examples/grpc-client/main.go
```diff
@@ -37,6 +40,34 @@ func run() error {
...
+		id := *deleteSession
+
+		_, err := client.DeleteSession(context.Background(), &runnerv1.DeleteSessionRequest{
+			Id: id,
+		})
+		if err != nil {
+			return err
+		}
+
+		_, _ = fmt.Printf("Successfully deleted session \"%s\"", id)
```
adambabik, line 66:
```suggestion
		_, _ = fmt.Printf("Successfully deleted session %q", id)
```

`%q` will quote the value.

sourishkrout:
@mxsdev how do I initiate a session from the CLI? Also, I'd suggest to mark the `--session` flag hidden until we decide it's stable/final

sourishkrout:
Ah, I passed an empty sessionID using `--session ""`. However, not ideal. Still need to figure out what the right UX would be here.

sourishkrout:
Question, @mxsdev, does the `LocalRunner` work precisely the same as the RemoteRunner's GRPC interface? In other words, are they using the same code path except not going through a remote interface?

mxsdev:
@sourishkrout 

> @mxsdev how do I initiate a session from the CLI?

Oh yeah, should've mentioned: I added a `--create-session` to the `./examples/grpc-client` module.

> I'd suggest to mark the --session flag hidden until we decide it's stable/final

Ok, will do!

> Ah, I passed an empty sessionID using --session "". However, not ideal. Still need to figure out what the right UX would be here.

Yeah, I was planning on expanding this once #176 is done. I think that will enable automatic session creation to be done more smoothly, with a cli flag like `--auto-session` or `--session-strategy=recent`, etc.

> Question, @mxsdev, does the LocalRunner work precisely the same as the RemoteRunner's GRPC interface? In other words, are they using the same code path except not going through a remote interface?

Pretty much (there are some minor differences that could be ironed out). This, I think, is what Babik meant when he said he was "uncomfortable with the amount of indirection" going on, since it's basically like `RemoteRunner` -> `gRPC` -> `Execute` loop -> `LocalRunner`.

sourishkrout:
> For the most part, yes. This, I think, is what Babik meant when he said he was "uncomfortable with the amount of indirection" going on, since it's basically like `RemoteRunner` -> `gRPC` -> `Execute` loop -> `LocalRunner`.
> 
> One big difference though is environment handling, and command preparation in general. IDK if you can set environment variables from a child process reliably for the parent process, so persistence would be have to done with something like a file for `LocalRunner`, which it is not currently.

I remember we talked about this with Babik. Wanted to doublecheck. As we use this more, we likely get a better sense what's required.

Issue #180, opened 2023-03-08
Unify defaults "everywhere"


sourishkrout:
Annotations defaults can't be different between `cli` & `extension`:

Examples:
1. The extension's `interactive` default is `true` whereas in the `CLI` it's false.
2. The TUI will block, expecting non-interactive behavior, on cells that aren't flagged interactive.

mxsdev:
I'm closing this for now since I've made `interactive` the default and defaulted to shell on unknown cells in `runme`. If another default issue comes up, I'm happy to fix it. But I don't know of anything else that needs immediate attention.

PR #183, opened 2023-03-09
Default unknown executors as shell

mxsdev:
Addresses #180, creating parity with `vscode-runme`


sourishkrout (changes requested):
See comment.

PR #184, opened 2023-03-10
internal/runner: add "most recent" session strategy

mxsdev:


internal/api/runme/runner/v1/runner.proto
```diff
@@ -57,6 +57,16 @@ enum ExecuteStop {
...
 }
 
+// strategy for selecting a session in an initial execute request
+enum SessionStrategy {
+  // uses session_id field to determine session
+  // if none is present, a new session is created
+  SESSION_STRATEGY_UNSPECIFIED = 0;
+  // uses most recently used session on the grpc server
+  // if there is none, a new one is created
+  SESSION_STRATEGY_MOST_RECENT = 1;
```
adambabik, line 67: It might bite you in the future. I am not sure what's the reason for doing it but if it's to simplify `runme run`, I would probably advise something less magical. For example:

```sh
RUNME_SESSION_ID=$(runme session new --server=$SERVER)
runme run --session-id=$RUNME_SESSION_ID
# or have "runme run" to recognize RUNME_SESSION_ID env
```

`runme run` could also recognize family of the `RUNME_*` environment variables so that `--server` is not needed to be constantly repeated, etc.

mxsdev: > It might bite you in the future

How so? Is it a security concern?

We're mainly doing this to simplify commands for CI/testable documentation, i.e. controlled environments. We feel like if you have to explicitly create a new session in that case it's a bit much.

For users in a user environment, or in the VSCode extension, we would much rather manage sessions explicitly in the way you've described.

Does that make sense?

adambabik: It introduces more ambiguity and makes it easier to mistakenly run commands in a session they are not intended to. But I don't know the whole design so I might exaggerate.

Issue #188, opened 2023-03-15
runme runs a different command than what I have in the README


mikeage:
Given  the following minimal README.md 
```
```bash
aws ec2 run-instances \
    --no-cli-pager \
    --image-id ami-0fec2c2e2017f4e7b \ 
    --instance-type g4dn.xlarge \
    --key-name FooKey \
    --security-group-ids sg-0ba8468ab13683325 \
    --block-device-mappings 'DeviceName=/dev/xvda,Ebs={VolumeSize=30,VolumeType=gp3}' \
    --user-data file://setup.sh \
    --tag-specifications 'ResourceType=spot-instances-request,Tags=[{Key=creator,Value=stable-diffusion-aws}]' \
    --instance-market-options 'MarketType=spot,SpotOptions={MaxPrice=0.20,SpotInstanceType=persistent,InstanceInterruptionBehavior=stop}'
\``` # Ignore the \ here, didn't know how else to get it to look correct in MD
```

If I run the first section, I see:

```
$ runme --filename README.broken.md
runme 0.6.5

> aws-ec2 bash
    aws ec2 run-instances \ (...)


usage: aws [options] <command> <subcommand> [<subcommand> ...] [parameters]
To see help text, you can run:

  aws help
  aws <command> help
  aws <command> <subcommand> help

Unknown options: MarketType=spot,SpotOptions=InstanceInterruptionBehavior=stop, MarketType=spot,SpotOptions=SpotInstanceType=persistent

exit code: 252
```

By adding a `set -x`, it looks like it might actually dropping a single quote, but I know that debugging quotes is always a pain (this screenshot is after I hit space to expand):

```
$ runme --filename README.broken.md
runme 0.6.5

> set-x bash
    set -x
    aws ec2 run-instances \
    --no-cli-pager \
    --image-id ami-0fec2c2e2017f4e7b \
    --instance-type g4dn.xlarge \
    --key-name FooKey \
    --security-group-ids sg-0ba8468ab13683325 \
    --block-device-mappings 'DeviceName=/dev/xvda,Ebs={VolumeSize=30,VolumeType=gp3}' \
    --user-data file://setup.sh \
    --tag-specifications 'ResourceType=spot-instances-request,Tags=[{Key=creator,Value=stable-diffusion-aws}]' \
    --instance-market-options 'MarketType=spot,SpotOptions={MaxPrice=0.20,SpotInstanceType=persistent,InstanceInterruptionBehavior=stop}'

[...]

PR #189, opened 2023-03-15
Simplify command normalization

mxsdev:
- Removes shlex and essentially all forms of command normalization; commands are joined with new lines rather than semicolons
 - Try to ignore comments when auto-generating name ~~(**needs tests before merge**)~~

The reason for removing command normalization is that it should be handled by bash already. The `Args` field of `exec.Command` can take new lines, and bash will deal with backslashes on its own. Also, new lines appear to act the same as semicolons for most purposes, at least in `bash`.

Attempting to do normalization on our own is causing issues for users, such as #188 

I suppose it might not be handled by some shells like `sh`, but that does not mean that we should attempt to do normalization ourselves.

cc @adambabik

### DIFF
--- a/internal/runner/shell.go
+++ b/internal/runner/shell.go
@@ -132,7 +132,7 @@ func PrepareScriptFromCommands(cmds []string, shell string) string {
 func prepareScriptFromCommands(cmds []string, shell string) string {
 	var b strings.Builder
 
-	writeShellPrefix(shell, &b)
+	_, _ = b.WriteString(getShellOptions(shell))
 
 	for _, cmd := range cmds {
 		_, _ = b.WriteString(cmd)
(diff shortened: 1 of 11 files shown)


internal/runner/shell.go
```diff
@@ -133,58 +132,36 @@ func PrepareScriptFromCommands(cmds []string, shell string) string {
...
 
-	_, _ = b.WriteString("set -e -o pipefail;")
+	writeShellPrefix(shell, &b)
 	_, _ = b.WriteString(script)
 	_, _ = b.WriteRune('\n')
 
 	return b.String()
 }
+
+func writeShellPrefix(shell string, b *strings.Builder) {
```
adambabik, line 157: Nit:

```suggestion
func getShellOptions(shell string) string {
```

internal/shell/format.go
```diff
@@ -0,0 +1,28 @@
+package shell
+
+import "strings"
+
+func StripComments(lines []string) (ret []string) {
+	for _, l := range lines {
+		l = strings.TrimSpace(l)
+
+		if !strings.HasPrefix(l, "#") {
```
adambabik, line 9: What about a situations like:
```sh
echo abc |    # normal comment OK here
  tr a-z A-Z | # another normal comment OK here
```

adambabik (commented):
> I suppose it might not be handled by some shells like sh, but that does not mean that we should attempt to do normalization ourselves.

I don't quite remember why we went with concatenating lines with  `;`. I think some initial commands we tried might have suggested that approach. But if it works without this approach then even better.

At least in a terminal, `sh` also handles new lines well:

```sh
sh -c "echo abc |
tr a-z A-Z"
```

Question is whether there is any assistance from the terminal and shell in which I executed that, e.g. converting `\n` to `\r\n`. I think the easiest to test it is to run tests with and without tty.

mxsdev:
@adambabik 

Having no issues with the pipe new line, either interactive or non-interactive.

I guess bash syntax just supports new lines after pipes as part of its language parser - that was something I did not know.

internal/shell/format.go
mxsdev: Yeah, good point

PR #190, opened 2023-03-15
internal/runner: Use `runme env dump` to dump environment

mxsdev:
Introduces new hidden sub-command `runme env dump` which acts exactly like `env -0`.

To find the current executable, we use `os.Executable()`. For tests, we assume the existence of a `runme` binary in the project root and use that. Note that this means that for tests you should build before testing - this has been added to `make test`.

Before merge I'd like to add tests against `executable.go` as well as the new subcommand.


internal/cmd/common.go
```diff
@@ -185,7 +185,7 @@ func getDefaultConfigHome() string {
 	// TODO(adamb): switch to os.UserConfigDir()
 	dir, err := os.UserHomeDir()
 	if err != nil {
-		panic(err)
+		dir = os.TempDir()
```
adambabik, line 188: What's the purpose of that? It's pretty odd as each `stateful` command requiring it would start with clean slate.

internal/cmd/environment.go
```diff
@@ -0,0 +1,43 @@
...
+	return &cmd
+}
+
+func environmentDumpCmd() *cobra.Command {
+	cmd := cobra.Command{
+		Use:   "dump",
+		Short: "Dump environment variables to stdout",
+		Long:  "Dumps all environment variables to stdout as a list of K=V separated by null terminators",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			dumped := strings.Join(os.Environ(), "\000")
```
adambabik, line 32: Nit: much more common:

```suggestion
			dumped := strings.Join(os.Environ(), "\x00")
```

Makefile
```diff
@@ -22,6 +22,7 @@ wasm:
 
 .PHONY: test
 test:
+	@TZ=UTC go build
```
adambabik, line 25: Use `test: build` instead ([more](https://web.mit.edu/gnu/doc/html/make_2.html#SEC5)).

internal/executable/executable.go
```diff
@@ -0,0 +1,45 @@
+package executable
+
+import (
+	"os"
+	"path/filepath"
+	"runtime"
+)
+
+var runmePath string
```
adambabik, line 9: More error-prone approach might be to define a function that creates a dump env command and replace it when running tests.

```go
// In internal/runner.
var DumpCmd string = getDumpCmd

func getDumpCmd() string {
    path, _ := os.Executable()
    ...
    // return `/path/to/runme env dump`
}

// In internal/runner/*_test.go
func init() {
  DumpCmd = `echo "STUB_ENV1=val1\x00STUB_ENV2=val2"`
}
```

mxsdev: Yep, that's a significantly better way to do it.

Makefile
mxsdev: Oh yeah, good point 👍

internal/cmd/common.go
mxsdev: This is because this function makes it so that you have to have a `$HOME` or else the program fails.

Since the gRPC runner has a minimal user environment by default, we can't assume that `$HOME` exists. I figured this is a reasonable approach for now, and maybe we can come up with a better solution later.

internal/cmd/environment.go
mxsdev: Yep, thanks for catching this!

PR #196, opened 2023-03-20
allow copying from tui with "y" or "c"

mxsdev:
Use "y" or "c" to copy command contents to clipboard.

Maybe we should escape new lines with `\`?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.19
 require (
 	github.com/Masterminds/semver/v3 v3.1.1
 	github.com/Microsoft/go-winio v0.6.0
+	github.com/atotto/clipboard v0.1.4
 	github.com/bufbuild/connect-go v1.4.1
 	github.com/bufbuild/connect-grpchealth-go v1.0.0
 	github.com/bufbuild/connect-grpcreflect-go v1.0.0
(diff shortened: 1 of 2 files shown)


mikeage (approved):
thanks! Perfect :-)

mikeage:
hi, is it possible to merge this? :-)

mxsdev:
@mikeage Hey, before merging I'd probably want to add it to the command list. But the command list is getting kinda large, so I probably would want to hide it behind a "show more" button

@sourishkrout thoughts on having a "silent command" that's not listed? Is that alright or poor UX in your mind?

sourishkrout:
> @mikeage Hey, before merging I'd probably want to add it to the command list. But the command list is getting kinda large, so I probably would want to hide it behind a "show more" button/menu
> 
> @sourishkrout thoughts on having a "silent command" that's not listed? Is that alright or poor UX in your mind?

looking at this now. we can probably turn this around fast.

sourishkrout:
@mxsdev, how about we ship this as is and create a follow-up ticket for the UX?

I've seen bubbletea TUI's having an intermediate "help" screen that shows all the shortcuts to not overflow the bottom line.

mxsdev:
See #204

sourishkrout:
@mikeage we shipped `runme` v1 with these shortcuts. If you're on homebrew, it should be a matter of running upgrade. Have fun, and let us know if there's anything else you need.

PR #205, opened 2023-03-30
internal/runner: add `-i` flag if tty present

mxsdev:
Required for loading bash profiles properly. Not required in zsh though.

### DIFF
--- a/internal/runner/command.go
+++ b/internal/runner/command.go
@@ -125,7 +125,13 @@ func newCommand(cfg *commandConfig) (*command, error) {
 
 		_, _ = script.WriteString(fmt.Sprintf("%s > %s\n", dumpCmd, filepath.Join(envStorePath, envEndFileName)))
 
-		extraArgs = []string{"-c", script.String()}
+		extraArgs = []string{}
+
+		if cfg.Tty {
+			extraArgs = append(extraArgs, "-i")
+		}
+
+		extraArgs = append(extraArgs, "-c", script.String())
 	}
 
 	session := cfg.Session


duvanmonsa (approved):
LGTM

sourishkrout:
Works now in `interactive` mode:
image

sourishkrout (approved):
Tested on codespace's VM

adambabik:
For the sake of completeness, `-i` enables interactive mode. Because we use `-c`, it needs to be explicitly enabled if desired. There are a few more things that are going to happen. From `man bash`:

> When bash is invoked as an interactive login shell, or as a non-interactive shell with the --login option, it first reads and executes commands from the file /etc/profile, if that file exists.  After reading that file, it looks for ~/.bash_profile, ~/.bash_login, and ~/.profile, in that order, and reads and executes commands from the first one that exists and is readable.  The --noprofile option may be used when the shell is started to inhibit this behavior.
> 
> When an interactive login shell exits, or a non-interactive login shell executes the exit builtin command, bash reads and executes commands from the file ~/.bash_logout, if it exists.
>
> When an interactive shell that is not a login shell is started, bash reads and executes commands from ~/.bashrc, if that file exists.  This may be inhibited by using the --norc option.  The --rcfile file option will force bash to read and execute commands from file instead of ~/.bashrc.

`zsh` also has `-i` which forces it to be interactive. However, `-c` might not disable it as it is for bash. Also, zsh uses different files like `/etc/zlogin` and `$ZDOTDIR/.zlogin`.

mxsdev:
Added your comment to the PR description ! @adambabik

PR #207, opened 2023-04-07
Set Stateful as license owner

christian-bromann:
I just discovered this. We should put our name on this license.


sourishkrout (approved):
Makes perfect sense. Can you do 2022-2023 for now please, @christian-bromann.

Otherwise LGTM 👍

Issue #123, opened 2023-02-10
Improve and better integrate changes from `suggest`


mxsdev:
Here is a compilation of things we may want to do before releasing the `suggest branch` feature.

**Higher Priority / Simple to Implement**:

 - API being down should have a better UX - currently, we get: `failed to run command: Post "f/graphql": Post "/auth/cli": unsupported protocol scheme`
   - Might want to point users to the Discord/GH Issue page here if this happens?
 - Invalid auth token should have a better UX - currently, we get: `failed to run command: returned error 401 Unauthorized: {"message":"JWT error","code":"unauthorized"}`
   - We should probably just redirect to the auth flow and get a new token in this case
 - `getDefaultConfigHome`, used for API token storage, should be configurable and should probably use `os.UserConfigDir` instead of `os.UserHomeDir`

@sourishkrout I didn't talk about the API here, since I'm assuming its good on your end.

I'll add anything else brought up to this list going forward 👍

adambabik:
I agree that most of the improvements at this moment would be around the UX and messages.

> getDefaultConfigHome, used for API token storage, should be configurable and should probably use os.UserConfigDir instead of os.UserHomeDir

I think it can be worth doing sooner than later. Otherwise, we will need to write some code for backward compatibility. Also, changing this will trigger re-authentication for everyone, even those using Stateful's CLI. Please note that currently, `runme` and `stateful` CLIs rely on the same authentication token stored in the same place.

sourishkrout:
> Please note that currently, `runme` and `stateful` CLIs rely on the same authentication token stored in the same place.

That's high up on the priority list 👍 .

duvanmonsa:
resolved item 1, 2 in this PR https://github.com/stateful/runme/pull/209

sourishkrout:
Done.

PR #211, opened 2023-04-11
feat: allowing token dir override

duvanmonsa:
Adding a new flag to override `getDefaultConfigHome()`

How to use it 
```
--config-dir /Users/username/custom-folder
```

Now we are using `os.UserConfigDir()` 

Address request in this issue https://github.com/stateful/runme/issues/123

### DIFF
--- a/cmd/gqltool/main.go
+++ b/cmd/gqltool/main.go
@@ -40,7 +40,6 @@ func main() {
 }
 
 func getDefaultConfigHome() string {
-	// TODO(adamb): switch to os.UserConfigDir()
 	dir, err := os.UserHomeDir()
 	if err != nil {
 		panic(err)
--- a/cmd/gqltool/main.go
+++ b/cmd/gqltool/main.go
@@ -40,9 +40,16 @@ func main() {
 }
 
 func getDefaultConfigHome() string {
-	dir, err := os.UserHomeDir()
+	dir, err := os.UserConfigDir()
 	if err != nil {
 		panic(err)
 	}
-	return filepath.Join(dir, ".config", "stateful")
+	_, fErr := os.Stat(dir)
+	if os.IsNotExist(fErr) {
+		mkdErr := os.MkdirAll(dir, os.ModePerm)
+		if mkdErr != nil {
+			dir = os.TempDir()
+		}
+	}
+	return filepath.Join(dir, "stateful")
 }
--- a/cmd/gqltool/main.go
+++ b/cmd/gqltool/main.go
@@ -6,17 +6,17 @@ import (
 	"fmt"
 	"log" // revive:disable-line
 	"os"
-	"path/filepath"
 
 	"github.com/stateful/runme/internal/auth"
 	"github.com/stateful/runme/internal/client"
 	"github.com/stateful/runme/internal/client/graphql"
+	"github.com/stateful/runme/internal/cmd"
 	"golang.org/x/oauth2"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


cmd/gqltool/main.go
```diff
@@ -40,10 +40,16 @@ func main() {
...
 	}
-	return filepath.Join(dir, ".config", "stateful")
+	_, fErr := os.Stat(dir)
+	if os.IsNotExist(fErr) {
+		mkdErr := os.MkdirAll(dir, os.ModePerm)
+		if mkdErr != nil {
+			dir = os.TempDir()
+		}
+	}
+	return filepath.Join(dir, "stateful")
```
mxsdev, line 54: Do you think we could eliminate the code duplication here by making the `internal` function public and importing it here? 

If we were ever gonna unify the two implementations, now would be a good time to do that for sure.

internal/cmd/api_common.go
```diff
@@ -56,11 +58,13 @@ func getTrace() bool       { return trace || traceAll }
 func getTraceAll() bool    { return traceAll }
 func getEnableChaos() bool { return enableChaos }
 func getAPIToken() string  { return apiToken }
+func getTokenDir() string  { return tokenDir }
 
 func setAPIFlags(flagSet *pflag.FlagSet) {
 	flagSet.StringVar(&authBaseURL, authURLF, defaultAuthURL, "Backend URL to authorize you")
 	flagSet.StringVar(&apiBaseURL, apiURLF, "https://api.stateful.com", "Backend URL with API")
 	flagSet.StringVar(&apiToken, apiTokenF, "", "API token")
+	flagSet.StringVar(&tokenDir, tokenDirF, getDefaultConfigHome(), "Location dir where token will be save")
```
mxsdev, line 67: I think that "config directory" makes a lot more sense than "token directory"

IMO we should change the language around throughout the codebase so it's e.g. `--config-dir` rather than `--token-dir`. Just an opinion though ¯\\\_(ツ)_/¯

mxsdev, line 67:
```suggestion
	flagSet.StringVar(&tokenDir, tokenDirF, getDefaultConfigHome(), "Location where token will be saved")
```

cmd/gqltool/main.go
```diff
@@ -40,10 +40,16 @@ func main() {
...
-	// TODO(adamb): switch to os.UserConfigDir()
-	dir, err := os.UserHomeDir()
+	dir, err := os.UserConfigDir()
 	if err != nil {
 		panic(err)
 	}
-	return filepath.Join(dir, ".config", "stateful")
+	_, fErr := os.Stat(dir)
+	if os.IsNotExist(fErr) {
+		mkdErr := os.MkdirAll(dir, os.ModePerm)
```
mxsdev, line 49:
```suggestion
		mkdErr := os.MkdirAll(dir, 0o700)
```

The permissions you used ("ModePerm") is equivalent to `777` which is **extremely dangerous**. We should use `700` which allows all permissions but only for the user. (See [here](https://mason.gmu.edu/~montecin/UNIXpermiss.htm) for an explanation).

mxsdev (changes requested):
See comments

internal/cmd/api_common.go
duvanmonsa: Yes I like that one better :)

cmd/gqltool/main.go
duvanmonsa: This was implemented

mxsdev (approved):
LGTM 👍

Issue #210, opened 2023-04-11
Pre-release polish for branch subcommand


sourishkrout:
Non-functional stuff:
- [x] Have `login` and `logout` be top-level commands
- [x] Add `token` and require flag `--insecure` to be passed

Changes to CLI:
- [x] Change home directory base to include `runme` vs `stateful` (concrete PR please)
- [x] Once released we want to see `runme branch` in the help section even if `runme suggest branch` is available

Changes to TUI:
- [x] Introduce `t` shortcut to toggle branch prefixes: e.g. `seb/web-extension-props` <-> `web-extension-props` (look for [chai/bubbletea](https://github.com/charmbracelet/bubbletea))
- [x] Introduce `e` to select+edit an entry from the available list: e.g. `seb/web-extension-props` <-> `seb/web-extension-v2`

Before release:
- [x] Login splash page for Runme
- [x] Create distinct auth endpoint e.g. `/auth/cli/runme` for auth token
- [x] Record signup origin in database (e.g. users table)
- [x] Polish OpenAI integration: don't include dependabot (automated) branches

PR #215, opened 2023-04-13
feat: adding new get token cmd plus remove auth as sub cmd

duvanmonsa:
Addressing Non-functional stuff requirements from this issue https://github.com/stateful/runme/issues/210
We remove auth as a valid command to use them direct instead
 Before 
 ```
 runme auth login 
 ```
 Now 
 ```
 runme login
 ```
 
 We also introduced a new `runme token --insecure` command to return currenct runeme token where insecure parameter is required.

### DIFF
--- a/internal/cmd/auth.go
+++ b/internal/cmd/auth.go
@@ -3,7 +3,6 @@ package cmd
 import (
 	"context"
 	"fmt"
-	"os"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -35,8 +34,6 @@ func logoutCmd() *cobra.Command {
 }
 
 func tokenCmd() *cobra.Command {
-	var insecureF bool
-
 	cmd := &cobra.Command{
 		Use:    "token",
 		Hidden: true,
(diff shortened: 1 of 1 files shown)


internal/cmd/auth.go
```diff
@@ -2,44 +2,72 @@ package cmd
...
+
+					token, err = auth.GetToken(cmd.Context())
+					if err != nil {
+						return err
+					}
+				}
+				_, err = fmt.Fprint(os.Stdout, token, "\n")
+				return err
+			}
+			_, err = fmt.Fprint(os.Stdout, "To use this command, please add the --insecure parameter")
```
mxsdev, line 64:
```suggestion
			return errors.New("To use this command, please add the --insecure parameter")
```

As mentioned in the other comment, you generally want to avoid using `os.Stdout` directly and you also generally want to discard errors from stdout.

In this case, we should definitely return an error with the message directly.

internal/cmd/auth.go
```diff
@@ -2,44 +2,72 @@ package cmd
...
+			return err
 		},
 	}
+
+	cmd.Flags().BoolVar(&insecureF, "insecure", false, "Using insecure helper to get the runme token.")
```
mxsdev, line 69:
```suggestion
```

There is no need to have this line, the `insecure` flag is global. Please use `fInsecure` instead

mxsdev (changes requested):
Functionality is good, see comments for changes.

mxsdev (approved):
LGTM, thank you!

PR #216, opened 2023-04-17
Feat/adding new cmd helpers

duvanmonsa:
New shortcut for suggest branch list 

- t  toggle the suffix if applies
- Before toggling the suffix 
image
After toggling the suffix 
image
- e edit the selected suggest
image
You can save it with crtl+s or enter it will take you to the previews screen. 
Handle invalid string
Before safe 
image
After save
image

### DIFF
--- a/internal/cmd/common.go
+++ b/internal/cmd/common.go
@@ -195,7 +195,7 @@ func GetDefaultConfigHome() string {
 			dir = os.TempDir()
 		}
 	}
-	return filepath.Join(dir, "stateful")
+	return filepath.Join(dir, "runme")
 }
 
 func setRunnerFlags(cmd *cobra.Command, serverAddr *string) func() ([]client.RunnerOption, error) {
--- a/internal/tui/model_list.go
+++ b/internal/tui/model_list.go
@@ -69,7 +69,7 @@ type ListModel struct {
 var listKeys = []key.Binding{
 	key.NewBinding(
 		key.WithKeys("t"),
-		key.WithHelp("t", "toggle suffix"),
+		key.WithHelp("t", "toggle prefix"),
 	),
 	key.NewBinding(
 		key.WithKeys("e"),
@@ -383,7 +383,7 @@ func newGetSuggestedBranch(description string, userBranches []project.Branch, re
 }
 
 func removeSpecialChars(unsanitized string) string {
-	pattern := regexp.MustCompile(`[^A-Za-z\-_\/]+`)
+	pattern := regexp.MustCompile(`[^A-Za-z0-9\-_\/]+`)
 	sanitized := pattern.ReplaceAllString(unsanitized, "")
 	return sanitized
 }
@@ -394,7 +394,7 @@ func sanitizeBranchName(unsanitized []string) []string {
 	// then we omit the suggestion.
 
 	// Edge case is of course that we are left with an empty list.
-	isInWhitelist := regexp.MustCompile(`^[A-Za-z-_\/]+$`).MatchString
+	isInWhitelist := regexp.MustCompile(`^[A-Za-z0-9-_\/]+$`).MatchString
 
 	sanitized := make([]string, 0)
 	for _, s := range unsanitized {
--- a/internal/tui/model_list.go
+++ b/internal/tui/model_list.go
@@ -226,59 +226,58 @@ func (m ListModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 
 			m.input, cmd = m.input.Update(msg)
 			return m, cmd
-		} else {
-			switch msg.String() {
+		}
+		switch msg.String() {
 
-			case "up":
-				m.list.CursorUp()
+		case "up":
+			m.list.CursorUp()
 
-				return m, nil
-			case "down":
-				m.list.CursorDown()
+			return m, nil
+		case "down":
+			m.list.CursorDown()
 
-				return m, nil
-			case "enter":
+			return m, nil
+		case "enter":
 
-				if m.loading {
-					return m, nil
-				}
-
-				selected, ok := m.list.SelectedItem().(item)
-
-				return m, func() tea.Msg {
-					if !ok {
-						return errorMsg{errors.New("unable to select an item")}
-					}
-
-					m.selected = selected
-
-					cwd, err := os.Getwd()
-					if err != nil {
-						return errorMsg{err}
-					}
-
-					cmdSlice := []string{"git", "checkout", "-b", m.selected.title}
-					fmt.Printf("Output: %s", cmdSlice)
-
-					cmd := exec.Command(cmdSlice[0], cmdSlice[1:]...)
-					cmd.Dir = cwd
-
-					out, err := cmd.CombinedOutput()
-					if err != nil {
-						return errorMsg{err}
-					}
-
-					return confirmMsg{string(out)}
-				}
-			case "t":
-				m := m.TogglePrefixes()
-				return m, nil
-			case "e":
-				selected := m.list.SelectedItem().(item)
-				m.input.SetValue(selected.title)
-				m.input.Focus()
+			if m.loading {
 				return m, nil
 			}
+
+			selected, ok := m.list.SelectedItem().(item)
+
+			return m, func() tea.Msg {
+				if !ok {
+					return errorMsg{errors.New("unable to select an item")}
+				}
+
+				m.selected = selected
+
+				cwd, err := os.Getwd()
+				if err != nil {
+					return errorMsg{err}
+				}
+
+				cmdSlice := []string{"git", "checkout", "-b", m.selected.title}
+				fmt.Printf("Output: %s", cmdSlice)
+
+				cmd := exec.Command(cmdSlice[0], cmdSlice[1:]...)
+				cmd.Dir = cwd
+
+				out, err := cmd.CombinedOutput()
+				if err != nil {
+					return errorMsg{err}
+				}
+
+				return confirmMsg{string(out)}
+			}
+		case "t":
+			m := m.TogglePrefixes()
+			return m, nil
+		case "e":
+			selected := m.list.SelectedItem().(item)
+			m.input.SetValue(selected.title)
+			m.input.Focus()
+			return m, nil
 		}
 
 		// Cool, what was the actual key pressed?
@@ -330,9 +329,8 @@ func (m ListModel) View() string {
 		s += inputStyle.Render(m.input.View()) + "\n"
 		s += helpStyle.Render(m.help.View(m.keys))
 		return appStyle.Render(s)
-	} else {
-		s += appStyle.Render(m.list.View())
 	}
+	s += appStyle.Render(m.list.View())
 
 	return s
 }
--- a/internal/tui/model_list.go
+++ b/internal/tui/model_list.go
@@ -55,7 +55,6 @@ type ListModel struct {
 	list        list.Model
 
 	editInput textinput.Model
-	editKeys  KeyMap
 	editHelp  help.Model
 
 	confirmed bool
@@ -68,17 +67,6 @@ type ListModel struct {
 	err error
 }
 
-var listKeys = []key.Binding{
-	key.NewBinding(
-		key.WithKeys("t"),
-		key.WithHelp("t", "toggle prefix"),
-	),
-	key.NewBinding(
-		key.WithKeys("e"),
-		key.WithHelp("e", "edit branch name"),
-	),
-}
-
 func NewListModel(ctx context.Context, description string, repoUser string, client *graphql.Client) ListModel {
 	s := spinner.New()
 	s.Spinner = spinner.Line
@@ -89,19 +77,13 @@ func NewListModel(ctx context.Context, description string, repoUser string, clie
 	l.SetFilteringEnabled(false)
 	l.SetShowStatusBar(false)
 	l.SetShowPagination(false)
-	l.AdditionalShortHelpKeys = func() []key.Binding {
-		return listKeys
-	}
-	l.AdditionalFullHelpKeys = func() []key.Binding {
-		return listKeys
-	}
 
 	ti := textinput.New()
 	ti.Prompt = ""
 	ti.CharLimit = 250
 	ti.Width = 50
 
-	return ListModel{
+	m := ListModel{
 		ctx:         ctx,
 		client:      client,
 		log:         log.Get().Named("SuggestionsListModel"),
@@ -109,12 +91,20 @@ func NewListModel(ctx context.Context, description string, repoUser string, clie
 		description: description,
 		list:        l,
 		editInput:   ti,
-		editKeys:    *editInputKeys(),
 		editHelp:    help.New(),
 		spinner:     s,
 		loading:     true,
 		usePrefix:   true,
 	}
+
+	m.list.AdditionalShortHelpKeys = func() []key.Binding {
+		return m.KeyMap().ShortHelp()
+	}
+	m.list.AdditionalFullHelpKeys = func() []key.Binding {
+		return m.KeyMap().ShortHelp()
+	}
+
+	return m
 }
 
 func (m ListModel) startSearch() tea.Msg {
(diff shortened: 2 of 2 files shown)


internal/tui/model_list.go
```diff
@@ -36,16 +52,31 @@ type ListModel struct {
...
+	selected  item
+	usePrefix bool
 
 	err error
 }
 
+var listKeys = []key.Binding{
+	key.NewBinding(
+		key.WithKeys("t"),
+		key.WithHelp("t", "toggle suffix"),
```
sourishkrout, line 72: this should be `toggle prefix` right? suffix is ending, prefix is preceding?

internal/tui/model_list.go
```diff
@@ -261,6 +382,12 @@ func newGetSuggestedBranch(description string, userBranches []project.Branch, re
 	}, nil
 }
 
+func removeSpecialChars(unsanitized string) string {
+	pattern := regexp.MustCompile(`[^A-Za-z\-_\/]+`)
```
sourishkrout, line 386: any particular reason we don't allow numbers? E.g. `seb/ticket-123` is a valid branch name, no?

sourishkrout (changes requested):
Functionality looks 👍 . Don't think we need to worry about edge-cases such as "adding a prefix" when prefix is toggled off yet.

Flagged some minor issues. I'll let Max take a look at code/structure etc.

internal/tui/model_list.go
duvanmonsa: Yes that's right my bad :/

internal/tui/model_list.go
duvanmonsa: I was following the current logic 
image
And we don't support numbers currently so I guest that was a mistake from the beginning I will fix it on both sides.

internal/tui/model_list.go
```diff
@@ -36,16 +52,31 @@ type ListModel struct {
...
+var listKeys = []key.Binding{
+	key.NewBinding(
+		key.WithKeys("t"),
+		key.WithHelp("t", "toggle prefix"),
+	),
+	key.NewBinding(
+		key.WithKeys("e"),
+		key.WithHelp("e", "edit suggestion"),
+	),
+}
```
mxsdev, line 78: There is already a system for custom key bindings given by the `KeyMap()` function. I personally would prefer to add these key binds to that system.

Then you can use the keymap's `ShortHelp` function to provide help to the `AdditionalShortHelpKeys` function. And you can go back to using `kmap.Matches` for the update function.

DM me on slack for more info on how to filter specific keys out further, or if this is otherwise confusing.

internal/tui/model_list.go
```diff
@@ -138,20 +171,73 @@ type errorMsg struct {
...
+		case "up":
+			m.list.CursorUp()
+
+			return m, nil
+		case "down":
+			m.list.CursorDown()
+
+			return m, nil
+		case "enter":
+
```
mxsdev, line 240:
```suggestion
		kmap := m.KeyMap()

		if kmap.Matches(msg, "t") {
			m := m.TogglePrefixes()
			return m, nil
		}

		if kmap.Matches(msg, "e") {
			selected := m.list.SelectedItem().(item)
			m.input.SetValue(selected.title)
			m.input.Focus()
			return m, nil
		}

		if kmap.Matches(msg, "enter") {
```

If you migrate to the `KeyMap` system, you'll do this instead

internal/tui/model_list.go
```diff
@@ -138,20 +171,73 @@ type errorMsg struct {
...
+var inputKeys = keyMap{
+	Save: key.NewBinding(
+		key.WithKeys("s", "enter"),
+		key.WithHelp("enter/ctrl+s", "save suggest"),
+	),
+	Quit: key.NewBinding(
+		key.WithKeys("esc", "esc", "ctrl+c"),
+		key.WithHelp("esc/ctrl+c", "quit"),
+	),
+}
```
mxsdev, line 199: I think for consistency it is best to use the same keymap system throughout. So then, how would you feel about refactoring out of this `keyMap` struct to be of type `KeyMap`?

This way, there would be an `EditKeyMap()` function on the model, analogous to the current `KeyMap()` function. 

One benefit, for example, you wouldn't have to define your own `ShortHelp` and `FullHelp` functions. There is also no more duplication of keyboard shortcuts.

internal/tui/model_list.go
```diff
@@ -138,20 +171,73 @@ type errorMsg struct {
...
+				m.list.SetItem(m.list.Cursor(), createListItem(removeSpecialChars(m.input.Value()), m.usePrefix))
+				m.input.Blur()
+				return m, nil
+			case tea.KeyCtrlC, tea.KeyEsc:
+				m.input.Blur()
+				return m, nil
+			}
+
+			m.input, cmd = m.input.Update(msg)
+			return m, cmd
```
mxsdev, line 227: If you refactor `inputKeys`, then you'll want to use `kmap.Matches` here instead. 

Feel free to DM on slack for clarification about this if needed 👍

internal/tui/model_list.go
```diff
@@ -226,21 +320,48 @@ func (m ListModel) View() string {
...
-	return appStyle.Render(m.list.View())
+
+	if m.input.Focused() {
+		s += titleStyle.Render("Edit suggestion:") + "\n"
+		s += inputStyle.Render(m.input.View()) + "\n"
+		s += helpStyle.Render(m.help.View(m.keys))
+		return appStyle.Render(s)
+	}
+	s += appStyle.Render(m.list.View())
+
```
mxsdev, line 331:
```suggestion

	if m.input.Focused() {
		s += titleStyle.Render("Edit suggestion:") + "\n"
		s += inputStyle.Render(m.input.View()) + "\n"
		s += helpStyle.Render(m.help.View(m.keys))
		return appStyle.Render(s)
	}
	
	m.list.Title = "How about one of these:"
	s += appStyle.Render(m.list.View())
```

Nit: Bringing the title change down after the input render section I think is cleaner and clearer code.

internal/tui/model_list.go
```diff
@@ -21,10 +24,23 @@ import (
...
+			Foreground(lipgloss.Color("241"))
+	inputStyle = lipgloss.NewStyle().
+			MarginLeft(1).
+			MarginBottom(2).
+			BorderStyle(lipgloss.NormalBorder()).
+			BorderForeground(lipgloss.Color("63")).
+			Foreground(lipgloss.Color("#87448b"))
+	titleStyle = lipgloss.NewStyle().
+			Foreground(lipgloss.Color("#FAFAFA")).
+			Background(lipgloss.Color("#5d5dd2")).
```
mxsdev, line 39: Long-term we might want to think about standardizing these styles in some way. Something to think about if we continue to develop around the TUI.

mxsdev (changes requested):
Functionality is great! Implementation-wise, there are a few things - see comments.

Main thing here is to use the internal `KeyMap` struct, which Babik made (I think) for the purpose of removing redundancy in keymap definitions. We're already using this around the code-base, so I figure why not keep consistency there.

internal/cmd/common.go
```diff
@@ -195,7 +195,7 @@ func GetDefaultConfigHome() string {
 			dir = os.TempDir()
 		}
 	}
-	return filepath.Join(dir, "stateful")
+	return filepath.Join(dir, "runme")
```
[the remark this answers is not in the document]
duvanmonsa: please let me know and I will move it to its own PR :)

internal/tui/model_list.go
```diff
@@ -226,21 +320,48 @@ func (m ListModel) View() string {
...
+	title      string
 }
 
-func (m item) FilterValue() string { return m.suggestion }
-func (m item) Title() string       { return listItemStyle.Render(m.suggestion) }
-func (m item) Description() string {
-	return listItemHelpStyle.Render("git checkout -b '" + m.suggestion + "'")
+func (i item) Title() string { return i.title }
+func (i item) Description() string {
+	return "git checkout -b '" + i.title
```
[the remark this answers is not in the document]
duvanmonsa: I am not sure this makes sense, here we need to return the string that will be printed by the list model as a subtitle. but the list model is the one in charge of printing.  does it make sense?

internal/tui/model_list.go
```diff
@@ -226,21 +320,48 @@ func (m ListModel) View() string {
...
+}
+
+func createListItem(suggestion string, showPrefix bool) item {
+	item := item{suggestion: suggestion, title: suggestion}
+	if !showPrefix {
+		parts := strings.Split(suggestion, "/")
+		if len(parts) > 0 {
+			item.title = strings.Replace(suggestion, parts[0]+"/", "", 1)
+		}
+		item.title = strings.Replace(suggestion, parts[0]+"/", "", 1)
```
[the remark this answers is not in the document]
duvanmonsa: I actually saw during the lint process this change suggestion but didn't want to accept it jeje I will do it now.

internal/tui/model_list.go
```diff
@@ -226,21 +320,48 @@ func (m ListModel) View() string {
...
 type item struct {
 	suggestion string
+	title      string
 }
 
-func (m item) FilterValue() string { return m.suggestion }
-func (m item) Title() string       { return listItemStyle.Render(m.suggestion) }
-func (m item) Description() string {
-	return listItemHelpStyle.Render("git checkout -b '" + m.suggestion + "'")
+func (i item) Title() string { return i.title }
```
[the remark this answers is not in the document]
duvanmonsa: I didn't see any difference when removing it that's why I removed it.

internal/tui/model_list.go
```diff
@@ -183,12 +269,20 @@ func (m ListModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+			m.input.SetValue(selected.title)
+			m.input.Focus()
+			return m, nil
 		}
 
 	case suggestionsMsg:
 		m.suggestions = msg.suggestions
 		for i, s := range m.suggestions {
-			m.list.InsertItem(0, createListItem(s, i))
+			m.list.InsertItem(i, createListItem(s, true))
```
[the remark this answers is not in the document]
duvanmonsa: I am putting back the `m.list.InsertItem(0, createListItem(s, true))` but removing the index param was doing nothing. Also, My feeling is that passing 0 is more confusing than actually passing the index position.

internal/tui/model_list.go
```diff
@@ -226,21 +320,48 @@ func (m ListModel) View() string {
...
+func (i item) FilterValue() string { return i.suggestion }
+
+func (m ListModel) TogglePrefixes() ListModel {
+	m.usePrefix = !m.usePrefix
+	for i, item := range m.list.Items() {
+		m.list.SetItem(i, createListItem(item.FilterValue(), m.usePrefix))
+	}
+	return m
+}
+
```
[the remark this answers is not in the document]
duvanmonsa: I can't do that because I need to be able to go back to the original value.

internal/tui/model_list.go
duvanmonsa: I agreed on this.

internal/tui/model_list.go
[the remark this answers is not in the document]
mxsdev: Ah yeah, I see what you're saying. Fell free to revert to what you did, and ignore me!

internal/tui/model_list.go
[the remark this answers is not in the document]
mxsdev: > I can't override the original suggest because I need to be able to go back to the original value if you press T again

Right, that's why you wouldn't change the internal `item.suggestion` value, you would just change the behavior of `item.Title()`, depending on whether or not prefixes are enabled.

internal/cmd/common.go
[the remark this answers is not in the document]
sourishkrout: 👍. Not necessary. Thank you.

internal/tui/model_list.go
```diff
@@ -56,6 +87,17 @@ func NewListModel(ctx context.Context, description string, repoUser string, clie
 	l.SetFilteringEnabled(false)
 	l.SetShowStatusBar(false)
 	l.SetShowPagination(false)
+	l.AdditionalShortHelpKeys = func() []key.Binding {
```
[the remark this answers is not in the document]
sourishkrout: Let's go ahead and leave it there for now. Once we've released the feature, we can move it off the short help.

internal/tui/model_list.go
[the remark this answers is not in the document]
duvanmonsa: Ok I see more complications with that approach than benefits, the way you propose we need to create a separate function because we need to do this for Title() and Description() plus the flag is in the model ` m.usePrefix` and Title() receiver only has the item. we would need to make this a global var. What are you thought @mxsdev based on this information?

duvanmonsa:
https://github.com/stateful/runme/issues/210

internal/tui/model_list.go
```diff
@@ -56,6 +89,17 @@ func NewListModel(ctx context.Context, description string, repoUser string, clie
...
+		return listKeys
+	}
+	l.AdditionalFullHelpKeys = func() []key.Binding {
+		return listKeys
+	}
```
mxsdev, line 97:
```suggestion
	l.AdditionalShortHelpKeys = func() []key.Binding {
		return m.KeyMap().ShortHelp()
	}
	l.AdditionalFullHelpKeys = func() []key.Binding {
		return m.KeyMap().ShortHelp()
	}

	m.list = l
```

This removes the need for `listKeys`

Note that you'll need to bring this logic to after the model is initialized.

internal/tui/model_list.go
```diff
@@ -118,31 +182,38 @@ func (m ListModel) KeyMap() *KeyMap {
...
 }
 
-func (m ListModel) Init() tea.Cmd {
-	return tea.Batch(m.startSearch, m.spinner.Tick)
-}
-
-type suggestionsMsg struct {
-	suggestions []string
-}
+func editInputKeys() *KeyMap {
```
mxsdev, line 195:
```suggestion
func (m ListModel) EditKeyMap() *KeyMap {
```

To be consistent with the other key map implementation

This removes the need for the `editKeys` field

internal/tui/model_list.go
```diff
@@ -118,31 +182,38 @@ func (m ListModel) KeyMap() *KeyMap {
 	))
 	kmap.Set("up", m.list.KeyMap.CursorUp)
 	kmap.Set("down", m.list.KeyMap.CursorDown)
+	kmap.Set("t", key.NewBinding(
+		key.WithKeys("t"),
+	))
+	kmap.Set("e", key.NewBinding(
+		key.WithKeys("e"),
+	))
```
mxsdev, line 190:
```suggestion
	kmap.Set("t", key.NewBinding(
		key.WithKeys("t"),
		key.WithHelp("t", "toggle prefix"),
	))
	kmap.Set("e", key.NewBinding(
		key.WithKeys("e"),
		key.WithHelp("e", "edit branch name"),
	))
```

internal/tui/model_list.go
```diff
@@ -151,7 +222,36 @@ func (m ListModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 
 	case tea.KeyMsg:
 		kmap := m.KeyMap()
```
mxsdev, line 224: You can bring this line down below the `if m.editInput.Focused()` block, since it is not used until after that.

internal/tui/model_list.go
```diff
@@ -163,14 +263,14 @@ func (m ListModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
-				m.selected = string(selected.suggestion)
+				m.selected = selected
 
 				cwd, err := os.Getwd()
 				if err != nil {
 					return errorMsg{err}
 				}
 
-				cmdSlice := []string{"git", "checkout", "-b", m.selected}
+				cmdSlice := []string{"git", "checkout", "-b", m.selected.branchName}
```
mxsdev, line 273:
```suggestion
				cmdSlice := []string{"git", "checkout", "-b", m.selected.Suggestion()}
```

Careful here, we don't want to use the branch name, we want to use the suggested branch name (e.g. without the prefix)

mxsdev (changes requested):
Few more changes, other than that it's great!

internal/tui/model_list.go
duvanmonsa: good catch thank you.

mxsdev (approved):
LGTM 👍

sourishkrout (approved):
👍 LGTM

PR #218, opened 2023-04-18
parser: add text range to code blocks

mxsdev:
Includes a "text range" for code blocks, which includes the start and end text offsets for the fenced code block.

This PR is needed for CodeLens support in `vscode-runme`.


sourishkrout (approved):
Looks straightforward 👍.

PR #227, opened 2023-04-25
internal/runner: remove `DisableEcho`

mxsdev:
Seems we manually disable tty echo by default in the runner.

This is causing issues for commands like `ssh git@gitlab.com`, which normally echo the output back, but now will not (since the default behavior is to echo, but we have overridden that).

@adambabik Could really use your insight here - why did you do this? I'm not able to replicate the "double input" issue in any of my tests - do you remember what kinds of scripts were leading to this?

### DIFF
--- a/internal/runner/service.go
+++ b/internal/runner/service.go
@@ -201,7 +201,7 @@ func (r *runnerService) Execute(srv runnerv1.RunnerService_ExecuteServer) error 
 		cmdCtx = context.Background()
 	}
 
-	if err := cmd.StartWithOpts(cmdCtx, &startOpts{DisableEcho: req.Tty}); err != nil {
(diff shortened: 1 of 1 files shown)


adambabik:
I think the issue exists if you have the same controlling terminal for the input and runme process. I backtracked it to https://github.com/stateful/runme/pull/126 and https://github.com/stateful/runme/pull/139. In the first one, you will find a script that you can test it with.

However, I think it could be a mistake applying it to the service which does not share the controlling terminal.

mxsdev:
Interesting, I am unable to reproduce with that script in my tests. 

I actually think the reason this was happening was because the terminal used to not be set to raw mode when running `runme run`. So then the input would be echoed twice: once for the internal terminal, and once for the external terminal.

Issue #229, opened 2023-04-27
Multiple tasks and parallelism in `runme run`


sourishkrout:
In a discreet PR let's make it possible to run multiple commands optionally sequential and parallel available.

The rough idea is `runme run one two three -p four five six -s seven eight nine` (please note the presence of `runme run` as a subcommand for now).

sourishkrout:
cc @mxsdev. In the meantime, I will review the other suggested concepts, including relaxing the file-centricity @christian-bromann.

mxsdev:
@christian-bromann 

For implementation I think we should introduce two new functions which take the `Runner` interface  (`internal/runner/client/client.go`). Probably the most idiomatic way to do this is to introduce a struct, let's call it `MultiRunner`, which encapsulates some internal runner(s), and then to define methods on that. You create a multirunner by passing a `[]RunnerOption` list, and each command run gets its own `Runner` instance.

The first function will be `RunBlocksSerial`. The implementation here is straight-forward: run each command one after the other, and exit if one of these commands fails.

The second function will be `RunBlocksParallel`. This implementation is less straight forward. Each call to `RunBlock` should have its own context and run in its own goroutine. We can use a `WaitGroup` here to wait for all goroutines to finish, and you'll also need to introduce a channel to pipe command results if you want to summarize them at the end. 

It will also be absolutely critical that each execution has it's own stdin and stdout objects. You can accomplish this by connecting an `io.Pipe()` to the assigned stdout, stderr, and stdin. (This also gives us the opportunity to prefix each stdout line with the associated command's name, if we so wish) 

You will also have to modify the `Runner` interface to allow "transformation" of the stdin/out objects - some method like `WithTransformStdout` which passes the current stdout and assigns the result would be a nice functional way of doing this in a way compatible with the `RunnerOption` paradigm. Then for each runner, you can create a new pipe and push/pull the necessary changes through that pipe, and all that logic could happen in the transform callback.

As for the command parsing, it would be very nice to keep manual parsing to an absolute minimum, ideally with everything being handled by [cobra](https://pkg.go.dev/github.com/spf13/cobra). I don't personally know all the limits of that library, but I implore you to read as much of the docs as possible before resorting to manual parsing.

I also anticipate some very complicated issues regarding stdin. You'll need to modify the current behavior of the runner clients in such a way that the multirunner cancels the stdin not once a single command finishes, but after they all finish.

PR #230, opened 2023-04-30
Improvements to `run` and `list` commands

christian-bromann:
According to our internal proposal this patch makes some changes to the `run` and `list` commands. These features are behind an environment flag `RUNME_EXPERIMENTAL_CLI`. So the current behavior doesn't change. The new (experimental) CLI behavior works as following:

- the `list` command now lists all commands of all markdown files within a project
  - it only shows the command id and the description, everything else seems not relevant
- the `run` command
  - now only takes the block name to execute it
  - can receive more block names which will be run sequentially
  - now has a `-p` or `--parallel` flag to run a set of commands in parallel
  - allows to execute sequential commands and parallel ones through one CLI call, e.g. `runme run foo bar loo -p faz baz -s loo bar` will run: `foo`, `bar` and `loo` sequentially, then will run `faz` and `baz` in parallel and lastly `loo` and `bar` sequential again

__List Command Demo:__
Screenshot 2023-05-01 at 22 59 55

Screenshot of the WebdriverIO project based on [`webdriverio/webdriverio#10237`](https://github.com/webdriverio/webdriverio/pull/10237)

__Run Command Demo:__
run

__Reviewer Notes:__
This patch basically does the following:
- moves flat functions like `getCodeBlocks` or `readMarkdownFile` into the project module
- implements a `getAllMarkdownFiles` in the project module to find all markdown files
- implement `LookUpCodeBlockById` to find a block just by its name
- __unrelated:__ removed `-s` for server address as it seems to be better suited (and more used) for providing the sequential flag

__Outstanding:__
- [x] read from `.gitignore` to not include markdown files matching these patterns (otherwise we would e.g. read from a bunch of `node_modules` which would be quite overwhelming
- [ ] write unit tests
- [ ] more thorough testing of new functionality
- [ ] detect duplicate commands and throw errors
- [x] improve error handling

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -11,10 +11,11 @@ import (
 )
 
 var (
-	fAllowUnknown bool
-	fChdir        string
-	fFileName     string
-	fInsecure     bool
+	fAllowUnknown   bool
+	fIgnoreNameless bool
+	fChdir          string
+	fFileName       string
+	fInsecure       bool
 )
 
 func Root() *cobra.Command {
@@ -47,6 +48,7 @@ func Root() *cobra.Command {
 	pflags := cmd.PersistentFlags()
 
 	pflags.BoolVar(&fAllowUnknown, "allow-unknown", true, "Display snippets without known executor")
+	pflags.BoolVar(&fIgnoreNameless, "ignore-nameless", true, "Ignore code blocks without name defined")
 	pflags.StringVar(&fChdir, "chdir", getCwd(), "Switch to a different working directory before executing the command")
 	pflags.StringVar(&fFileName, "filename", "README.md", "Name of the README file")
 	pflags.BoolVar(&fInsecure, "insecure", false, "Run command in insecure-mode")
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -16,6 +16,7 @@ var (
 	fChdir          string
 	fFileName       string
 	fInsecure       bool
+	fExperimental   bool
 )
 
 func Root() *cobra.Command {
@@ -52,6 +53,7 @@ func Root() *cobra.Command {
 	pflags.StringVar(&fChdir, "chdir", getCwd(), "Switch to a different working directory before executing the command")
 	pflags.StringVar(&fFileName, "filename", "README.md", "Name of the README file")
 	pflags.BoolVar(&fInsecure, "insecure", false, "Run command in insecure-mode")
+	pflags.BoolVar(&fExperimental, "experimental", false, "Enable experimental features")
 
 	setAPIFlags(pflags)
 
--- a/internal/cmd/run.go
+++ b/internal/cmd/run.go
@@ -41,11 +41,7 @@ func runCmd() *cobra.Command {
 		Args:              cobra.ArbitraryArgs,
 		ValidArgsFunction: validCmdNames,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			p, err := project.New(fChdir)
-			if err != nil {
-				return err
-			}
-
+			p := project.New(fChdir)
 			if !isInExperimentalMode() {
 				filePath := path.Join(fChdir, fFileName)
 				blocks, err := p.GetCodeBlocks(filePath[len(p.RootDir):], fAllowUnknown)
--- a/internal/cmd/run.go
+++ b/internal/cmd/run.go
@@ -44,7 +44,7 @@ func runCmd() *cobra.Command {
 			p := project.New(fChdir)
 			if !isInExperimentalMode() {
 				filePath := path.Join(fChdir, fFileName)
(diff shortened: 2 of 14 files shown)


sourishkrout:
The scope of the original ticket https://github.com/stateful/runme/issues/229 is about changes, there wasn't green light for any other changes yet @christian-bromann.

internal/cmd/root.go
```diff
@@ -47,9 +49,11 @@ func Root() *cobra.Command {
 	pflags := cmd.PersistentFlags()
 
 	pflags.BoolVar(&fAllowUnknown, "allow-unknown", true, "Display snippets without known executor")
+	pflags.BoolVar(&fIgnoreNameless, "ignore-nameless", true, "Ignore code blocks without name defined")
```
mxsdev, line 52:
```suggestion
	pflags.BoolVar(&fIgnoreNameless, "ignore-nameless", false, "Ignore code blocks without name defined")
```

Should probably be false by default.

internal/project/project.go
```diff
@@ -1,241 +1,232 @@
...
-func (r *repo) Err() error { return r.err }
-
-func (r *repo) BranchName() string {
-	if r.Err() != nil {
-		return ""
-	}
-
-	ref, err := r.Head()
+func (p *Project) getAllMarkdownFiles() ([]string, error) {
+	root, err := os.Getwd()
```
mxsdev, line 60:
```suggestion
	root := p.RootDir
```

internal/project/project.go
```diff
@@ -1,241 +1,232 @@
...
 	if err != nil {
-		r.err = err
-		return ""
+		return nil, err
 	}
 
-	if ref.Name().IsBranch() {
-		return ref.Name().Short()
+	fsys := os.DirFS(root)
+	matches, err := fs.Glob(fsys, "**/*.md")
```
mxsdev, line 66: This [does not work](https://github.com/golang/go/issues/11862) as expected, and in particular will ignore files in the root directory.

You should probably use a `WalkDir`-based solution instead.

internal/project/project.go
```diff
@@ -1,241 +1,232 @@
...
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
-	branches, err := GetBranchNames(cwd)
-	if err != nil {
-		return nil, errors.New("error while querying repository branches")
+	for _, file := range files {
+		codeBlock, err := p.GetCodeBlocks(file[len(p.RootDir):], false, true)
```
mxsdev, line 219:
```suggestion
		codeBlock, err := p.GetCodeBlocks(file[len(p.RootDir):], allowUnknown, ignoreNameless)
```

Should pass these down

internal/cmd/run.go
```diff
@@ -3,126 +3,344 @@ package cmd
...
+
+	var stdin io.Reader
+
+	if isInExperimentalMode() && block.Interactive() {
+		warn := color.New(color.Bold, color.FgYellow)
+		_, err := warn.Print("Warning: ")
+		if err != nil {
+			return err
+		}
+		fmt.Println("Interactive mode was disabled as it is not supported with the experimental CLI")
```
mxsdev, line 245:
```suggestion
		fmt.Println("Interactivity was disabled as it is not supported with the experimental CLI")
```

internal/cmd/run.go
```diff
@@ -3,126 +3,344 @@ package cmd
...
+
+	go func() {
+		wg.Wait()
+		close(wgDone)
+	}()
+
+	select {
+	case <-wgDone:
+		// carry on
+		break
```
mxsdev, line 337: This break does nothing since there is no surrounding loop

mxsdev (changes requested):
Really loving the UX changes, would love to see this implemented asap!

The runner-based implementation is also reasonable, though I do have some changes in mind. But the broad-strokes idea is good!

internal/cmd/run.go
```diff
@@ -3,126 +3,344 @@ package cmd
...
-			}
+	setDefaultFlags(&cmd)
+
+	cmd.Flags().BoolVar(&dryRun, "dry-run", false, "Print the final command without executing.")
+	cmd.Flags().StringArrayVarP(&replaceScripts, "replace", "r", nil, "Replace instructions using sed.")
+
+	if isInExperimentalMode() {
+		cmd.Flags().BoolVarP(&runInParallel, "parallel", "p", false, "Run commands in parallel. (experimental)")
+		cmd.Flags().BoolVarP(&runSequential, "sequential", "s", true, "Run commands sequentially. (experimental)")
+		cmd.Flags().BoolVarP(&onlyCommandIO, "onlyCommandOutput", "", false, "If set, Runme will only output command output. (experimental)")
```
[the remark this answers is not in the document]
christian-bromann: Oh yeah totally! Nice

christian-bromann:
Closing in favor of #236

PR #235, opened 2023-05-05
feat: adding extension cmd

duvanmonsa:
Porting extension cmd to runme


mxsdev:
@duvanmonsa 

Would you be willing to generate the extension mock files (`make generate`) and also bring in the `internal/cmd/extension_test.go` file? Let me know if it doesn't work

PR #236, opened 2023-05-06
Support multi-runner in `run` command

mxsdev:
Implements #229 

image

Implementation is slightly different than the PR description. It runs the provided arguments in the order given, unless `-p` is provided, in which case it runs them in parallel.

@christian-bromann If you could try this out and let me know what issues you run into, that would be great!

### DIFF
--- a/internal/cmd/run.go
+++ b/internal/cmd/run.go
@@ -185,7 +185,7 @@ func runCmd() *cobra.Command {
 
 	cmd.Flags().BoolVar(&dryRun, "dry-run", false, "Print the final command without executing.")
 	cmd.Flags().StringArrayVarP(&replaceScripts, "replace", "r", nil, "Replace instructions using sed.")
-	cmd.Flags().BoolVarP(&parallel, "parallel", "p", false, "Run scripts in parallel.")
+	cmd.Flags().BoolVarP(&parallel, "parallel", "p", false, "Run tasks in parallel.")
 
 	getRunnerOpts = setRunnerFlags(&cmd, &serverAddr)
 
--- a/internal/cmd/run.go
+++ b/internal/cmd/run.go
@@ -35,6 +35,10 @@ func runCmd() *cobra.Command {
 		Args:              cobra.ArbitraryArgs,
 		ValidArgsFunction: validCmdNames,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) == 0 {
+				return errors.New("must provide at least one command to run")
+			}
+
 			runBlocks := make([]*document.CodeBlock, 0)
 
 			{
--- a/internal/cmd/run.go
+++ b/internal/cmd/run.go
@@ -144,7 +144,7 @@ func runCmd() *cobra.Command {
 			}
 
 			if parallel {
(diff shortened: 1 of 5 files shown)


christian-bromann:
Some feedback:

- running `runme run --help` has:
   ```
   Usage:
      runme run [flags]
   ```
   but should indicate that you can pass in command ids and not only flags
- we use the same icon for executing and printing the result:
   ```
    ►  Running script echo...
    Hello, runme!
    ►  Script echo exited with code 0
    ```
    I wonder it makes sense to have a different icon like ✅ for exit code 0 or 🔴 for exit code != 0
- it doesn't seem like switching between parallel and sequential execution is not supported
   loop
   I am not sure if this is executing the commands in parallel or not, all commands should finish at the same time since I call the same workflows. I am assuming it is out of scope in this PR? Do we still consider to add it in a later PR?
- the following parameter:
   ```
   -p, --parallel              Run scripts in parallel.
   ```
   should we word it as _workflows_ instead of _scripts_?

mxsdev:
> it doesn't seem like switching between parallel and sequential execution is not supported

See the PR description:

> Implementation is slightly different than the PR description. It runs the provided arguments in the order given, unless -p is provided, in which case it runs them in parallel.

Basically, it either runs them all in serial or all in parallel. I'm considering that out of scope for now, since it's not supported well in cobra.

At any rate, can't you do like `runme run a b c && runme run x y -p` if you want multiple "stages"?

> should we word it as workflows instead of scripts?

@sourishkrout Could use your input here

Everything else I will address 👍

christian-bromann:
> At any rate, can't you do like `runme run a b c && runme run x y -p` if you want multiple "stages"?

You are right, this would totally provide the same experience even though it would be more verbose. I still believe using positional arguments can enable us much more use cases here, e.g. passing in arguments or setting environment variables for making scripts more dynamic, e.g.:

```
runme run \
    workflowA --env="FOO=bar" --args=foobar
    workflowB --env="FOO=bar" --args=foobar --env="FOO=bar" --args=foobar
    ...
```

where

    ```sh { name=workflowA }
    echo "environment variable FOO=$FOO and argument $0"
    ```

would output `environment variable FOO=bar and argument foobar`.

Anyway, happy to see multi command support moving forward.

sourishkrout:
> > Implementation is slightly different than the PR description. It runs the provided arguments in the order given, unless -p is provided, in which case it runs them in parallel.
> 
> Basically, it either runs them all in serial or all in parallel. I'm indeed considering the previously suggested behavior out of scope for now, since it's not supported well in cobra.
> 
> At any rate, can't you do like `runme run a b c && runme run x y -p` if you want multiple "stages"?

As @mxsdev said, the feature is flexible enough and sufficient to enable the required use cases. It fits with the constraints of the CLI framework and it is mergeable now. A migration to a different CLI framework is disproportional at this point in time.

mxsdev:
OK, I'll make the following UX changes:

 - Help indicates you can put multiple commands
 - Replace "scripts" with "tasks"
 - Add check or error indicator on finish

And then I'll merge

PR #241, opened 2023-05-10
Add success message

sourishkrout:
Error scenarios are currently not bubbled up to the bubbletea program. Merging this for now but needs a bit of rework, please. Would appreciate your help @mxsdev.

### DIFF
--- a/internal/cmd/suggest.go
+++ b/internal/cmd/suggest.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"fmt"
 	"log"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


internal/cmd/suggest.go
```diff
@@ -77,7 +78,13 @@ or just bad. Please use with discretion.
 			}
 
 			model := tui.NewListModel(ctx, description, repoUser, client)
-			return newProgram(cmd, model).Start()
+			err = newProgram(cmd, model).Start()
+
+			if err == nil {
```
sourishkrout, line 83: Neither this `err` nor model's err (if you expose it through eg `model.Err()`) ever reflect the internal error state. Not sure why.

mxsdev:
@sourishkrout Can you give an example test case?

sourishkrout:
> @sourishkrout Can you give an example test case?

Oh snap. I ran the API locally and made the rate-limiting ultra-sensitive. Lemme think if there's an easier way.

sourishkrout:
Here's the issue. Please note how when the error occurs, it's being handled inside the Bubbletea program and upon quitting the error does not bubble up. It's probably a bubbletea thing. Not super important @mxsdev.

https://github.com/stateful/runme/assets/250527/6444bc54-5ee2-466c-8c6d-30edbf8b315e

PR #243, opened 2023-05-15
Move to project-based model

mxsdev:
Moves to project-based model from file-centric one.


mxsdev:
@sourishkrout 

> Could we get an ENV interface to set --project such as RUNME_PROJECT=. to toggle the opt-in globally?

Done!

> Ahead of working through the UI/UX challenges with the project-level listing of commands, can we filter out unnamed cells by default (only in project mode for now)? Perhaps add a flag to include them.

Added with `--allow-unnamed`. Defaults to false in project mode, and true in file mode (for compat purposes)

sourishkrout:
This is awesome @mxsdev! Thank you. If @duvanmonsa, you could focus on a code review. @christian-bromann and I will make sure the functionality aligns with the use cases were are pursuing.

sourishkrout:
@mxsdev is it possible to go from least to most nested for ordering per group? The order within a markdown group should continue to follow the natural order (as mentioned in file).

mxsdev:
> is it possible to go from least to most nested for ordering per group?

By this, do you mean that scripts from the same file should appear in the order they are in the file? Or that the files deepest in the file hierarchy should appear at the bottom?

Either of these is do-able, at any rate.

sourishkrout:
> > is it possible to go from least to most nested for ordering per group?
> 
> By this, do you mean that scripts from the same file should appear in the order they are in the file? Or that the files deepest in the file hierarchy should appear at the bottom?
> 
> Either of these (I think) is do-able, at any rate.

Both. I believe the former is already the case (let's be sure). Yep, "files deepest in the file hierarchy should appear at the bottom", is what It'd like, please. I'd expect above's listing to look as follows:

```
NAME	FILE	FIRST COMMAND	# OF COMMANDS	DESCRIPTION
brew-bundle	README.md	brew bundle --no-lock	1	Ensure you have dev setup and runningInstall MacOS dependencies
precommit	README.md	pre-commit install	1	Setup pre-commit
git-submodules	README.md	git submodule update --init	1	Initialize or update git submodules
pipenv-install	README.md	pipenv install --dev --keep-outdated	1	Install project dependencies
setup-config	README.md	rm -f .env	2	Setup environment
run-api	README.md	pipenv run api	1	Run api
run-tasks	README.md	pipenv run tasks	1	Or tasks
run-ingress	README.md	pipenv run ingress	1	Or ingress
run-tests	README.md	pipenv run test	1	Run tests
new-migration	submodules/dovecotes/README.md	pipenv run bash -c 'TIMESCALE_DATABASE_URL=postgresql+psycopg2://postgres:postgres@127.0.0.1:15430/platform alembic revision -m "Some message" --autogenerate'	1	Generate new migration.
upgrade	submodules/dovecotes/README.md	pipenv run bash -c 'TIMESCALE_DATABASE_URL=postgresql+psycopg2://postgres:postgres@127.0.0.1:15430/platform alembic upgrade head'	1	Run migrations in local dev.
downgrade	submodules/dovecotes/README.md	pipenv run bash -c 'TIMESCALE_DATABASE_URL=postgresql+psycopg2://postgres:postgres@127.0.0.1:15430/platform alembic downgrade -1'	1	Remember to also test downgrade.
upgrade-stg	submodules/dovecotes/README.md	pipenv run bash -c 'CLOUDSDK_CORE_PROJECT=staging-4fa8 alembic upgrade head'	1	Run migrations in staging.
```

christian-bromann:
> * If the command is not unique, you are given a tui selector prompt to select the one you'd like

What happens in CI? Would we throw an error?

> NAME	FILE	FIRST COMMAND	# OF COMMANDS	DESCRIPTION

I wonder if we want to get rid of `FIRST COMMAND` and `# OF COMMANDS` in project mode as they seem to make not much sense in this context.

sourishkrout:
> > * If the command is not unique, you are given a tui selector prompt to select the one you'd like
> 
> What happens in CI? Would we throw an error?
> 
> > NAME	FILE	FIRST COMMAND	# OF COMMANDS	DESCRIPTION
> 
> I wonder if we want to get rid of `FIRST COMMAND` and `# OF COMMANDS` in project mode as they seem to make not much sense in this context.

Is it possible to check if a tty is attached and, if not, fail?

mxsdev:
> What happens in CI? Would we throw an error?

> Is it possible to check if a tty is attached and, if not, fail?

Yes, there is a line to detect TTY and fail in that case (there is a special error message for this).

> I wonder if we want to get rid of FIRST COMMAND and # OF COMMANDS in project mode as they seem to make not much sense in this context.

First command might still be useful though to tell what a task actually does, no? # of commands I agree is probably useless clutter most of the time

internal/cmd/run.go
```diff
@@ -43,34 +49,62 @@ func runCmd() *cobra.Command {
...
 						if err != nil {
 							return err
 						}
 
-						if err := replace(replaceScripts, block.Lines()); err != nil {
+						var block project.CodeBlock
+
+						if len(blocks) > 1 {
+							if !isTerminal(os.Stdout.Fd()) {
+								return errors.New("multiple matches found for code block; please use a file specifier in the form \"{file}:{task-name}\"")
```
mxsdev, line 95: Here is the line for reference @christian-bromann

sourishkrout:
> # of commands I agree is probably useless clutter most of the time

Agreed. let's remove it. The first command, we'll keep for now.

internal/cmd/run.go
christian-bromann: Awesome 👍

mxsdev:
Just pushed a commit removing # of commands.

Also a commit which will sort in ascending nested order 👍 

As for maintaining README order, it does an in-order DFS on goldmark's AST. Goldmark's ast appears to itself have nodes in order of appearance, though there may be edge cases. 

I could implement a sort on node position, but I don't think its worth the effort, especially since this is likely to itself create more problems. I think it will be fine to use the default order for now, and look into it if issues arise.

Finally, I will fix the tests before merge, after implementation have stabilized

sourishkrout:
Awesome @mxsdev. I should have probably mentioned that the TUI behave the same as `list`. Which it does not after this change?

> As for maintaining README order, it does an in-order DFS on goldmark's AST. Goldmark's ast appears to itself have nodes in order of appearance, though there may be edge cases.

No need right now. Thank you, though. The way `runme list` works right now is excellent.

mxsdev:
> TUI 

Good catch!

sourishkrout:
Getting a panic with this command `runme run -p precommit brew-bundle setup-config` in https://github.com/stateful/baldaquin#main (internal repo).

sourishkrout:
Looks like `runme print` is also broken. Unless, it's a simple fix we might just want to prioritize a merge and treat it as a "bug" (as long as it's opt-in).

mxsdev:
> Getting a panic with this command runme run -p precommit brew-bundle setup-config

Unfortunately I was unable to reproduce:

image

> Looks like runme print is also broken. Unless, it's a simple fix we might just want to prioritize a merge and treat it as a "bug" (as long as it's opt-in).

Simple fix, pushed

sourishkrout:
Gimme a sec. I'll try again.

sourishkrout:
> runme run -p precommit brew-bundle setup-config

No longer panics for me either 🤷. Odd. I'll keep an eye on it.

internal/project/repo.go
```diff
@@ -0,0 +1,221 @@
...
+
+	// once sync.Once
+	// repo *repo
+}
+
+func NewResolver(dir string) *Resolver {
+	return &Resolver{cwd: dir}
+}
+
+// func (r *Resolver) openRepo() *repo {
```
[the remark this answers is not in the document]
mxsdev: Sure, I guess it will always be there in the git log.

Not sure to be honest why it was there in the first place

sourishkrout (approved):
@mxsdev: I tested some more and couldn't repro the panic. Let's merge please for now so we can pressure test it more as long as its opt-in.

Issue #246, opened 2023-05-15
Support new flag run -all


duvanmonsa:
- [x] Add new flag --all for run cmd
- [x] Adding confirmation before run all the tasks
- [x] By passing --all the cell name parameter is not required. 
- [x] Implement all the logic required to run all the cells inside the document. 
- [x] Make sure multi-runner uses the same session probably solution will be using grpc server

duvanmonsa:
Thank you @mxsdev for doing the major work on this ticket.

PR #248, opened 2023-05-15
feat: adding --all --category flag to run cmd

duvanmonsa:
Adding new flag --all to run all the tasks

image

image

image

Added new flag --category my-category
image
Filtering category
image

No task found 
image

### DIFF
--- a/internal/cmd/run.go
+++ b/internal/cmd/run.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"os/signal"
+	"strconv"
 	"strings"
 	"syscall"
 
@@ -16,11 +17,13 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/stateful/runme/internal/document"
 	"github.com/stateful/runme/internal/runner/client"
+	"github.com/stateful/runme/internal/tui"
 )
 
 func runCmd() *cobra.Command {
 	var (
 		dryRun         bool
+		runAll         bool
 		parallel       bool
 		replaceScripts []string
 		serverAddr     string
@@ -35,7 +38,7 @@ func runCmd() *cobra.Command {
 		Args:              cobra.ArbitraryArgs,
 		ValidArgsFunction: validCmdNames,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if len(args) == 0 {
+			if !runAll && len(args) == 0 {
 				return errors.New("must provide at least one command to run")
 			}
 
@@ -47,17 +50,28 @@ func runCmd() *cobra.Command {
 					return err
 				}
 
-				for _, arg := range args {
-					block, err := lookupCodeBlock(blocks, arg)
-					if err != nil {
-						return err
-					}
+				if runAll {
+					runBlocks = blocks
+				} else {
+					for _, arg := range args {
+						block, err := lookupCodeBlock(blocks, arg)
+						if err != nil {
+							return err
+						}
 
-					if err := replace(replaceScripts, block.Lines()); err != nil {
-						return err
-					}
+						if err := replace(replaceScripts, block.Lines()); err != nil {
+							return err
+						}
 
-					runBlocks = append(runBlocks, block)
+						runBlocks = append(runBlocks, block)
+					}
+				}
+			}
+
+			if runAll {
+				err := confirmExecution(cmd, strconv.Itoa(len(runBlocks)), parallel)
+				if err != nil {
+					return err
 				}
 			}
 
@@ -123,10 +137,16 @@ func runCmd() *cobra.Command {
 						blockNames[i] = block.Name()
 						blockNames[i] = blockColor.Sprint(blockNames[i])
 					}
+					if runAll && parallel {
+						blockNames = []string{blockColor.Sprint("all tasks")}
+					}
 
 					scriptRunText := "Running task"
+					if runAll && parallel {
+						scriptRunText = "Running"
+					}
 
-					if len(blocks) > 1 {
+					if len(blocks) > 1 && !runAll {
 						scriptRunText += "s"
 					}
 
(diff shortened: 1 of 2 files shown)


duvanmonsa:
https://github.com/stateful/runme/issues/246

internal/document/block.go
```diff
@@ -135,6 +135,19 @@ func (b *CodeBlock) Value() []byte {
 	return b.value
 }
 
+func (b *CodeBlock) Category() string {
+	return b.Attributes()["category"]
+}
+
+func (b *CodeBlock) IsExcludedFromRunAll() bool {
```
mxsdev, line 142:
```suggestion
func (b *CodeBlock) ExcludeFromRunAll() bool {
```

Consistent naming convention 😉

internal/cmd/run.go
```diff
@@ -254,3 +287,29 @@ func inRawMode(cb func() error) error {
 
 	return err
 }
+
+func confirmExecution(cmd *cobra.Command, numberTask string, parallel bool) error {
+	text := "Run all " + numberTask + " tasks"
+	if parallel {
+		text += " (in parallel)"
+	}
```
mxsdev, line 295:
```suggestion
func confirmExecution(cmd *cobra.Command, numTasks int, parallel bool) error {
	text := fmt.Sprintf("Run all %d tasks", numTasks)
	if parallel {
		text += " (in parallel)"
	}
```

internal/cmd/run.go
```diff
@@ -254,3 +287,29 @@ func inRawMode(cb func() error) error {
...
+	text += "?"
+
+	model := tui.NewStandaloneQuestionModel(
+		text,
+		tui.MinimalKeyMap,
+		tui.DefaultStyles,
+	)
+	finalModel, err := newProgram(cmd, model).Run()
+	if err != nil {
+		return errors.Wrap(err, "failed to prompt")
```
mxsdev, line 306:
```suggestion
		return errors.Wrap(err, "cli program failed")
```

internal/cmd/run.go
```diff
@@ -47,17 +51,38 @@ func runCmd() *cobra.Command {
...
+						}
+
+						runBlocks = append(runBlocks, block)
 					}
+				}
+			}
 
-					runBlocks = append(runBlocks, block)
+			if runAll {
+				err := confirmExecution(cmd, strconv.Itoa(len(runBlocks)), parallel)
```
mxsdev, line 83: fyi its pretty uncommon to use `strconv.Itoa`, its much more common to use something like `fmt.Sprintf("%d")`

Generally `Sprint` is the "swiss army knife" of string formatting in golang

internal/cmd/run.go
```diff
@@ -123,10 +148,16 @@ func runCmd() *cobra.Command {
...
 						blockNames[i] = blockColor.Sprint(blockNames[i])
 					}
+					if runAll && parallel {
+						blockNames = []string{blockColor.Sprint("all tasks")}
+					}
 
 					scriptRunText := "Running task"
+					if runAll && parallel {
+						scriptRunText = "Running"
+					}
```
mxsdev, line 158:
```suggestion
					scriptRunText := "Running task"
					if runAll && parallel {
						scriptRunText = "Running"
						blockNames = []string{blockColor.Sprint("all tasks")}
					}
```

You can bring these into one

mxsdev (changes requested):
See comments.

One other thing: I think it might make more sense to not require the `all` flag if `category` is present. So for instance I feel I should be able to do `runme run --category cat1` and have it work. What do you think?

mxsdev (changes requested):
See comments.

One other thing: I think it might make more sense to not require the `all` flag if `category` is present. So for instance I feel I should be able to do `runme run --category cat1` and have it work. What do you think?

duvanmonsa:
> See comments.
> 
> One other thing: I think it might make more sense to not require the `all` flag if `category` is present. So for instance I feel I should be able to do `runme run --category cat1` and have it work. What do you think?

I agree with this, I will do the change.

internal/cmd/run.go
duvanmonsa: Thank you I removed already

internal/cmd/run.go
duvanmonsa: Good catch

mxsdev (changes requested):
A few more things 😅

internal/cmd/run.go
```diff
@@ -125,8 +149,12 @@ func runCmd() *cobra.Command {
 					}
 
 					scriptRunText := "Running task"
+					if (runAll || category != "") && parallel {
+						scriptRunText = "Running"
+						blockNames = []string{blockColor.Sprint("all tasks")}
```
mxsdev, line 154: Might be a good idea to specify the category here. So for example like "cat1 tasks" for category cat1, instead of "all tasks"

internal/cmd/run.go
```diff
@@ -47,17 +50,38 @@ func runCmd() *cobra.Command {
 					return err
 				}
 
-				for _, arg := range args {
-					block, err := lookupCodeBlock(blocks, arg)
-					if err != nil {
-						return err
+				if runAll || category != "" {
+					for _, block := range blocks {
+						if !block.ExcludeFromRunAll() {
```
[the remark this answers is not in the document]
duvanmonsa: My feeling here is that using the --category applies the same behavior that --all so it will exclude it as well @sourishkrout do you have feedback on this one?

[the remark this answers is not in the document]
mxsdev: That's fine, I just wanted to make sure that was an intentional choice.

internal/cmd/run.go
```diff
@@ -149,12 +150,15 @@ func runCmd() *cobra.Command {
 					}
 
 					scriptRunText := "Running task"
-					if (runAll || category != "") && parallel {
+					if runMany && parallel {
 						scriptRunText = "Running"
 						blockNames = []string{blockColor.Sprint("all tasks")}
+						if category != "" {
+							blockNames = []string{blockColor.Sprintf("for category %s", category)}
```
mxsdev, line 157:
```suggestion
scriptRunText += " tasks"
blockNames = []string{blockColor.Sprintf("for category %s", category)}
```

So that it says

> Running tasks for category cat1 in parallel...

Instead of

> Running for category cat1 in parallel...

mxsdev (changes requested):

mxsdev (approved):

PR #250, opened 2023-05-17
internal/runner: add option to store last stdout in `$__`

mxsdev:
Adds field `store_last_output` to ExecuteRequest, which stores last stdout in the `__` variable of its session.

Currently not worried about non-gRPC support (local runner scenario). 

To test this out, make sure to run `runme run`/`runme tui` in server mode.


sourishkrout:
Tested with the server-attached and got the output. However, do we add an extra newline at the end?

```
❯ runme run hello-world
hello world
❯ runme run echo-
hello world

~/Projects/stateful/oss/runme mxsdev/reference-last *11 !2 ?1 ❯ 
```

sourishkrout:
Also, @mxsdev, will this feature ultimately be available in the TUI?

mxsdev:
> However, do we add an extra newline at the end?

`echo` prints a new line at the end, you'll have to use `-n` to suppress that.

> Also, @mxsdev, will this feature ultimately be available in the TUI?

Should be working in the TUI currently, but you need session-strategy set to recent, so that the environment vars persist.

sourishkrout:
> > Also, @mxsdev, will this feature ultimately be available in the TUI?
> 
> Theoretically should be working in the TUI currently. (With gRPC server)

What the reason for the TUI not persisting ENV across runs by default? Please refresh my mind, I can't recall right now.

mxsdev:
@sourishkrout 

> What the reason for the TUI not persisting ENV across runs by default? Please refresh my mind, I can't recall right now.

This does happen without the server, but as per the main post:

> Currently not worried about non-gRPC support (local runner scenario).

In other words, I've only worried about gRPC implementation. If you want it in local runner let me know, its just more work.

sourishkrout:
> @sourishkrout
> 
> > What the reason for the TUI not persisting ENV across runs by default? Please refresh my mind, I can't recall right now.
> 
> Envs are persisted without the server, but as per the main post:
> 
> > Currently not worried about non-gRPC support (local runner scenario).
> 
> In other words, I've only done the gRPC implementation. If you think tui + local runner scenario is important let me know, it can be done.

👍 gotcha. Makes sense. Let's merge what we have and add local-runner-support in a follow-up PR.

PR #253, opened 2023-05-19
feat: prompt variables on cli

duvanmonsa:
Prompting for variables before execute blocks.

Placeholder without quotes 
image
With quotes use the value 
image

### DIFF
--- a/internal/cmd/run.go
+++ b/internal/cmd/run.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"os/signal"
+	"regexp"
 	"strings"
 	"syscall"
 
@@ -17,7 +18,20 @@ import (
 	"github.com/stateful/runme/internal/document"
 	"github.com/stateful/runme/internal/runner/client"
 	"github.com/stateful/runme/internal/tui"
+	"github.com/stateful/runme/internal/tui/prompt"
 )
+
+const (
+	exportExtractRegex string = `(\n*)export (\w+=)(("[^"]*")|('[^']*')|[^;\n]+)`
+)
+
+type CommandExportExtractMatch struct {
+	Key            string
+	Value          string
+	Match          string
+	HasStringValue bool
+	LineNumber     int
+}
 
 func runCmd() *cobra.Command {
 	var (
@@ -75,6 +89,20 @@ func runCmd() *cobra.Command {
 						}
 
 						runBlocks = append(runBlocks, block)
+					}
+				}
+			}
+
+			for _, block := range runBlocks {
+				if block.PromptEnv() {
+					varPrompts := getCommandExportExtractMatches(block.Lines())
+					for _, ev := range varPrompts {
+						newVal, err := promptForEnvVar(cmd, ev)
+						block.SetLine(ev.LineNumber, replaceVarValue(ev, newVal))
+
+						if err != nil {
+							return err
+						}
 					}
 				}
 			}
@@ -211,7 +239,6 @@ func runCmd() *cobra.Command {
 				if len(runBlocks) > 1 {
 					return multiRunner.RunBlocks(ctx, runBlocks, parallel)
 				}
-
 				return runner.RunBlock(ctx, runBlocks[0])
 			})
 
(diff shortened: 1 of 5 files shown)


internal/cmd/run.go
```diff
@@ -79,6 +93,20 @@ func runCmd() *cobra.Command {
 				}
 			}
 
+			for _, block := range runBlocks {
+				if block.PromptEnv() {
+					varPrompts := getCommandExportExtractMatches(block.Lines())
+					for _, ev := range varPrompts {
+						newVal, err := promptForEnvVar(cmd, ev)
+						block.SetLine(ev.LineNumber, replaceVarValue(ev, newVal))
```
mxsdev, line 101: Modifying `lines` may cause issues, because for example in TUI, the same script may be run multiple times. This will persist the environment variables instead of letting the user select each time.

I think we should implement a `Clone` method on `document.CodeBlock` so that you don't modify the underlying code block.

internal/document/block.go
```diff
@@ -135,10 +135,23 @@ func (b *CodeBlock) Value() []byte {
...
+}
+
 func (b *CodeBlock) Category() string {
 	return b.Attributes()["category"]
 }
 
+func (b *CodeBlock) PromptEnv() bool {
+	val, err := strconv.ParseBool(b.Attributes()["promptEnv"])
+	if err != nil {
+		return false
```
mxsdev, line 149:
```suggestion
		return true
```

In VSCode this is true by default

internal/cmd/run.go
```diff
@@ -317,3 +344,60 @@ func confirmExecution(cmd *cobra.Command, numTasks int, parallel bool, category
...
+	test := regexp.MustCompile(exportExtractRegex)
+	result := []CommandExportExtractMatch{}
+
+	for i, line := range lines {
+		for _, match := range test.FindAllStringSubmatch(line, -1) {
+			e := match[0]
+
+			parts := strings.Split(strings.TrimSpace(e[len("export "):]), "=")
+			key := parts[0]
+			ph := parts[1]
```
mxsdev, line 383: This can theoretically panic. You might want to ensure `len(parts) > 1` and if not throw a more useful error.

internal/cmd/run.go
```diff
@@ -317,3 +344,60 @@ func confirmExecution(cmd *cobra.Command, numTasks int, parallel bool, category
...
+
+func getCommandExportExtractMatches(lines []string) []CommandExportExtractMatch {
+	test := regexp.MustCompile(exportExtractRegex)
+	result := []CommandExportExtractMatch{}
+
+	for i, line := range lines {
+		for _, match := range test.FindAllStringSubmatch(line, -1) {
+			e := match[0]
+
+			parts := strings.Split(strings.TrimSpace(e[len("export "):]), "=")
```
mxsdev, line 381:
```suggestion
			parts := strings.Split(strings.TrimSpace(e)[len("export "):], "=")
```

The javascript impl is subtly different 😉

internal/cmd/run.go
```diff
@@ -317,3 +344,60 @@ func confirmExecution(cmd *cobra.Command, numTasks int, parallel bool, category
 
 	return nil
 }
+
+func promptForEnvVar(cmd *cobra.Command, ev CommandExportExtractMatch) (string, error) {
+	label := fmt.Sprintf("Set Environment Variable \" %s\"", ev.Key)
```
mxsdev, line 349:
```suggestion
	label := fmt.Sprintf("Set Environment Variable %q", ev.Key)
```

You added a space here by mistake

mxsdev, line 349:
```suggestion
	label := fmt.Sprintf("Set Environment Variable \" %s\":", ev.Key)
```

I think this looks cleaner w/ a colon

internal/cmd/run.go
```diff
@@ -317,3 +344,60 @@ func confirmExecution(cmd *cobra.Command, numTasks int, parallel bool, category
...
+	val, ok := finalModel.(tui.StandaloneInputModel).Value()
+	if !ok {
+		return "", errors.New("canceled")
+	}
+	return val, nil
+}
+
+func replaceVarValue(ev CommandExportExtractMatch, newValue string) string {
+	parts := strings.SplitN(ev.Match, "=", 2)
+	replacedText := fmt.Sprintf("%s=%s", parts[0], newValue)
```
mxsdev, line 369:
```suggestion
	replacedText := fmt.Sprintf("%s=%q", parts[0], newValue)
```

RHS should be quoted

mxsdev (changes requested):
You forgot about the tui!

internal/cmd/run.go
duvanmonsa: i love this `q`

internal/cmd/run.go
duvanmonsa: It was working lol I already change it

duvanmonsa:
> You forgot about the tui!

@mxsdev then I forgot it as well for `--all and --category`

internal/cmd/common.go
```diff
@@ -276,3 +279,80 @@ type runFunc func(context.Context) error
 const tlsFileMode = os.FileMode(int(0o700))
 
 var defaultTLSDir = filepath.Join(GetDefaultConfigHome(), "tls")
+
+func promptEnvVars(cmd *cobra.Command, runBlocks []project.FileCodeBlock) error {
```
mxsdev, line 283:
```suggestion
func promptEnvVars(cmd *cobra.Command, runBlocks... project.FileCodeBlock) error {
```

This syntax prevents the awkward conversion to a list for single blocks

internal/cmd/tui.go
```diff
@@ -133,7 +133,14 @@ func tuiCmd() *cobra.Command {
 						return err
 					}
 
-					err = runnerClient.RunBlock(ctx, result.block)
+					tempBlock := result.block.Block.Clone()
```
mxsdev, line 136:
```suggestion
					tempBlock := &result.block.Block.Clone()
```

This is only ever accessed as a pointer, so you might as well address here

duvanmonsa:
https://github.com/orgs/stateful/projects/6/views/1?pane=issue&itemId=2825847

internal/cmd/tui.go
```diff
@@ -133,7 +133,14 @@ func tuiCmd() *cobra.Command {
...
+
+					err = promptEnvVars(cmd, []project.FileCodeBlock{&tempBlock})
```
mxsdev, line 138: You should take this code block out of `inRawMode` since I don't think bubbletea is meant to run in raw mode.

internal/cmd/tui.go
mxsdev, line 136:
```suggestion
					runBlock := result.block.Block.Clone()
```

Minor nitpick, but I personally prefer the name "runBlock" here

internal/cmd/tui.go
```diff
@@ -133,7 +133,14 @@ func tuiCmd() *cobra.Command {
...
+					if err != nil {
+						return err
+					}
+
+					err = runnerClient.RunBlock(ctx, &tempBlock)
```
mxsdev, line 143: Calling `RunBlock` directly on the block prevents the runner from determining the correct location for `cwd`, since `document.Block` has no indication of file location.

I think implementing a `Clone` method for `project.CodeBlock` and using that instead would fix this issue.

mxsdev (changes requested):

internal/cmd/tui.go
duvanmonsa: image
It is now part of the clone method. So I removed the pointers here.

internal/cmd/tui.go
```diff
@@ -123,6 +123,13 @@ func tuiCmd() *cobra.Command {
 
 				ctx, cancel := ctxWithSigCancel(cmd.Context())
 
+				runBlock := result.block.Clone()
+
+				err = promptEnvVars(cmd, []project.FileCodeBlock{runBlock}...)
```
mxsdev, line 128:
```suggestion
err = promptEnvVars(cmd, runBlock)
```

mxsdev (changes requested):

mxsdev (approved):

PR #261, opened 2023-05-22
internal/cmd/run: switch to `/` for filename specifier

### LINKED ISSUE #258: Can't find command
For some reasons Runme is not able to find my `test` command even though it is clearly defined and visible when calling the `list` command:

Screenshot 2023-05-22 at 14 00 50

mxsdev:
Should fix #258

### DIFF
--- a/internal/cmd/run.go
+++ b/internal/cmd/run.go
@@ -303,7 +303,7 @@ func inRawMode(cb func() error) error {
 	return err
 }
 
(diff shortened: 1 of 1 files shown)


sourishkrout (approved):
👍 LGTM

sourishkrout:
@christian-bromann you can either build from source, or we can put out an RC. What would you prefer? Don't want to do a full release for little changes.

PR #266, opened 2023-05-23
support custom shells

mxsdev:
Users can use the `shell` frontmatter option to set a default shell for all cells in the document.

For demonstration purposes:
```sh
go run . --project "./examples/frontmatter/shells" --allow-unnamed
```

I've also exposed a new `Frontmatter` option in the gRPC serializer `Notebook` struct, which can be used to provide additional support in the extension.


sourishkrout (approved):
👍 LGTM

PR #270, opened 2023-05-25
file/annotation specific cwd

mxsdev:
Implementation is about what you'd expect, though there is an ambiguity around absolute paths. One option is to resolve absolute paths relative to the project, and the other is to resolve them as absolute paths.

I decided to resolve absolute paths as system-absolute paths, because the "project" directory is currently not really stable.

Resolution order goes as follows:

 - Current cwd
 - Markdown file folder
 - Frontmatter `cwd`
 - Annotation `cwd`

### DIFF
--- a/examples/frontmatter/cwd/RELATIVE.md
+++ b/examples/frontmatter/cwd/RELATIVE.md
@@ -1,0 +1,15 @@
+---
+cwd: ../../../
+---
+
+```sh { name=relative-pwd }
+pwd
+```
+
+```sh { name=relative-rel-pwd cwd=../}
+pwd
+```
+
+```sh { name=relative-abs-pwd cwd=/opt}
+pwd
+```
--- a/examples/frontmatter/cwd/ABSOLUTE.md
+++ b/examples/frontmatter/cwd/ABSOLUTE.md
@@ -1,0 +1,15 @@
+---
+cwd: /tmp
+---
+
+```sh { name=absolute-pwd }
+pwd
+```
+
+```sh { name=absolute-rel-pwd cwd=../}
+pwd
+```
+
+```sh { name=absolute-abs-pwd cwd=/opt}
+pwd
+```
--- a/examples/frontmatter/cwd/NONE.md
+++ b/examples/frontmatter/cwd/NONE.md
@@ -1,0 +1,11 @@
+```sh { name=none-pwd }
+pwd
+```
+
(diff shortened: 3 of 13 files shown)


examples/frontmatter/cwd/RELATIVE.md
```diff
@@ -0,0 +1,15 @@
+---
+cwd: ../../../
```
christian-bromann, line 2: That's neat 👍

christian-bromann (approved):
LGTM

mxsdev:
@christian-bromann 

> I decided to resolve absolute paths as system-absolute paths, because the "project" directory is currently not really stable.

How do you feel about this? Do you think using primarily relative files will be sufficient for now

sourishkrout:
🤔 that's interesting, I sort of assumed that the markdown file base dir is top of the priority stack?

The absolute path seems like a rare edge case that almost should be discouraged.

mxsdev:
@sourishkrout 

> 🤔 that's interesting, I sort of assumed that the markdown file base dir is top of the priority stack?

The way I described it is somewhat confusing. The resolution process goes down the list top to bottom; if it finds an absolute path, it uses that, and otherwise it resolves relative to the previous path. If the thing in the list doesn't exist, we skip to the next one.

It is internally possible in the code-base for a code-block to run "detatched" from a file which is why I included the cwd in the list. In all runnable files, however, resolution effectively starts from the markdown file base dir, since that is an absolute path, so it will overwrite the cwd in the resolution path.

Does that make sense? 

> The absolute path seems like a rare edge case that almost should be discouraged.

Until runme has a clear notion of what is a "runme project" at a file-system level (like a `.runmerc` configuration file) I'm personally wary of trying to resolve absolute paths relative to project root. Since then the ability to run the markdown file depends not just on the markdown file, but also how and from where it is run (in this case where the user selects for `--project`). More broadly, module resolution is a tough problem - TypeScript is a good case study.

Let me know if you think it is worth coming up with a solution for project-relative addressing in the short-term. Otherwise, in the meantime, I think treating absolute paths as system absolute paths is at least consistent with intuition. 

I agree that this is probably not desired by many (though it would have use for the "local run-book" scenario?) For shared projects, I agree users should be encouraged to use relative paths for now.

sourishkrout:
Roger that on both 👍, @mxsdev.

PR #272, opened 2023-05-25
internal/runner/client: fix multirunner cwd

mxsdev:
Multi-runner was using the old `document.CodeBlock`, which does not have file information. This prevented scripts run with multi-runner from being aware of their file location, so their cwd would be set relative/same as the parent.


sourishkrout:
👍 LGTM

Issue #281, opened 2023-06-01
CLI `run --all` "bugs"


sourishkrout:
Using the canonical example at `vscode-runme/examples`

- [x] The completion of `Set Environment Variable "LICENSE": $(cat ../LICENSE)` should be auto-skipped since it's using shell var population. The extension will skip this particular scenario based via pattern detection.
- [x] Cells with `promptEnv: false` should be skipped.

```
runme run --all
                                                                  
Set Environment Variable "DENO_ACCESS_TOKEN": <insert-token-here> 
                                                                  
                                                     
Set Environment Variable "SERVICE_FOO_TOKEN": foobar 
                                                     
                                                     
Set Environment Variable "SERVICE_BAR_TOKEN": barfoo 
                                                     
                                                  
Set Environment Variable "PATH": /some/path:$PATH 
                                                  
                                                     
Set Environment Variable "LICENSE": $(cat ../LICENSE)
                                                     
                                                                       
Set Environment Variable "PRIVATE_KEY": -----BEGIN RSA PRIVATE KEY---- 
                                                                       
                         
Run all 19 tasks? [Y/n] Y
```

duvanmonsa:
@sourishkrout I was testing skip for `promptEnv: false` and it is fine it is working correctly is there any particular test you did where it didn't work?

sourishkrout:
Interesting, I thought it didn't skip when I encountered an instance. I'll test again and will provide specifics.

Issue #282, opened 2023-06-01
RemoteRunner does not work with `run --all`


sourishkrout:
Let's be sure this works when the kernel is involved long-term. Short-term let's have a better error message.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x28 pc=0x101488e7c]

goroutine 1 [running]:
github.com/stateful/runme/internal/runner/client.(*RemoteRunner).RunBlock(0x14000344bb8, {0x101900918, 0x1400012c140}, {0x101901670?, 0x14000477680?})
        github.com/stateful/runme/internal/runner/client/client_remote.go:109 +0x6c
github.com/stateful/runme/internal/runner/client.MultiRunner.RunBlocks.func3({0x101901670, 0x14000477680})
        github.com/stateful/runme/internal/runner/client/client_multi.go:127 +0x4c
github.com/stateful/runme/internal/runner/client.MultiRunner.RunBlocks({{0x101903328, 0x14000344810}, {0x0, 0x0}, 0x1400012c440, 0x1400012c480, {0x140005c8130, 0x1, 0x1}}, {0x101900918?, ...}, ...)
        github.com/stateful/runme/internal/runner/client/client_multi.go:145 +0x634
github.com/stateful/runme/internal/cmd.runCmd.func1.3()
        github.com/stateful/runme/internal/cmd/run.go:239 +0xb0
github.com/stateful/runme/internal/cmd.inRawMode(0x140004ddc20)
        github.com/stateful/runme/internal/cmd/run.go:316 +0x8c
github.com/stateful/runme/internal/cmd.runCmd.func1(0x1400009ec00, {0x140001855f0, 0x0, 0x1?})
        github.com/stateful/runme/internal/cmd/run.go:237 +0xabc
github.com/spf13/cobra.(*Command).execute(0x1400009ec00, {0x140001855e0, 0x1, 0x1})
        github.com/spf13/cobra@v1.6.1/command.go:916 +0x5c8
github.com/spf13/cobra.(*Command).ExecuteC(0x14000010600)
        github.com/spf13/cobra@v1.6.1/command.go:1044 +0x35c
github.com/spf13/cobra.(*Command).Execute(...)
        github.com/spf13/cobra@v1.6.1/command.go:968
main.root()
        github.com/stateful/runme/main.go:14 +0x128
main.main()
        github.com/stateful/runme/main.go:22 +0x1c
```

PR #283, opened 2023-06-01
chore: skip propmt vars for exec params

duvanmonsa:
Address this issues https://github.com/stateful/runme/issues/281

### DIFF
--- a/internal/cmd/common.go
+++ b/internal/cmd/common.go
@@ -327,6 +327,12 @@ func getCommandExportExtractMatches(lines []string) []CommandExportExtractMatch 
 			}
 			key := parts[0]
 			ph := parts[1]
+
+			isExecValue := strings.HasPrefix(ph, "$(") || strings.HasPrefix(ph, ")")
+			if isExecValue {
+				continue
+			}
+
 			hasStringValue := strings.HasPrefix(ph, "\"") || strings.HasPrefix(ph, "'")
 			placeHolder := ph
 			if hasStringValue {


duvanmonsa:
Also including this small bug here https://github.com/stateful/runme/issues/282

duvanmonsa:
Resolving issue reported here https://github.com/stateful/runme/issues/286#event-9425141935

internal/cmd/common.go
```diff
@@ -321,12 +321,18 @@ func getCommandExportExtractMatches(lines []string) []CommandExportExtractMatch
...
 			ph := parts[1]
+
+			isExecValue := strings.HasPrefix(ph, "$(") || strings.HasPrefix(ph, ")")
+			if isExecValue {
+				continue
+			}
+
 			hasStringValue := strings.HasPrefix(ph, "\"") || strings.HasPrefix(ph, "'")
 			placeHolder := ph
 			if hasStringValue {
```
mxsdev, line 338: First, you should put the `isExecValue` test after the test for a string value, since we can have `"$(command)"`.

Second, the test `strings.HasPrefix(ph, ")")` doesn't make sense - it would be better I think to just remove that test, and test only for the `$(` at the beginning.

Finally, you should test against `strings.TrimSpace(placeHolder)`, since there could be some leading spaces.

mxsdev (changes requested):

internal/cmd/common.go
duvanmonsa: @mxsdev So basically it is fine to have it first because if the command starts like `"$(command)"` it won't apply to this condition  `$(` != `"$(`

duvanmonsa: Leading spaces could be a problem :/

mxsdev: Discussed this offline, `isExecValue` position is totally fine! I misunderstood the original problem.

mxsdev (approved):

PR #291, opened 2023-06-06
Retain underscores or asterisks respectively for emphasis in markdown

sourishkrout:
Let's try to avoid unnecessary diffs as much as possible. Have another fix forthcoming to handle new-line vs no new-line at file endings.


internal/renderer/cmark/cmark.go
```diff
@@ -314,11 +314,21 @@ func (r *renderer) Render(doc ast.Node, source []byte) ([]byte, error) {
...
+			for i := 0; i < p.Lines().Len(); i++ {
+				line := p.Lines().At(i)
+				underscores += bytes.Count(line.Value(source), []byte{'_'})
 			}
+			mark := []byte{'_'}
+			if underscores == 0 {
+				mark = []byte{'*'}
+			}
+			n := node.(*ast.Emphasis)
+			mark = bytes.Repeat(mark, n.Level)
```
mxsdev, line 331: This will fail on the following:

```markdown
[[ *TOC* _TOC_ ]]
```

Since it will read the first asterisk, and assume they are both asterisks.

I looked into goldmark's custom parsing capabilities quite a bit, but I wasn't able to find anything super promising. I thought you could wrap the inline parser and set an attribute on the node, but that didn't seem possible. And the AST transform system doesn't seem flexible enough. Though it still may be possible

mxsdev (changes requested):

internal/renderer/cmark/cmark.go
sourishkrout: As discussed in person, mixed usage like that is rather unlikely, and we're ok unifying one or the other.

PR #292, opened 2023-06-06
internal/document/editor: send code blocks without supported languages

mxsdev:
Currently, the editor service doesn't send code blocks as code blocks if the language is unsupported.

This removes this guardrail, creating code blocks from the editor service regardless of block language.


sourishkrout (approved):

sourishkrout:
👍 LGTM

PR #295, opened 2023-06-07
Retain trailing new line breaks

sourishkrout:
Another attempt to avoid unnecessary diffs which cause confusion.

### DIFF
--- a/internal/document/editor/editor.go
+++ b/internal/document/editor/editor.go
@@ -31,7 +31,7 @@ func Deserialize(data []byte) (*Notebook, error) {
 		contentOffset: sections.ContentOffset,
 	}
 
-	finalLinesBreaks := document.CountFinalLineBreaks(data, []byte{'\n'})
+	finalLinesBreaks := document.CountFinalLineBreaks(data, document.DetectLineBreak(data))
 	notebook.Metadata = map[string]string{
 		PrefixAttributeName(InternalAttributePrefix, document.FinalLineBreaksKey): fmt.Sprint(finalLinesBreaks),
 	}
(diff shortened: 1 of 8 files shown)


internal/document/editor/editor.go
```diff
@@ -40,9 +47,9 @@ func Deserialize(data []byte) (*Notebook, error) {
...
 	var result []byte
 
-	if intro, ok := notebook.Metadata[FrontmatterKey]; ok {
+	if intro, ok := notebook.Metadata[PrefixAttributeName(InternalAttributePrefix, FrontmatterKey)]; ok {
 		intro := []byte(intro)
-		lb := detectLineBreak(intro)
+		lb := document.DetectLineBreak(intro)
 		result = append(
 			intro,
 			append(lb, lb...)...,
```
mxsdev, line 55:
```suggestion
			bytes.Repeat(lb, 2)...,
```

internal/renderer/cmark/cmark.go
```diff
@@ -21,8 +21,18 @@ func Render(doc ast.Node, source []byte) ([]byte, error) {
 		lineBreak = []byte{'\r', '\n'}
 	}
 
+	finalLineBreaks := 1
+	if flb, ok := doc.AttributeString("finalLineBreaks"); ok {
+		val, ok := flb.(int)
+		if !ok {
+			return nil, errors.Errorf("invalid type for %s expected int", "finalLineBreaks")
```
mxsdev, line 28:
```suggestion
			return nil, errors.Errorf("invalid type for %s expected int", document.FinalLineBreaksKey)
```

mxsdev (changes requested):

PR #301, opened 2023-06-13
internal/project: ignore node_modules by default

mxsdev:
Fixes stateful/vscode-runme#608

Introduces `--ignore-pattern` which allows specifying custom ignore patterns in project mode. Think like adding extra lines to `gitignore` (or creating it if it doesn't exist). 

By default, this includes `node_modules`.


sourishkrout (approved):
👍 LGTM

PR #302, opened 2023-06-14
Support custom cell executions, "temp file" mode

mxsdev:
Adds the following options to `ExecuteRequest`:

 - `command_mode` - either `INLINE_SHELL`, which sends commands to `-c` as an argument, or `TEMP_FILE`, which puts the contents of the script into a temporary file in the given directory, and uses the filename as an argument for the script
 - `language_id` - optional language id of code block, used to try and auto-resolve the program to run with
 - `file_extension` - File extension for `TEMP_FILE` mode. If omitted, it tries to resolve this from `language_id`.

Local & remote runners are integrated with this, and will try and auto-resolve the program to use based on the language id. Users can also use the `program` annotation to select a program manually.

Depends on #292


sourishkrout (approved):
👍 LGTM

PR #308, opened 2023-06-16
internal/document: migrate to JSON

mxsdev:
Introduces `attributeParser` internal interface to deal with attribute parsing.

Tries to parse as ~~TOML~~ JSON first, then falls back to our old parsing language written by @adambabik ("babikML").

### DIFF
--- a/internal/document/attributes.go
+++ b/internal/document/attributes.go
@@ -32,7 +32,7 @@ func (p *babikMLParser) ParseAttributes(raw []byte) (Attributes, error) {
 	return p.parseRawAttributes(p.rawAttributes(raw)), nil
 }
 
-func (_ *babikMLParser) parseRawAttributes(source []byte) map[string]string {
+func (*babikMLParser) parseRawAttributes(source []byte) map[string]string {
 	items := bytes.Split(source, []byte{' '})
 	if len(items) == 0 {
 		return nil
@@ -51,7 +51,7 @@ func (_ *babikMLParser) parseRawAttributes(source []byte) map[string]string {
 	return result
 }
 
-func (_ *babikMLParser) rawAttributes(source []byte) []byte {
+func (*babikMLParser) rawAttributes(source []byte) []byte {
 	start, stop := -1, -1
 
 	for i := 0; i < len(source); i++ {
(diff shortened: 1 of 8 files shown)


sourishkrout:
Bye bye 👋  **BabikML**, you served us well!

internal/document/attributes.go
```diff
@@ -0,0 +1,219 @@
...
+	"go.uber.org/multierr"
+	"golang.org/x/exp/slices"
+
+	"github.com/pelletier/go-toml"
+)
+
+type Attributes map[string]string
+
+type attributeParser interface {
+	ParseAttributes(raw []byte) (Attributes, error)
```
adambabik, line 18: As the name already includes `attribute`, it can be simplified:

```go
type attributeParser interface {
	Parse(raw []byte) (Attributes, error)
    Write(attr Attributes, w io.Writer) error
}
```

adambabik (commented):
Overall, there are a few risks:
- What was the reasoning behind TOML? It requires unmaintained TOML v1 and additional operations after parsing and writing. Why not JSON?
- Ideally, there is a test for encoding/decoding attributes that looks like this `attrs == encode(decode(attrs))`. It's not straightforward with the additional operations.

sourishkrout:
> * What was the reasoning behind TOML? It requires unmaintained TOML v1 and additional operations after parsing and writing. Why not JSON?

I guess, we could do `JSON` as long as we curb or restrain the use of nesting objects

internal/document/attributes.go
```diff
@@ -0,0 +1,219 @@
...
+
+	parsedAttr := root["attr"]
+
+	attr := make(Attributes, len(parsedAttr))
+
+	for k, v := range parsedAttr {
+		kind := reflect.TypeOf(v).Kind()
+
+		// primitive type
+		if kind < reflect.Array || kind == reflect.String {
```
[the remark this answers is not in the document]
mxsdev: With json it is easier to just treat the argument as a string (JSON.stringify for non-strings), so I'll probably do that

mxsdev:
Will be switching to JSON

internal/document/attributes.go
```diff
@@ -0,0 +1,198 @@
...
+	Write(attr Attributes, w io.Writer) error
+}
+
+// Original attribute language used by runme prior to v1.3.0
+//
+// Only supports strings, and does not support spaces. Example:
+//
+// { key=value hello=world string_value=2 }
+//
+// Pioneered by Adam Babik
```
sourishkrout, line 26: 👍 trailblazer!

PR #309, opened 2023-06-20
Allow successive cancelations for non-zero exit codes

sourishkrout:
Without this change, the subsequent use of CTRL+C when running from the `tui` will only work when the process exits with error code 0. We want it to work irrespective of the error code.


mxsdev:
Can you give a simple example? I'm a bit confused by the PR description

mxsdev (approved):
Ah, nevermind, I see the issue. Thanks for catching this!

sourishkrout:
> Can you give a simple example? I'm a bit confused by the PR description

👍 . I'll put a little video on it next time.

PR #310, opened 2023-06-22
internal/cmd/fmt: support project mode

mxsdev:
Supports project mode for `runme fmt`. 

Additionally, I made the `--flatten` command true by default, since when set to false, this can break formatter. I added a notice to the help info as well.


sourishkrout (approved):
Makes sense. LGTM 👍 .

PR #313, opened 2023-06-22
async task loader

mxsdev:
Moves the task loader to an asynchronous model, so that loading information can be "streamed" and consumed by e.g. TUI loaders.

Includes a bubbletea model that all the `internal/cmd` should use:


sourishkrout (approved):
Great work 👍 . LGTM.

There's a few cosmetic changes I'd like to make. However, I'll do them post-merge.

PR #314, opened 2023-06-23
internal/runner: sanitize last command output

mxsdev:
Removes null bytes from last command output. Prevents errors in notebook if command output contains null bytes.

You can repro this error on main by using `echo "\0"` - things will break after the first command if a session is used.

### DIFF
--- a/internal/runner/service.go
+++ b/internal/runner/service.go
@@ -1,6 +1,7 @@
 package runner
 
 import (
+	"bytes"
 	"context"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


sourishkrout:
This is the original error message. Just as a reference:

Screenshot 2023-06-23 at 11 40 45 AM

sourishkrout:
Could we add a test case to prevent regressions @mxsdev? Alternatively, I could do an e2e test in the extension.

sourishkrout (changes requested):
See comment.

mxsdev:
@sourishkrout Done 👍

sourishkrout (approved):
👍 LGTM

PR #316, opened 2023-06-27
Better error message when no code blocks were found in project-mode

sourishkrout:
Ahead of making UX changes to better support out-of-the-box use cases, such as showing all code blocks if no named code blocks were found.

### DIFF
--- a/internal/cmd/tui.go
+++ b/internal/cmd/tui.go
@@ -50,7 +50,7 @@ func tuiCmd() *cobra.Command {
 				if !fAllowUnnamed {
 					return errors.Errorf("no named code blocks, consider adding flag --allow-unnamed")
 				}
-				return errors.Errorf("no named code blocks")
+				return errors.Errorf("no code blocks")
 			}
 
 			if visibleEntries <= 0 {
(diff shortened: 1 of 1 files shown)


internal/cmd/tui.go
```diff
@@ -44,7 +44,13 @@ func tuiCmd() *cobra.Command {
...
 
 			if len(blocks) == 0 {
-				return errors.Errorf("no code blocks in %s", fFileName)
+				if fFileMode {
+					return errors.Errorf("no code blocks in %s", fFileName)
+				}
+				if !fAllowUnnamed {
+					return errors.Errorf("no named code blocks, consider adding flag --allow-unnamed")
+				}
+				return errors.Errorf("no named code blocks")
```
mxsdev, line 53:
```suggestion
				return errors.Errorf("no code blocks")
```

At this point, `fAllowUnnamed` is true. So there are no named or unnamed code blocks.

sourishkrout: Am I crazy? I just tested this. Let me look again.

PR #318, opened 2023-06-30
Project mode is default unless chdir or filename flags are being used

sourishkrout:
Question to consider: Will it be too confusing to turn on `--allow-unnamed` in file mode? It makes sense to me because when a file/dir is specified, the intention is to "consider all commands contained in one file". However, some folks might interpret it as an inconsistency. Wdyt @mxsdev?

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -49,7 +49,7 @@ func Root() *cobra.Command {
 
 			fFileMode = cmd.Flags().Changed("chdir") || cmd.Flags().Changed("filename")
 
-			if fFileMode {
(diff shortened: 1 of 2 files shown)


sourishkrout:
Working tests.

mxsdev:
@sourishkrout 

> Will it be too confusing to turn on --allow-unnamed in file mode? 

I think that's fine, consistency is valuable, but I think your initial expectation on the user's instincts is accurate. Two-way door anyhow 🚪

mxsdev:
> Also, in the case where --index & --filename is being used it's actually important to consider all blocks vs just named ones.

Right, good catch!

mxsdev (approved):

PR #319, opened 2023-07-03
don't prompt if already in session

mxsdev:
Prevents prompting for environment variables when they are already in session at the beginning.

Part of stateful/vscode-runme#643


sourishkrout (approved):
Works 👍. Let's merge.

PR #320, opened 2023-07-05
Retain the legacy attribute parsing for upcoming release

sourishkrout:
Decided it makes more sense to switch attribute formats for v2 (more straightforward to explain). For the pending release of project mode, we want to continue using BabikML.

This is a stop-gap PR to get the release out and then "revert this commit" for the upcoming v2. Obviously, it's not great to have tests against JSON attributes, whereas the app logic uses BabikML. However, there's little value in rewriting the tests for a 1-2 weeks lifetime.


mxsdev (approved):
LGTM, not as worried about tests since we'll just be reverting soon anyways 👍

PR #323, opened 2023-07-07
fix: Adding required header on cloud to generate tokens

duvanmonsa:
Adding header required on cloud-backend to generate tokens


mxsdev (approved):
LGTM, should probably have been this way originally

PR #325, opened 2023-07-10
(core): automatically update Homebrew core formula with every release

christian-bromann:
I've never used this Github Action so let's pay attention on the next release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -59,3 +59,13 @@ jobs:
       - name: Copy to latest
         if: env.pre_release == 'false'
         run: gsutil -m cp "gs://runme-artifacts-41eac6/${{ env.version }}/*" gs://runme-artifacts-41eac6/latest
+      - name: Bump Homebrew Formula
+        uses: mislav/bump-homebrew-formula-action@v2
+        # skip prereleases
+        if: ${{ !contains(github.ref, '-') }}
+        with:
+          formula-name: runme
+          homebrew-tap: Homebrew/homebrew-core
+          base-branch: master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -65,7 +65,5 @@ jobs:
         if: ${{ !contains(github.ref, '-') }}
         with:
           formula-name: runme
-          homebrew-tap: Homebrew/homebrew-core
-          base-branch: master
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -59,3 +59,13 @@ jobs:
...
         if: env.pre_release == 'false'
         run: gsutil -m cp "gs://runme-artifacts-41eac6/${{ env.version }}/*" gs://runme-artifacts-41eac6/latest
+      - name: Bump Homebrew Formula
+        uses: mislav/bump-homebrew-formula-action@v2
+        # skip prereleases
+        if: ${{ !contains(github.ref, '-') }}
+        with:
+          formula-name: runme
+          homebrew-tap: Homebrew/homebrew-core
+          base-branch: master
```
mxsdev, line 69:
```suggestion
          base-branch: main
```

.github/workflows/release.yml
```diff
@@ -59,3 +59,13 @@ jobs:
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
mxsdev, line 71: According to the [docs for this action](https://github.com/mislav/bump-homebrew-formula-action):

> You should enable GITHUB_TOKEN only if the repository that runs this Action is private and if COMMITTER_TOKEN has the public_repo scope only. GITHUB_TOKEN will be used for verifying the SHA256 sum of the downloadable archive for this release.

Apparently we're supposed to create a new personal access token for this purpose, with the `repo` and `workflow` scopes.

mxsdev (changes requested):

.github/workflows/release.yml
christian-bromann: This refers to https://github.com/Homebrew/homebrew-core main branch afaik. I guess we could also just remove this as it is the default value I assume

.github/workflows/release.yml
christian-bromann: Do you think we can use the same token for both?

christian-bromann: @sourishkrout mind adding a repository secret? I don't have access to that section.

sourishkrout: You're admin on the repo now

christian-bromann:
Added a `COMMITTER_TOKEN` secret with the right access rights.

mxsdev (approved):
Let's keep an eye out next release 👍

PR #328, opened 2023-07-11
headless runme with `runme open`

mxsdev:
Introduces the `runme serve` command, which downloads the latest runme, installs it into `code-server`, and then runs `code-server`. Users can use the following command line flags:

 - `--args` which can be used to provide args to the `code-server` command
 - `--preview` which installs preview runme over latest stable
 - `--install` which just installs the runme into code-server, without launching code-server
 - `--config` which allows selecting a config file for `code-server`. Defaults to the default runme config dir

Current command name is not set in stone by any means.

### DIFF
--- a/internal/cmd/code_server.go
+++ b/internal/cmd/code_server.go
@@ -40,7 +40,12 @@ func codeServerCmd() *cobra.Command {
 		Short: "Launch Runme in a headless web client",
 		Args:  cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			openDir, _ := os.Getwd()
+			proj, err := getProject()
+			if err != nil {
+				return errors.Wrap(err, "failed to get project")
+			}
+
+			openDir := proj.Dir()
 
 			if len(args) > 0 {
 				openDir = args[0]


sourishkrout:
@mxsdev can we add prompting to install code-server, please? Like ask for consent to run: https://coder.com/docs/code-server/latest/install#installsh?

mxsdev:
@sourishkrout 

I thought I did add that - Is this not occurring for you?

image

sourishkrout:
> @sourishkrout
> 
> I thought I did add that - Is this not occurring for you?
> 
> image

Oh, you know what... I have MSFT's legacy code-server installed which is likely why I doesn't. Perhaps check the version or something that makes sure that's not the case?

mxsdev:
> Perhaps check the version or something that makes sure that's not the case?

I can't install MSFT code-server since it seems to have been wiped in favor of `code tunnel`. Can you give me the version string I can regex against?

sourishkrout:
Also, if I nuke `~/.config/code-server/config.yaml` and run the command the webapp prompts me for a password. I'm wondering if we overwrite `password` with `none` before launching as long as `bind-addr` is loopback (127.0.0.1).

```
bind-addr: 127.0.0.1:8080
auth: password
password: cb9027d3cdbbd090d80f7bce
cert: false
```

mxsdev:
> I'm wondering if we overwrite password with none before launching as long as bind-addr is loopback (127.0.0.1).

Maybe, it does feel a little invasive though to be messing with user configs/default configs.

Maybe a good solution here is to manage the config file ourselves, somewhere like `~/.config/runme/code-server/config.yml` and provide a `--config` flag if the user wants to provide their own?

sourishkrout:
Good point on config. Let's go with your suggestion @mxsdev.

Here's my version string - farewell MSFT `code-server`:
```
4.14.1 5c199629305a0b935b4388b7db549f77eae82b5a with Code 1.79.2
```

mxsdev:
@sourishkrout 

Added checking against version string, as well as configuration with the `--config` command, which by default creates the following config file:

```yml
bind-addr: 127.0.0.1:8080
auth: none
cert: false
```

sourishkrout:
This is shaping up quite nicely! Can we set `-an Runme` and default the folder that is being opened to the current working directory @mxsdev? Thank you!

mxsdev:
@sourishkrout 

Did the following:

Added `--title` for changing the `--app-name` (defaults to runme)

Allow specifying a folder as an argument (defaults to cwd, do you think  project dir is preferable here?). Can also specify a file, in which case it just goes to the file's root folder.

Changed command name to `open`

sourishkrout:
> Allow specifying a folder as an argument (defaults to cwd, do you think project dir is preferable here?). Can also specify a file, in which case it just goes to the file's root folder.

That's a good point. Let's use the logic that we have for project-mode. Git project over cwd.

PR #331, opened 2023-07-12
internal/runner: allocate big enough buffer when loading env

mxsdev:
Buffer can currently overflow when reading


sourishkrout (approved):
buffer overflow! LGTM 👍

PR #332, opened 2023-07-13
(release): add download-url to brew update step

christian-bromann:
Looking at other examples, e.g. https://github.com/Homebrew/homebrew-core/pull/136419/files this seems to be the way to force the action to update the git tag and sha rather than the url.


sourishkrout (approved):
There's only one way to find out 👍

PR #339, opened 2023-07-17
internal/runner: limit env store size

mxsdev:
Limits env store size to 32760 (the lowest env store size of the main platforms).

This fixes a bug where scripts that have very very large outputs would cause the notebook terminal to refuse to run a script.


sourishkrout (approved):
👍 LGTM

PR #343, opened 2023-07-19
internal/cmd/open: no longer download vscode extension manually

### LINKED ISSUE #342: Rely on OpenVSX distribution for `runme open`
We should be able to, instead of downloading the extension, just rely on the built-in marketplace distribution (https://open-vsx.org/extension/stateful/runme).

The difference should be simple: instead of giving `-install-extension` a file, it should just be the handle `stateful.runme`:
https://github.com/stateful/runme/blob/main/internal/cmd/code_server.go#L179-L181

mxsdev:
Fixes #342

### DIFF
--- a/internal/cmd/code_server.go
+++ b/internal/cmd/code_server.go
@@ -169,14 +169,7 @@ func codeServerCmd() *cobra.Command {
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


sourishkrout (approved):
👍 LGTM

PR #354, opened 2023-07-28
(bugfix): support multiple categories

christian-bromann:
Given a user has the following markdown defined:

````md
```sh { name=a category=a }
echo "Category A"
```
```sh { name=b category=a,b }
echo "Category A,B"
```
```sh { name=c category=a,b,c }
echo "Category A,B,C"
```
````

If I currently run:

```sh
 ./runme run -c b
```

it returns:

```
No tasks to execute with the category provided
```

even though I expect the following:

```
Run 2 tasks for category b? [Y/n] Y

 ►  Running task b...
Category A,B
 ►  ✓ Task b exited with code 0
 ►  Running task c...
Category A,B,C
 ►  ✓ Task c exited with code 0
```

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -90,3 +90,15 @@ To still display unknown snippets, provide `--allow-unknown` to the `list` comma
 username = admin
 password = admin
 ```
+
+## Categories
+
+```sh { name=a category=a }
+echo "Category A"
+```
+```sh { name=b category=a,b }
+echo "Category A,B"
+```
+```sh { name=c category=a,b,c }
+echo "Category A,B,C"
+```
--- a/internal/cmd/run.go
+++ b/internal/cmd/run.go
@@ -22,6 +22,7 @@ import (
 	"github.com/stateful/runme/internal/project"
 	"github.com/stateful/runme/internal/runner/client"
 	"github.com/stateful/runme/internal/tui"
+	"golang.org/x/exp/slices"
 )
 
 const (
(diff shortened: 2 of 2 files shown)


sourishkrout (changes requested):
I'm fairly certain this change won't just work with the extension.

christian-bromann:
> I'm fairly certain this change won't just work with the extension.

Can you elaborate? This patch fixes a bug within the CLI when adding `-c` as category parameter. I also tested on the VS Code extension and it was able to run the example we have in the repository.

sourishkrout:
> > I'm fairly certain this change won't just work with the extension.
> 
> Can you elaborate? This patch fixes a bug within the CLI when adding `-c` as category parameter. I also tested on the VS Code extension and it was able to run the example we have in the repository.

What happens in the Notebook Runner when you assign multiple categories? Will the UI treat it as "a,b,c" or as "a", "b", "c"?

christian-bromann:
> What happens in the Notebook Runner when you assign multiple categories? Will the UI treat it as "a,b,c" or as "a", "b", "c"?

Screenshot 2023-08-01 at 14 47 44

Apparently more as `ab`. According to the [docs](https://docs.runme.dev/configuration#run-all-cells-by-category)

````
```sh { category=category-two,category-two }
echo "Category one and two single line"
```
````

is a valid way to define a cell for two categories. I raised [an issue](https://github.com/stateful/vscode-runme/issues/733) for this some days ago as part of this investigation which I can tackle and fix as well.

sourishkrout:
Yes, please. We want both to work consistently.

christian-bromann:
> Yes, please. We want both to work consistently.

On it! I think we can move forward with this PR (merge and release) since the CLI feature is not connected with how runme is integrated in VS Code.

PR #356, opened 2023-07-31
(docs): minor structural updates to the contributing doc

christian-bromann:
Minor structural updates to better follow the development lifecycle: setup > build > testing > release

I wonder if we still need the WASM part or if we can remove it? Anyway, that would be part of a different PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -115,7 +115,7 @@ pre-commit run --files */**
 Tests are run with go's default test runner. So, for example, you can run all tests with:
 
 ```sh
-go test ./...
+make test
 ```
 
 ## Generated Files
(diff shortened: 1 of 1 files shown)


sourishkrout:
> I wonder if we still need the WASM part or if we can remove it? Anyway, that would be part of a different PR.

There's a ticket to remove it from the extension first, which is a precursor. It's ok to keep it in place for now.

sourishkrout:
Should most steps reference target's in the Makefile @christian-bromann?

sourishkrout (changes requested):
See last comment.

christian-bromann:
> Should most steps reference target's in the Makefile @christian-bromann?

Sorry, I am not entirely sure what you mean by that. If you are suggesting to transform calls like `pre-commit run --files */**` into e.g. `make precommit` or `brew bundle --no-lock` into `make bundle`, I am totally open to that. I pushed a commit to replace the `go test` call with `make test`

sourishkrout (approved):
👍 LGTM

Issue #297, opened 2023-06-12
Disable CLI prompts


christian-bromann:
With #253 we introduced a feature that would enable the CLI to prompt the user for environment variables. However there are situations where this is not desired:

- if the environment variable is already set
- in CI
- if I like the env to be empty

I would suggest to make this behavior optional and only have the CLI prompt use if the user opts in.
Wdyt?

duvanmonsa:
@christian-bromann Thank you for reporting I think it makes sense, we can set that feature off by default and add a new flag `--prompt-envs` would love to hear @mxsdev and @sourishkrout

sourishkrout:
How about skipping prompts when no tty is attached? Provide a flag to overwrite it (force prompting)?

mxsdev:
I agree with the three cases Christian provided:

> if the environment variable is already set
> in CI
> if I like the env to be empty

I already thought we had a flag - we should definitely add one.

> I would suggest to make this behavior optional and only have the CLI prompt use if the user opts in.

This breaks parity with behavior in the extension.

> How about skipping prompts when no tty is attached? Provide a flag to overwrite it (force prompting)?

Yeah, we should definitely not ask for a env prompt in CI

sourishkrout:
Let's create a solution where we have a 1.) flag to force skipping as well as 2.) a detection whether or not a TTY is attached and, if not, auto-skip, @duvanmonsa.

sourishkrout:
Done (https://github.com/stateful/runme/pull/359) and released.

PR #359, opened 2023-08-02
fix: adding flag to skip prompting for variables

duvanmonsa:
Solving issue https://github.com/stateful/runme/issues/297

Introducing new `runme run --skip-prompts` to skip prompting for variables
you can also do `runme run --skip-prompts=false` `runme run --skip-prompts=true`

### DIFF
--- a/internal/cmd/run.go
+++ b/internal/cmd/run.go
@@ -40,6 +40,7 @@ func runCmd() *cobra.Command {
 	var (
 		dryRun         bool
 		runAll         bool
+		skipPromptVars bool
 		parallel       bool
 		replaceScripts []string
 		serverAddr     string
@@ -172,9 +173,11 @@ func runCmd() *cobra.Command {
 				return err
 			}
 
-			err = promptEnvVars(cmd, sessionEnvs, runBlocks...)
-			if err != nil {
-				return err
+			if isTerminal(os.Stdout.Fd()) && !skipPromptVars {
+				err = promptEnvVars(cmd, sessionEnvs, runBlocks...)
+				if err != nil {
+					return err
+				}
 			}
 
 			if runMany {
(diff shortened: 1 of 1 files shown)


internal/cmd/run.go
```diff
@@ -287,6 +290,7 @@ func runCmd() *cobra.Command {
 	cmd.Flags().StringArrayVarP(&replaceScripts, "replace", "r", nil, "Replace instructions using sed.")
 	cmd.Flags().BoolVarP(&parallel, "parallel", "p", false, "Run tasks in parallel.")
 	cmd.Flags().BoolVarP(&runAll, "all", "a", false, "Run all commands.")
+	cmd.Flags().BoolVarP(&skipPromptVars, "skip-prompt-vars", "v", false, "Skip prompting for variables.")
```
duvanmonsa, line 293: @sourishkrout I am not sure what alias should we use for this right now `v`

duvanmonsa:
@sourishkrout Also wondering if this is something we need to do on TUI as well?

internal/cmd/run.go
sourishkrout: `v` is very common for `verbose` so let's not use it, please. How about we rename it to `--skip-prompts` with `s` as alias the `vars` part is redundant.

internal/cmd/run.go
```diff
@@ -172,9 +173,11 @@ func runCmd() *cobra.Command {
 				return err
 			}
 
-			err = promptEnvVars(cmd, sessionEnvs, runBlocks...)
-			if err != nil {
-				return err
+			if isTerminal(os.Stdout.Fd()) && !skipPromptVars {
```
sourishkrout, line 176: Wondering if `skipPromptVars` should be set to true if it's not a terminal and the flag's not changed explicitly. That way one could always enforce behavior with `--skip-prompts=true` or `--skip-prompts=false`.

sourishkrout (changes requested):
See my two comments please @duvanmonsa.

sourishkrout:
> @sourishkrout Also wondering if this is something we need to do on TUI as well?

That's separate. The main desire for this flag is to be used in non-interactive use cases. The TUI is inherently interactive.

internal/cmd/run.go
duvanmonsa: we are removing the alias only `--skip-prompts` will be available.

internal/cmd/run.go
sourishkrout: What about ☝️ @duvanmonsa ?

duvanmonsa: I changed the default value to `true` @sourishkrout

sourishkrout:
```
runme run --filename README.md --all --skip-prompts
                         
Run all 24 tasks? [Y/n] Y
```

We should probably skip this prompt too @duvanmonsa, no?

duvanmonsa:
> 

@sourishkrout my position is kind of divided right now, I get why you think we should skip it based on the flag name, but also I feel we should actually ask for that confirmation it is up to you.

sourishkrout:
let's skip it too @duvanmonsa. I feel like `--skip-prompts` just do what it says especially in non-TTY places like C/I.

duvanmonsa:
Ok changes to skip that confirmation have been done as well!

sourishkrout (approved):
👍 LGTM

PR #365, opened 2023-08-10
Fix dangling tempfiles

sourishkrout:
In the event of an error, the cleanup which used to be irrelevant (process quits) up until file resources were introduced to unlock shebang support. The example below would keep dangling temp files on the filesystem after exiting with an error due to a "corrupt" `ts-node` installation.

```
$ runme run --filename README.md function-unnest
No preset version installed for command ts-node
Please install a version by running one of the following:

asdf install nodejs 18.13.0

or add one of the following versions in your config file at /Users/sourishkrout/.tool-versions
nodejs 19.9.0
exit code: 126
```

### DIFF
--- a/internal/runner/command.go
+++ b/internal/runner/command.go
@@ -14,8 +14,8 @@ import (
 	"syscall"
 
 	"github.com/creack/pty"
-	"github.com/google/uuid"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


duvanmonsa (approved):
LGTM

PR #372, opened 2023-09-06
Feat/get blocks

duvanmonsa:
Add getBlocks api to the binary


sourishkrout:
Closing since we now have https://github.com/stateful/runme/pull/438 and subsequent work on the project service.

PR #379, opened 2023-09-21
Add json to list cmd

### LINKED ISSUE #380: Allow json output for the list command
I am thinking of ways of using the list command in the [neovim plugin](https://github.com/catdevman/runme.nvim) I create for runme.  I would like to add a `Runme all` command that brings up a list of all the files from the current project that could have runme ran on them. 

I went ahead and threw a [PR](https://github.com/stateful/runme/pull/379) together to show what I was thinking.

 I can get what I am looking for by using list command and making a set with the file names but parsing that from the table format would be difficult.  So far it looks like this is the only command that has table output so I just wrapped that in a isJSON flag that defaults to false.  This will keep the current output the same but if `--json` is added to the command it will output json data that is the same as the table besides the named field. I made that a bool instead of converting it to YES/NO.

catdevman:
This should close #380

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,6 @@ require (
 	github.com/charmbracelet/bubbletea v0.24.2
 	github.com/charmbracelet/lipgloss v0.8.0
 	github.com/cli/cli/v2 v2.34.0
-	github.com/cli/go-gh/v2 v2.3.0
 	github.com/containerd/console v1.0.4-0.20230313162750-1ae8d489ac81
 	github.com/creack/pty v1.1.18
 	github.com/fatih/color v1.15.0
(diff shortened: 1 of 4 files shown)


catdevman:
Changes:
- changed isJSON to [formatJSON](https://github.com/stateful/runme/pull/379/files#diff-744fdc7b8f23410eaaa16b4944f27aff8ce740f150d6b4ed81cdfd18d3fb6226R28)
- moved [row](https://github.com/stateful/runme/pull/379/files#diff-744fdc7b8f23410eaaa16b4944f27aff8ce740f150d6b4ed81cdfd18d3fb6226R19) struct top level of the file so it get be used throughout
- moved displaying [table](https://github.com/stateful/runme/pull/379/files#diff-744fdc7b8f23410eaaa16b4944f27aff8ce740f150d6b4ed81cdfd18d3fb6226R91) and [json](https://github.com/stateful/runme/pull/379/files#diff-744fdc7b8f23410eaaa16b4944f27aff8ce740f150d6b4ed81cdfd18d3fb6226R119) into functions that take iostreams.IOStreams and []row
- added txtar/testscript [test](https://github.com/stateful/runme/pull/379/files#diff-a29265db1d437868eb16800af59c24b8501cc8df653c8e531d8725d3c58867efR5) for `runme ls --json`

sourishkrout (approved):
LGTM 👍

sourishkrout:
Tidied the modules and ran gofumpt (`make fmt`) which C/I will enforce. Merging on green build @catdevman.

PR #382, opened 2023-09-25
Refactor project package to allow linkage

sourishkrout:
No change to functionality whatsover.


pastuxso (approved):
Looks good to me. :rocket:

internal/cmd/tui.go
```diff
@@ -37,7 +37,8 @@ func tuiCmd() *cobra.Command {
 				return err
 			}
 
-			blocks, err := loadTasks(proj, cmd.OutOrStdout(), cmd.InOrStdin(), false)
+			loader := getLoader(cmd)
```
pastuxso, line 40: It would be fine if `project.NewLoader` would have a return with `error` as well.

sourishkrout: @pastuxso can you push a commit on top with the changes you have in mind?

sourishkrout:
Thanks @pastuxso!

PR #383, opened 2023-09-25
Use min() and max() from Go 1.21

adambabik:
As runme supports Go 1.21+, it can benefit from newly introduced `min()` and `max()` functions.

I needed to bump gosec to recognize the new functions and it triggered a few false positives, which I commented.


sourishkrout:
Nice, thanks @adambabik! Much appreciated.

sourishkrout (approved):
👍 LGTM

Issue #408, opened 2023-10-24
Persistent id for every cell


sourishkrout:
Runme currently tracks cells (`runme.dev/uuid`; [see extension](https://github.com/stateful/vscode-runme/blob/main/src/schema.ts#L50)) with a lifecycle that spans the lifetime of a notebook document (see `GrpcSerializer` in extension) in VS Code.

Let’s add a persistent serialized `cuid` or `xid` before removing the reliance on `runme.dev/uuid` on the extension-side. UUIDs are less desirable due to their long-ish nature.

The steps are roughly (in multiple PRs to allow for a transition phase):

Closed:

- [x] stateful/runme#411 
- [x] stateful/runme#412 
- [x] https://github.com/stateful/runme/pull/416
- [x] https://github.com/stateful/runme/pull/421
- [x] https://github.com/stateful/runme/pull/423
- [x] https://github.com/stateful/vscode-runme/pull/950
- [x] https://github.com/stateful/runme/pull/428

Pending:

- [ ] https://github.com/stateful/vscode-runme/pull/938
- [ ] https://github.com/stateful/runme/issues/429
- [ ] stateful/vscode-runme#976

joseeantonior:
After a discussion, it was decided that we would use ULIDs (https://github.com/ulid/spec) due to their time-ordered and unique nature. It also greatly helps that they are timecoded.

There was also the discussion about adding IDs to notebooks. This particular change allows people to restructure directories without losing their notebook's history.

Finally, we will also save the Runme version number used to save the document.

Hence, there is work being done to generate an ULID to identify the notebook, and then ULIDs to identify the cells.

PR #411, opened 2023-10-31
switches from xid to ulid

pastuxso:
Related: https://github.com/stateful/runme/issues/408


sourishkrout (changes requested):
Works as expected. Without wanting to be nit-picky, we've tried hard to avoid catch all "utils" - probably haven't mentioned that to you before @pastuxso.

Perhaps we can just rename the package `ulid` or something of that sort. Let's see how much longer we can get away without utils.

sourishkrout (approved):
👌 @pastuxso let's merge!

PR #412, opened 2023-11-01
Adds unique identifiers at both the notebook and cell levels

pastuxso:
Related: https://github.com/stateful/runme/issues/408

https://github.com/stateful/runme/assets/43882/b2e53035-446c-4368-a902-05f1b414d83a


internal/document/editor/editor.go
```diff
@@ -37,10 +37,10 @@ func Deserialize(data []byte) (*Notebook, error) {
 		PrefixAttributeName(InternalAttributePrefix, constants.FinalLineBreaksKey): fmt.Sprint(finalLinesBreaks),
 	}
 
-	// If Front Matter exists, store it in Notebook's metadata.
```
pastuxso, line 40: I will refactor this part in the upcoming PR.

internal/document/frontmatter.go
```diff
@@ -1,19 +1,29 @@
...
 	"strings"
 
 	"github.com/pelletier/go-toml/v2"
 	parserv1 "github.com/stateful/runme/internal/gen/proto/go/runme/parser/v1"
+	"github.com/stateful/runme/internal/idgen"
+	"github.com/stateful/runme/internal/version"
 	"gopkg.in/yaml.v3"
 )
 
+type RunmeMetaData struct {
```
adambabik, line 17: It's a bit odd that `Frontmatter` has only `yaml` tags, but this struct has more. Is there a reason for that?

internal/document/frontmatter.go
```diff
@@ -67,10 +87,104 @@ func ParseFrontmatter(raw string) (f Frontmatter, info FrontmatterParseInfo) {
 	return
 }
 
+func ToJSONStr(dest map[string]interface{}, bytes []byte) string {
```
adambabik, line 90: General comments to this set of functions:
* They are not used outside of `internal/document`; they should be unexported.
* It seems that it's a quite serious bug if unmarshaling fails; I think we should report that and handle, otherwise it will be extremely hard to find the bug. It seems that this function is only called in `Deserialize()` which returns an error.

internal/document/frontmatter.go
```diff
@@ -67,10 +87,104 @@ func ParseFrontmatter(raw string) (f Frontmatter, info FrontmatterParseInfo) {
...
+			Version: fmtr.Runme.Version,
+		},
 		Shell:       fmtr.Shell,
 		Cwd:         fmtr.Cwd,
 		SkipPrompts: fmtr.SkipPrompts,
 	}
 }
+
+// InjectYamlTestID injects a test id into a yaml document
+func InjectYamlTestID(s string) string {
```
adambabik, line 180: I'm not sure if this is a correct description. It seems like `InjectFrontmatter(documentString, runmeID, runmeVersion string) string` might explain the purpose better. It may also make more sense to place it in `internal/document/editor`.

internal/document/frontmatter.go
sourishkrout: Likely because `yaml` is the most common frontmatter encoding we encounter. I've been meaning to test json and toml.

sourishkrout: In fact, @joseeantonior, could you please track a ticket for us to test `json` & `toml` frontmatter encoding, just so we don't forget? Thank you!

sourishkrout:
## Functional aspects @pastuxso (cc @joseeantonior):
1. Should we update the version number once it changes? To always reflect the most recent version. My might want to consider shortening it for those purposes to, e.g. only include major.minor but not patch.
2. It seems like when the runme frontmatter metadata prop is incomplete but present it, won't complete it. See examples below.

### Won't add nested props
```yaml
---
runme:
shell: bash
---
```

turns it into:
---
runme: null
shell: bash
---


### Won't add version number
```yaml
---
runme:
  id: 01HEBAF6W797022GRA5QV0VZS6
shell: bash
---
```

internal/document/frontmatter.go
joseeantonior: https://github.com/stateful/runme/issues/415

sourishkrout: > There are tests that use TOML and JSON in their metadata, so I had to make sure they worked.

I specifically mean integration testing with the extension. Wasn't clear; will add that bit to the ticket.

joseeantonior:
> Should we update the version number once it changes? To always reflect the most recent version. My might want to consider shortening it for those purposes to, e.g. only include major.minor but not patch (to avoid diff too frequently).

I feel like major.minor should be published, while we internally know patch versions and/or use them as milestones.

Makefile
```diff
@@ -63,6 +67,17 @@ proto/generate:
 proto/clean:
 	rm -rf internal/gen/proto
 
+.PHONY: proto/dev
+proto/dev: build proto/clean proto/generate
```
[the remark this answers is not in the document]
sourishkrout: Formally, you're correct on the dependency graph. However, very few people contributing to the extension also contribute to CLI/kernel. While inverted, it makes sense from the perspective of the people. So, let's keep it as is, please.

sourishkrout:
> > Should we update the version number once it changes? To always reflect the most recent version. My might want to consider shortening it for those purposes to, e.g. only include major.minor but not patch (to avoid diff too frequently).
> 
> I feel like major.minor should be published, while we internally know patch versions and/or use them as milestones.

👍, let's do it (cc @pastuxso)

pastuxso:
> ## Functional aspects @pastuxso (cc @joseeantonior):
> 1. Should we update the version number once it changes? To always reflect the most recent version. My might want to consider shortening it for those purposes to, e.g. only include major.minor but not patch (to avoid diff too frequently).
> 2. It seems like when the runme frontmatter metadata prop is incomplete but present it, won't complete it. See examples below.
> 
> ### Won't add nested props
> ```yaml
> ---
> runme:
> shell: bash
> ---
> ```
> 
> turns it into:
> 
> ```yaml
> ---
> runme: null
> shell: bash
> ---
> ```
> 
> ### Won't add version number
> ```yaml
> ---
> runme:
>   id: 01HEBAF6W797022GRA5QV0VZS6
> shell: bash
> ---
> ```

@sourishkrout Solved: https://github.com/stateful/runme/pull/412/commits/61f577c6b93aaed488f9a330beb00ac3f9334788

Makefile
```diff
@@ -63,6 +67,17 @@ proto/generate:
...
+	cp -vrf internal/gen/proto/ts/runme $(RUNME_EXT_BASE)/node_modules/@buf/stateful_runme.community_timostamm-protobuf-ts
```
adambabik, line 72: Should `RUNME_EXT_BASE` have a default value? I guess typically it will be in the same parent dir, so maybe:
```sh
ifeq ($(RUNME_EXT_BASE),)
RUNME_EXT_BASE := "../vscode-runme"
endif
```

Or whatever makes sense.

examples/grpc-client/main.go
```diff
@@ -64,17 +65,20 @@ func run() error {
 	if *parseDocument {
 		client := parserv1.NewParserServiceClient(conn)
 
+		filename := "examples/grpc-client/hello.md"
```
adambabik, line 68: Hint: it's ok, but it can be improved by using [`//go:embed`](https://pkg.go.dev/embed) so that it won't depend on from where `main.go` is run.

internal/document/frontmatter.go
adambabik: Nit: Ok, makes sense. Still, `protobuf` likely should be removed to avoid confusion; it is used only in generated stubs.

internal/version/version.go
```diff
@@ -1,7 +1,23 @@
...
+	"github.com/Masterminds/semver/v3"
+)
+
 var (
 	BuildDate    = "unknown"
 	BuildVersion = "0.0.0"
 	Commit       = "unknown"
 )
+
+// BaseVersion returns the base version of the application.
```
adambabik, line 15: Suggestion: As a new term "the base version of the application" is introduced, it would be good to add a comment explaining it:

```suggestion
// BaseVersion returns the base version of the application.
//
// The base version is defined using the "MAJOR.MINOR" format. Patch and any metadata
// are removed.
```

internal/version/version_test.go
```diff
@@ -0,0 +1,47 @@
+package version
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetMinorVersion(t *testing.T) {
```
adambabik, line 9:
```suggestion
func TestBaseVersion(t *testing.T) {
```

internal/version/version_test.go
```diff
@@ -0,0 +1,47 @@
...
+	// Test cases
+	tests := []struct {
```
adambabik, line 11: Important: This test table misses the case where the semver is invalid.

sourishkrout (approved):
👍 LGTM.

PR #416, opened 2023-11-06
Expose interface for the client to control identity

sourishkrout:
1. Minor rename; it seems like go packages favor entity-/data-centric names over operations-centric ones.
2. Introducing a way to turn doc/cell identity-tracking off selectively. I will handle the client side (extension).

Could you please fill in the gap on the serializer side, @pastuxso? Hope this won't take up too much time (cc @joseeantonior).


internal/api/runme/parser/v1/parser.proto
```diff
@@ -51,6 +58,7 @@ message DeserializeResponse {
 
 message SerializeRequest {
   Notebook notebook = 1;
+  RunmeIdentity identity = 2;
```
sourishkrout, line 61: @pastuxso I'm pondering if we should have an "options" field here that "identity" will be part of. Make the distinction between message/data and config/options more obvious; what do you think?

internal/document/editor/editorservice/service.go
```diff
@@ -70,8 +71,13 @@ func (s *parserServiceServer) Serialize(_ context.Context, req *parserv1.Seriali
 
 	cells := make([]*editor.Cell, 0, len(req.Notebook.Cells))
 	for _, cell := range req.Notebook.Cells {
-		if _, ok := cell.Metadata["id"]; !ok && cell.Kind == parserv1.CellKind_CELL_KIND_CODE {
-			cell.Metadata["id"] = idgen.GenerateID()
+		if req.Identity.Number() == parserv1.RunmeIdentity_RUNME_IDENTITY_ALL.Number() ||
```
sourishkrout, line 74: I'm fairly certain you can compare the props without calling `Number()` on it, no?

sourishkrout (commented):
See remark on enums.

internal/api/runme/parser/v1/parser.proto
pastuxso: I like the idea. Are you thinking of something like a **`Params`** structure?"

sourishkrout: > @sourishkrout I like the idea, Are you thinking of something like a **`Params`** structure?"

As discussed out-of-band, let's go with `Options`.

pastuxso: Done :white_check_mark:

pastuxso (approved):

Issue #418, opened 2023-11-07
PHP shebang not working


sourishkrout:
It appears the Runme kernel applies shell/bash normalization to the PHP code, which strips the leading `$` for variables.

```php
<?php
// PHP Script Example: Greeting with Date and Time

// Define a variable for the greeting
$greeting = "Hello, World!";

// Get the current date and time
$currentDateTime = date('Y-m-d H:i:s');

// Concatenate the greeting with the current date and time
$fullGreeting = $greeting . " It's now " . $currentDateTime;

// Output the full greeting
echo $fullGreeting;
?>
```

Reported by @edeediong.

sourishkrout:
Will fix this as part of releasing v2.

sourishkrout:
Fixed. Will go out with the next release.

PR #419, opened 2023-11-07
Attempt php auto-detection

sourishkrout:
This will look for a local `php` to present in the system's `$PATH`.

Enables https://github.com/stateful/runme/issues/418.


edeediong (approved):
Small change. Looks good!

PR #422, opened 2023-11-07
Not pretty but practical fix to have node shebang hit

sourishkrout:
When a node version manager is involved (asdf in my case), this test keeps failing due to the way asdf "interjects" calling the node runtime. This fixes it locally and should be harmless in CI/CD.

Btw, there is a way to set up the Environ in the test harness. However, that will apply it to all test cases and make other things fail for unrelated reasons.

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -1,3 +1,4 @@
 brew "bufbuild/buf/buf"
 brew "grpcurl"
 brew "go"
+brew "node"
--- a/testdata/script/basic.txtar
+++ b/testdata/script/basic.txtar
@@ -10,6 +10,7 @@ exec runme ls --allow-unnamed=true
 cmp stdout golden-list-allow-unnamed.txt
 ! stderr .
 
+env PATH=/opt/homebrew/bin:$PATH
 ! exec runme ls --filename nonexistent.md
 stderr 'failed to open markdown file .*/nonexistent.md: no such file or directory'
 ! stdout .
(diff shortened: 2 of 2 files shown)


sourishkrout:
@degrammer I believe you had the same issue where tests fail locally but pass in the pipeline. See if this fixes it, please.

sourishkrout:
What the heck.

sourishkrout:
> > @degrammer I believe you had the same issue where tests fail locally but pass in the pipeline. See if this fixes it, please.
> 
> Still failing for me locally: https://app.runme.dev/cell/clop0qsil0025s601b8nvw46j

Can you please confirm that you're on Golang 1.21.x by running `go version`?

sourishkrout:
> > > > @degrammer I believe you had the same issue where tests fail locally but pass in the pipeline. See if this fixes it, please.
> > > 
> > > 
> > > Still failing for me locally: https://app.runme.dev/cell/clop0qsil0025s601b8nvw46j
> > 
> > 
> > Can you please confirm that you're on Golang 1.21.x by running `go version`?
> 
> @sourishkrout : go version go1.21.3 darwin/arm64

K, the version seems right. However, the error stack trace is all over the place. Unclear what's causing that for you, but it's not the testscript/textvar stuff that's causing trouble when node's involved (my issue). We can diagnose separately, however, don't think that's important right now.

pastuxso (approved):

PR #423, opened 2023-11-08
Fix delete existing IDs on unspecified IDT

pastuxso:


internal/version/version.go
```diff
@@ -19,5 +19,5 @@ func BaseVersion() string {
 		return BuildDate
 	}
 
-	return fmt.Sprintf("%d.%d", v.Major(), v.Minor())
+	return fmt.Sprintf("v%d.%d", v.Major(), v.Minor())
```
pastuxso, line 22: Prefix the letter 'v' to prevent YAML parsing from converting a '1.0' string into a number, i.e. 0.0 to 0, 1.0 to 1, etc.

internal/document/editor/editor_test.go
```diff
@@ -33,48 +32,11 @@ func TestEditor(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(
 		t,
-		document.InjectFrontmatter(string(testDataNestedFlattened)),
+		string(testDataNestedFlattened),
 		string(result),
 	)
 }
 
-func TestEditorEmptyRunme(t *testing.T) {
```
pastuxso, line 41: Invalid tests related to IDT are now located in the Serialize GRPC Land.(Serialize)

internal/document/frontmatter.go
```diff
@@ -75,8 +75,6 @@ func toJSONStr(f *Frontmatter, source []byte, requireIdentity bool) (string, err
 	if requireIdentity {
 		f.ensureID()
 		m["runme"] = f.Runme
-	} else {
```
pastuxso, line 78: Here we are avoiding removing existent frontmatter metadata, but what happens if the **`runme`** metadata is wrong or invalid?

sourishkrout: Like an invalid ID (ulid format) or version number (not following the vx.x convention) @pastuxso? You have some examples of invalidity?

pastuxso: @sourishkrout i.e. any hardcoded string in the metadata.

sourishkrout: I think we should be super conservative with `IDs`. Unless we really can't work with it, we shouldn't reset it. The version we can just overwrite since it does not constitute "identity". Does that help @pastuxso?

pastuxso: @sourishkrout Yes, that helps, I think we can tackle it in an upcoming task.

sourishkrout (changes requested):
@pastuxso I'm getting a panic because `metadata` is `nil` here now: https://github.com/stateful/runme/blob/fix/unspecified-idt/internal/document/editor/editorservice/service.go#L73.

Testing with https://github.com/stateful/vscode-runme/blob/runme-identity-persist/examples/shebang.md.

internal/document/editor/editor.go
```diff
@@ -37,10 +37,29 @@ func Deserialize(data []byte) (*Notebook, error) {
...
-	f, info := document.ParseFrontmatter(string(sections.FrontMatter))
+	// If Front Matter exists, store it in Notebook's metadata.
+	if len(sections.FrontMatter) > 0 {
+		notebook.Metadata[PrefixAttributeName(InternalAttributePrefix, FrontmatterKey)] = string(sections.FrontMatter)
+	}
+
+	return notebook, nil
+}
+
+func DeserializeV2(data []byte, requireIdentity bool) (*Notebook, error) {
```
sourishkrout, line 48: @pastuxso can we consolidate `Deserialize` & `DeserializeV2`, the `fmt` won't have flags to toggle doc/cell identity options (default `false` for now). We can spin off a ticket to add those later.

sourishkrout (approved):
Let's consolidate the two versions of `Deserialize` before merging. Integration testing passed for me ✅.

sourishkrout (approved):
LGTM 👍! Let's merge.

PR #427, opened 2023-11-13
Fix brew buf installation

dantehemerson:
Had some issues while installing `bufbuild/buf/buf`
Screenshot 2023-11-12 at 8 12 10 PM

However installing `buf` worked:
https://formulae.brew.sh/formula/buf#default


Brewfile
```diff
@@ -1,4 +1,4 @@
-brew "bufbuild/buf/buf"
```
adambabik, line 1: https://buf.build/docs/installation still advertises to use `bufbuild/buf/buf`.

What is likely missing is `tap "bufbuild/buf"` in this file at the top.

sourishkrout (approved):
👍 LGTM after @adambabik's comments were addressed. Thank you so much, @dantehemerson!

Issue #426, opened 2023-11-10
Add document/cell identity options to fmt sub-command


sourishkrout:
Essentially allow options to be passed by flags from `runme fmt` according to how defaults shake out. This will require moving the cell-identity logic out of the GRPC service.

https://github.com/stateful/runme/blob/cc380d9ab27092165f1845b08d179afe8f52417c/pkg/project/formatter.go#L31

This

PR #428, opened 2023-11-13
Lifecycle identity refactor

sourishkrout:
1. Treat cell IDs similar to the way we handle generated names.
2. Move biz logic into document.
3. Separate ulid spec and identity impl.
4. Brings lifecycle identity to `runme fmt` (solves https://github.com/stateful/runme/issues/426)

Missing (not merge critical):
- Tests for identityResolver itself (covered mostly by package integration unit tests)
- Bring back `v0.0` overwrite protection
- Consider todos on ID semantics, specifically whether or not to retain invalid ULIDs in Cell IDs


pastuxso (approved):
LGTM :rocket:

PR #431, opened 2023-11-15
 internal/document: include natively support for frontmatter

adambabik:
This change includes native support for frontmatter in `document.Document`. It consolidates operations on frontmatter and encapsulates the details behind `Document`'s API.

It's a prep work for the refactoring of how the idea of projects is implemented.

### Testing

- [ ] Verify if there is no breaking change, or the breaking change is accepted, in terms of parsing and serializing frontmatter.

### Todo

- [ ] Most of the tests from `editorservice/service_test.go` should be moved to `editor_test.go`. From the `editorservice` perspective, they only test proper passing of `parserv1.DeserializeRequestOptions`, but their content validates `editor.Deserialize()`'s logic.


internal/document/editor/editorservice/service_test.go
```diff
@@ -95,7 +95,7 @@ func Test_IdentityUnspecified(t *testing.T) {
 			assert.True(t, ok)
 			assert.Len(t, dResp.Notebook.Metadata, 2)
 			assert.Contains(t, rawFrontmatter, "prop: value\n")
-			assert.Contains(t, rawFrontmatter, "id: 123\n")
+			assert.Contains(t, rawFrontmatter, "id: \"123\"\n")
```
adambabik, line 98: Important: It's interesting why it changed. It should be debugged.

sourishkrout:
Thanks @adambabik. @pastuxso if you want to take a first review pass. I'll go over it as well.

internal/ulid/generator.go
```diff
@@ -1,4 +1,4 @@
-package identity
+package ulid
```
sourishkrout, line 1: 🤦 ugh, that's exactly what I wanted to do. good catch!

sourishkrout:
Looked over the code which looks 👍 . However, I want to do a bit more integration testing (next 24h) before approving for merge.

internal/document/editor/editorservice/service_test.go
sourishkrout: This is odd indeed, it appears that `abc123` won't require the quotes whereas a number `123` does. Wondering if the YAML type serialization quotes all numbers when casting to string?

pastuxso:
> It's a prep work for the refactoring of how the idea of projects is implemented.

Could you share more something about this idea? looks like related to its architecture, it will be very helpful to me :wink:.

pastuxso (approved):
LGTM :rocket:

sourishkrout:
> > It's a prep work for the refactoring of how the idea of projects is implemented.
> 
> Could you share more something about this idea? looks like related to its architecture, it will be very helpful to me 😉.

It's a pre-cursor to the pre-cursor (some recursion) to be able to have Project-mode (read blog https://runme.dev/blog/tasks-over-files-in-runme-cli) in other places that is not just the CLI. The plan is to move it into the gRPC layer so that the CLI and different parts (explorer, tasks) of the extension can use a unified API. Turns out there's a lot of coupling to undo for that to be viable, @pastuxso.

adambabik:
> > It's a prep work for the refactoring of how the idea of projects is implemented.
> 
> Could you share more something about this idea? looks like related to its architecture, it will be very helpful to me 😉.

@sourishkrout explained it well in https://github.com/stateful/runme/pull/431#issuecomment-1819358423. Overall, this change is not necessarily fixing any big coupling between the projects implementation and the rest of the program, just a small one, but more is coming. 

The steps are as follows:
1. Get https://github.com/stateful/runme/pull/433 merged which brings `internal/project` into the CLI; unused for now.
2. Implement the projects gRPC service based on `internal/project`.
3. Validate the service in the extension.
4. Replace `pkg/project` usage in the CLI with `internal/project`.
5. Expose only necessary details of `internal/project` in `pkg/project`, so that the CLI, service, and other programs depending on runme use the same logic.

internal/document/editor/editorservice/service_test.go
adambabik: Yes, if a number is treated as a string, it should be quoted. 

However, in this case, I changed the logic and used formatted `RunmeMetadata` always. If we don't require identity, the raw value should be used. It's fixed in https://github.com/stateful/runme/pull/431/commits/e58c2060567aacf38b9d2eb6d252d84dd60ed02e.

PR #432, opened 2023-11-18
internal/gitrepo: extract git functions from pkg/project

adambabik:
The goal is to move logic back from "pkg/project" to the "internal/*" packages and expose in "pkg/project" back only necessary public functions.


adambabik:
@sourishkrout please let me know if any of the moved public functions are used by other packages and should be still exposed.

sourishkrout (approved):
This looks great. Tested it and works perfectly 👍 .

PR #433, opened 2023-11-18
Add internal/project

adambabik:
This PR adds a refactored implementation of `pkg/project` in `internal/project`. For now, it's unused. Further adjustments will be added, when the usage of `internal/project` is incorporated in the rest of the program.


sourishkrout:
Great @adambabik! Should we review it now or once it's integrated?

adambabik:
@sourishkrout let's do it now so that the PR is not massive. I will create a projects gRPC service based on it first, and when it's accepted, I will integrate "internal/project" into other parts of the program.

adambabik:
@sourishkrout After giving it a bit more thought, I think it might be hard to review from the functional point of view. Let's do it when the gRPC service is implemented.

sourishkrout:
> @sourishkrout After giving it a bit more thought, I think it might be hard to review from the functional point of view. Let's do it when the gRPC service is implemented.

Aye aye, sir!

adambabik:
Replaced with https://github.com/stateful/runme/pull/438

PR #435, opened 2023-11-20
Adds version overwrite protection

pastuxso:


sourishkrout:
@pastuxso can you fill in a description please

.vscode/settings.json
```diff
@@ -1,5 +1,9 @@
 {
     "go.testTimeout": "10s",
+    "go.buildFlags": [
+        // Hardcoded the runme version because VSCode couldn't resolve it.
+        "-ldflags=-X 'github.com/stateful/runme/internal/version.BuildVersion=2.0.0'",
```
sourishkrout, line 5: @pastuxso wondering if we should set it something more fixture-like such as `1.2.3`. I think it'll stand out more since we're now officially on 2.x.x.

sourishkrout (changes requested):
See comment about using a base v1 as fixture since we will likely have more major releases going forward and it makes it more clear that it's a fixture.

sourishkrout (approved):
@pastuxso can you make it `v99.x.x` please. It's not inconceivable that we get to `v9` in a reasonable amount of time. Otherwise, ready to merge 👍 .

Issue #305, opened 2023-06-15
Symlink-ing generated TS files won't work


sourishkrout:
When integrating with gRPC as part of development, a simple symlink won't work due to a superfluous `.ts` per proto (perhaps a CJS vs ESM thing). This is not a problem with the NPM-published artifact.

Below `runner_pb.ts` and `parser_pb.ts` will overwrite definitions in the respective `*.d.ts` files, breaking the TS compilation.

Let's figure out if this is fixable or add some docs to get ahead of complications.

```
.
├── parser
│   └── v1
│       ├── parser_pb.client.d.ts
│       ├── parser_pb.client.js
│       ├── parser_pb.d.ts
│       ├── parser_pb.js
│       └── parser_pb.ts
└── runner
    └── v1
        ├── runner_pb.client.d.ts
        ├── runner_pb.client.js
        ├── runner_pb.d.ts
        ├── runner_pb.js
        └── runner_pb.ts

5 directories, 10 files
```

PR #439, opened 2023-11-27
Only use timostamm-protobuf-ts

sourishkrout:
Removes odd side-effect of directly relying on the generate files in TS/JS. We are using `timostamm-protobuf-ts` which is mostly important on the client-side to get full Typescript benefits.

(cc @pastuxso)

Fixes https://github.com/stateful/runme/issues/305.


sourishkrout:
I'm also wondering if we should remove buf connect, @adambabik, since we aren't actually using it to connect to the kernel APIs.

adambabik:
@sourishkrout yeah, it was an experiment. If we use gRPC exclusively, we can totally get rid of connect. There is a plugin to remove and some code from `server.go`.

sourishkrout:
Can I get an approval, please, @pastuxso? Thank you.

pastuxso (approved):
LGTM

PR #440, opened 2023-11-27
Includes notebook lifecycle id in protos

pastuxso:
It's required in the extension in order to be included as part of the mutation to persis a cell execution.

PR: https://github.com/stateful/vscode-runme/pull/979


internal/document/editor/editorservice/service.go
```diff
@@ -58,6 +58,17 @@ func (s *parserServiceServer) Deserialize(_ context.Context, req *parserv1.Deser
 			Cwd:         notebook.Frontmatter.Cwd,
 			SkipPrompts: notebook.Frontmatter.SkipPrompts,
 		}
+
+		var runme *parserv1.Runme
+
+		if notebook.Frontmatter.Runme.ID != "" && notebook.Frontmatter.Runme.Version != "" {
```
sourishkrout, line 64: @pastuxso I think we want to set ID and Version separately since, from a validity standpoint, they can show up either/or or both.

Otherwise :+1:.

internal/api/runme/parser/v1/parser.proto
```diff
@@ -4,6 +4,11 @@ package runme.parser.v1;
 
 option go_package = "github.com/stateful/runme/internal/gen/proto/go/runme/parser/v1;parserv1";
 
+message Runme {
```
sourishkrout, line 7: @pastuxso how about FrontmatterRunme? Move down above Frontmatter?

pastuxso: I like it, let me change it.

sourishkrout (approved):
:+1: LGMT. Please note, it'll require myself to publish the latest protos to our NPM registry.

Issue #413, opened 2023-11-02
Casting int to uint spells danger


sourishkrout:
In an error scenario, `exitCode` can return `-1`, which will incorrectly be turned into 4294967295.

https://github.com/stateful/runme/blob/a1052d7a98398918e7a9f1e53ce040560e36a575/internal/runner/service.go#L470

What's the fix? As far as I remember, negative exit codes aren't actually valid.

PS: Still need to investigate the underlying issue causing the exit code to be returned as -1 (stack trace below).

```
signal: hangup
github.com/stateful/runme/internal/runner.(*command).ProcessWait
	/Users/sourishkrout/Projects/stateful/oss/runme/internal/runner/command.go:498
github.com/stateful/runme/internal/runner.(*runnerService).Execute
	/Users/sourishkrout/Projects/stateful/oss/runme/internal/runner/service.go:429
github.com/stateful/runme/internal/gen/proto/go/runme/runner/v1._RunnerService_Execute_Handler
	/Users/sourishkrout/Projects/stateful/oss/runme/internal/gen/proto/go/runme/runner/v1/runner_grpc.pb.go:246
google.golang.org/grpc.(*Server).processStreamingRPC
	/Users/sourishkrout/go/pkg/mod/google.golang.org/grpc@v1.58.3/server.go:1668
google.golang.org/grpc.(*Server).handleStream
	/Users/sourishkrout/go/pkg/mod/google.golang.org/grpc@v1.58.3/server.go:1755
google.golang.org/grpc.(*Server).serveStreams.func1.1
	/Users/sourishkrout/go/pkg/mod/google.golang.org/grpc@v1.58.3/server.go:986
runtime.goexit
	/opt/homebrew/Cellar/go/1.21.3/libexec/src/runtime/asm_arm64.s:1197
```

sourishkrout:
Perhaps you remember this @adambabik?

adambabik:
I took a quick look:
- runner's code uses `exec.Cmd.Wait()` from stdlib which can return `exec.ExitError` but can also return a generic error.
- `exitCodeFromErr()` which translates the error to exit code only recognises `exec.ExitError`; otherwise it returns -1.
- Later, [the logic](https://github.com/stateful/rdme/blob/25f12ee8b5058483c43f57f05473ba04a6d810a7/internal/runner/service.go#L447) merely logs if the exit code is equal to -1 and does not handle it in any way.

[...]

PR #443, opened 2023-11-29
First stab at serializing output

sourishkrout:
Add capabilities to serialize (no deserialization yet) to the ParserService. Whether or not outputs are being considered can be toggled in the `Options`.

The extension integration is partially done but it's time to merge this to avoid long-running branches with protobuf changes. It's a pain to keep rebasing them. Relevant PR on the extension: https://github.com/stateful/vscode-runme/pull/988.

Also includes unrelated fix for https://github.com/stateful/runme/issues/413.

PR #445, opened 2023-12-04
Avoid unit overflow when signal interrupt processes

sourishkrout:
Fixes the downstream overflow issue https://github.com/stateful/runme/issues/413.

However, the culprit of `-1` as an error code is likely upstream when signals not already covered (e.g. SIGHUP) below are involved.

https://github.com/stateful/runme/blob/1f01109235ac9e575a1f77afe4130551993dae27/internal/runner/command.go#L537-L546

I am creating a separate issue to look into improving ☝️ (cc @adambabik).

### DIFF
--- a/internal/runner/service.go
+++ b/internal/runner/service.go
@@ -442,10 +442,6 @@ func (r *runnerService) Execute(srv runnerv1.RunnerService_ExecuteServer) error 
 
 	logger.Info("command was finalized successfully")
 
-	if exitCode == -1 {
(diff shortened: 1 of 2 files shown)


sourishkrout:
Retesting https://github.com/stateful/runme/issues/413 again, I realized that it wasn't covered in https://github.com/stateful/runme/pull/443 like I first thought (my bad).

pastuxso (approved):

Issue #447, opened 2023-12-08
Issues with `.gitignore` in ProjectService


sourishkrout:
Two issues. One nested `.gitignore` files have no effect. When one runs `npm install` inside `./integration/subject` the `.gitignore` is supposed to exclude `node_modules` but won't:

```
./integration/subject/.gitignore
./internal/project/testdata/dir-project/.gitignore
./.gitignore
```

Another issue appears to be that glob style excludes don't seem to take effect either:

```
**/node_modules
```

PR #448, opened 2023-12-10
Fix .gitignore in internal/projects

adambabik:
It fixes issues described in https://github.com/stateful/runme/issues/447:
* `node_modules` was explicitly specified as [a default value of `--ignore-pattern`](https://github.com/stateful/rdme/blob/d41435c27def3d1dce1e929bec0d80ae0c252022/internal/cmd/root.go#L74)
* Nested `.gitignore`s are fixed with [`strings.Split(path, string(filepath.Separator))`](https://github.com/stateful/runme/pull/448/files#diff-cf4f5504efe89b688e171549faeccdeaa0723a75e9883aa22ef4fee73cee2b64R298)

Closes https://github.com/stateful/runme/issues/447


sourishkrout (approved):
👍 Works. Now I remember that we included `node_modules` by default. Slipped my mind.

PR #458, opened 2023-12-27
Add session metadata to serialization + frontmatter

sourishkrout:
This allows the extension to deny the notebook UX for markdown files written as session outputs until deserialization has been implemented (currently one-way; serialization only).

Chose to pass the session output metadata as option separately (not modifying frontmatter directly) because as long as outputs serialization is one-way, the extension uses two serializer runs (original markdown + session outputs markdown). If we decided we won't ever support outputs deserialization, we could create a separate operation. However, let's not jump ahead.

These changes are required for https://github.com/stateful/vscode-runme/pull/1025.


internal/document/editor/editorservice/service_test.go
```diff
@@ -323,13 +323,30 @@ func Test_parserServiceServer_Outputs(t *testing.T) {
 				},
 			},
 		}
-		notebook := &parserv1.Notebook{Cells: []*parserv1.Cell{cell}}
 
-		serializeOptions := &parserv1.SerializeRequestOptions{Outputs: &parserv1.SerializeRequestOutputOptions{Enabled: true, Summary: true}}
+		prasedFm := "---\nrunme:\n  id: 01HF7B0KK32HBQ9X4AC2GPMZG5\n  version: v2.0\nsidebar_position: 1\ntitle: Examples\n---"
```
pastuxso, line 327: Do you mean parsedFm?

pastuxso (approved):
Apart from the typo, looks good to me.

internal/document/editor/editorservice/service_test.go
sourishkrout: Thanks! Totally a typo. Fixed it.

PR #459, opened 2023-12-28
Skip files that include Runme session IDs (those are output files)

sourishkrout:
The extension will write Outputs to a separate session file. The idea is to make saving outputs available ahead of load (deserialization) and chunk up the dev effort required to unlock both.

Since, by default, the Session Outputs file will be colocated with its original document, it's highly likely to be included when a project is being loaded. The extension UX rejects running Session Outputs files in the notebook UX (https://github.com/stateful/vscode-runme/pull/1025).

This PR introduces a check for a Runme session ID, which will skip these docs/tasks accordingly.

PS: Also brings back a launch cfg to attach to dlv which is required for tui debugging.

### DIFF
--- a/internal/project/project.go
+++ b/internal/project/project.go
@@ -410,6 +410,11 @@ func getCodeBlocksFromFile(path string) (document.CodeBlocks, error) {
 func getCodeBlocks(data []byte) (document.CodeBlocks, error) {
 	identityResolver := identity.NewResolver(identity.DefaultLifecycleIdentity)
 	d := document.New(data, identityResolver)
+
+	if f, _ := d.Frontmatter(); f != nil && f.Runme.Session.ID != "" {
+		return document.CodeBlocks{}, nil
+	}
+
 	node, err := d.Root()
 	if err != nil {
 		return nil, err
--- a/internal/project/project.go
+++ b/internal/project/project.go
@@ -411,8 +411,8 @@ func getCodeBlocks(data []byte) (document.CodeBlocks, error) {
 	identityResolver := identity.NewResolver(identity.DefaultLifecycleIdentity)
 	d := document.New(data, identityResolver)
 
-	if f, _ := d.Frontmatter(); f != nil && f.Runme.Session.ID != "" {
-		return document.CodeBlocks{}, nil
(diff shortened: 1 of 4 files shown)


sourishkrout:
Please review as per description, @adambabik.

adambabik:
> The idea is to make saving outputs available ahead of load (deserialization) and chunk up the dev effort required to unlock both.

I don't quite understand "[...] saving outputs available ahead of load (deserialization)". Is it to cover the case of a new or modified and unsaved file?

internal/project/project.go
```diff
@@ -410,6 +410,11 @@ func getCodeBlocksFromFile(path string) (document.CodeBlocks, error) {
 func getCodeBlocks(data []byte) (document.CodeBlocks, error) {
 	identityResolver := identity.NewResolver(identity.DefaultLifecycleIdentity)
 	d := document.New(data, identityResolver)
+
+	if f, _ := d.Frontmatter(); f != nil && f.Runme.Session.ID != "" {
```
adambabik, line 414: Suggestion: To be completely safe in the case the API changes, I recommend:

```suggestion
	if f, err := d.Frontmatter(); err == nil && f != nil && f.Runme.Session.ID != "" {
```

internal/project/project.go
```diff
@@ -410,6 +410,11 @@ func getCodeBlocksFromFile(path string) (document.CodeBlocks, error) {
...
+		return document.CodeBlocks{}, nil
```
adambabik, line 415: Hint: This is more idiomatic and does not allocate any memory (negligible in this case):

```suggestion
		return nil, nil
```

adambabik (approved):

sourishkrout:
> > The idea is to make saving outputs available ahead of load (deserialization) and chunk up the dev effort required to unlock both.
> 
> I don't quite understand "[...] saving outputs available ahead of load (deserialization)". Is it to cover the case of a new or modified and unsaved file?

Neither. It's to get closer to make https://github.com/stateful/vscode-runme/pull/996 an opt-out versus opt-in feature (which it currently is). Runme knows how to write markdowns with cell outputs but can't do the reverse (yet). See screenshot in PR for clarity.

PR #462, opened 2024-01-03
Respect WithFindRepoUpward

sourishkrout:
This matches the existing behavior where sub-dirs within a repo will start at the root of the repo instead of the context sub-dir.

Probably just an oversight.

### DIFF
--- a/internal/project/project.go
+++ b/internal/project/project.go
@@ -283,8 +283,12 @@ func (p *Project) load(
 
 	switch {
 	case p.repo != nil:
-		// Even if p.repo is not nil, the current logic is identical to
-		// a dir-based project.
+		wt, err := p.repo.Worktree()
+		if err != nil {
+			p.logger.Warn("invalid worktree for repo; using base directory instead", zap.Error(err))
+		} else {
+			p.fs = wt.Filesystem
+		}
 		fallthrough
 	case p.fs != nil:
 		p.loadFromDirectory(ctx, eventc, options)
--- a/internal/project/project.go
+++ b/internal/project/project.go
@@ -181,6 +181,14 @@ func NewDirProject(
 		return nil, errors.WithStack(err)
 	}
 
+	if p.repo != nil {
+		wt, err := p.repo.Worktree()
+		if err != nil {
+			return nil, errors.WithStack(err)
+		}
+		p.fs = wt.Filesystem
+	}
+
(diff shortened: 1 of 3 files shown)


internal/project/project.go
```diff
@@ -283,8 +283,12 @@ func (p *Project) load(
 
 	switch {
 	case p.repo != nil:
-		// Even if p.repo is not nil, the current logic is identical to
-		// a dir-based project.
+		wt, err := p.repo.Worktree()
```
adambabik, line 286: Should we do this in the constructor instead?

adambabik:
> This matches the existing behavior where sub-dirs within a repo will start at the root of the repo instead of the context sub-dir.
> 
> Probably just an oversight.

It makes sense. And yes, it's an oversight. I forgot to cover this case at all.

internal/project/project_test.go
```diff
@@ -240,6 +240,144 @@ func TestProjectLoad(t *testing.T) {
 		}
 	})
 
+	gitProjectNestedDir := testdata.GitProjectNestedPath()
+
+	t.Run("GitProjectWithNested", func(t *testing.T) {
```
adambabik, line 245: If the result of `NewDirProject(gitProjectNestedDir)` and `NewDirProject(gitProjectDir)` is expected to be the same, I suggest simplified test:

```go
pRoot, err := NewDirProject(gitProjectDir)
pNested, err := NewDirProject(gitProjectNestedDir)

// Here, either compare `pRoot.fs` and `p.Nested` or collect load events for both and verify that the contents of slices are equal. The second is more resilient to implementation changes. The first requires a change I suggested in another comment.
```

sourishkrout:
> > This matches the existing behavior where sub-dirs within a repo will start at the root of the repo instead of the context sub-dir.
> > Probably just an oversight.
> 
> It makes sense. And yes, it's an oversight. I forgot to cover this case at all.

If you agree that moving it upstream is better, i'm happy to do that. I thought about it but was unclear/unsure if it might have unintended effects. Let me address that as part of your feedback.

adambabik:
> If you agree that moving it upstream is better, i'm happy to do that.

Thanks! From the description it is clear that if we are in a sub-dir of a repo, the project should be anchored in the root of the repo. Doing it in the constructor will result in more consistent output from other methods and in logs.

PR #465, opened 2024-01-04
Restore project flag behavior

sourishkrout:
Empty project flag actually means default to cwd but with find repo upwards whereas non-empty project flag pins the specified dir (without applying find repo upwards).

PS: This matches the behavior of the pre-project-svc implementation.

### DIFF
--- a/internal/cmd/list.go
+++ b/internal/cmd/list.go
@@ -58,7 +58,7 @@ func listCmd() *cobra.Command {
 				block := task.CodeBlock
 				lines := block.Lines()
 				relPath, err := filepath.Rel(getCwd(), task.DocumentPath)
-				if err != nil || !isTerminal(io.Out.Fd()) {
+				if err != nil {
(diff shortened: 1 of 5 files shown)


pastuxso (approved):

internal/cmd/list.go
```diff
@@ -58,7 +58,7 @@ func listCmd() *cobra.Command {
 				block := task.CodeBlock
 				lines := block.Lines()
 				relPath, err := filepath.Rel(getCwd(), task.DocumentPath)
-				if err != nil || !isTerminal(io.Out.Fd()) {
```
sourishkrout, line 61: @adambabik removing the tty check for consistency. as mentioned, we'll get an absolute path here when the repo’s worktree is assigned from searching upward for `.git`. in all other cases the path is relative. i thought about fixing it upstream but it seemed ugly having to reconstruct the fs struct with a relative as opposed to an absolute path.

PR #466, opened 2024-01-04
Bump test dependencies for integration tests

sourishkrout:
Only being used to have a source of different markdown files.


sourishkrout:
@pastuxso quick review please. these are dependencies that are only relevant for integration testing.

pastuxso (approved):

PR #468, opened 2024-01-05
TUI using relative paths

sourishkrout:
Apply same logic to to both TUI and list commands. Added a test case to guard against unintentional breakage.

cc @adambabik.

### DIFF
--- a/internal/cmd/common.go
+++ b/internal/cmd/common.go
@@ -436,7 +436,7 @@ func replaceVarValue(ev CommandExportExtractMatch, newValue string) string {
 }
 
 func getRelativePath(base string, documentPath string) string {
-	relPath, err := filepath.Rel(getCwd(), documentPath)
+	relPath, err := filepath.Rel(base, documentPath)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


internal/cmd/common.go
```diff
@@ -434,3 +434,11 @@ func replaceVarValue(ev CommandExportExtractMatch, newValue string) string {
 	replacedText := fmt.Sprintf("%s=%q", parts[0], newValue)
 	return replacedText
 }
+
+func getRelativePath(base string, documentPath string) string {
+	relPath, err := filepath.Rel(getCwd(), documentPath)
```
adambabik, line 439:
```suggestion
	relPath, err := filepath.Rel(base, documentPath)
```

adambabik (changes requested):

internal/cmd/common.go
sourishkrout: whoopsie

pastuxso (approved):
LGTM :+1:

Issue #467, opened 2024-01-04
[internal/project] `DocumentPath` in tasks is consistent


adambabik:
Recently, we discovered [some inconsistencies](https://github.com/stateful/runme/pull/465#discussion_r1442169829) when observing paths returned from `runme ls` and `runme tui`, and more precisely from `Project.Load()` from the `internal/project` package. 

The special case is when the project is bootstrapped by walking the directory upwards. However, the same thing happens if `--project` is an absolute path.

There are two options:
- Always return absolute paths and do post processing by the caller.
- Always return relative paths; relative to the `Project.Root()`.

cc @sourishkrout

sourishkrout:
> * Always return absolute paths and do post processing by the caller.

let's do this please. thank you @adambabik!

PR #469, opened 2024-01-05
internal/project: force absolute paths

adambabik:
This change should merge nicely with https://github.com/stateful/runme/pull/468.

Closes https://github.com/stateful/runme/issues/467


sourishkrout (approved):
👍 LGTM. @adambabik please lemme know if we can merge.

PS: Sonar won't see the integration tests as part of coverage which is why we are ok with 77.8%

PR #472, opened 2024-01-10
Copy runme binary to extension

peraltafederico:
This PR copies the binary created for the kernel into the /bin/ folder of vscode-runme when running `proto/dev`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,6 +98,7 @@ proto/clean:
 proto/dev: build proto/clean proto/generate
 	rm -rf $(RUNME_EXT_BASE)/node_modules/@buf/stateful_runme.community_timostamm-protobuf-ts/runme
 	cp -vrf internal/gen/proto/ts/runme $(RUNME_EXT_BASE)/node_modules/@buf/stateful_runme.community_timostamm-protobuf-ts
+	rm -f $(RUNME_EXT_BASE)/bin/runme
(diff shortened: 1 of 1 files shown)


peraltafederico:
Note: Tests are failing but they seem to be fixed in #471

sourishkrout (changes requested):
The extension has a setting `runme.server.binaryPath` that can specify an arbitrary path to the binary. I think we prefer dependencies to be pull vs push.

The only exception is the generation of protobufs and code. We inverted that because we have a lot fewer people work on kernel & extension vs just the extension.

I personally use a nested `.vscode/settings.json` with:
```
{
  "runme.server.binaryPath": "/Users/sourishkrout/Projects/stateful/oss/runme/runme",
}
```

peraltafederico:
> The extension has a setting `runme.server.binaryPath` that can specify an arbitrary path to the binary. I think we prefer dependencies to be pull vs push.
> 
> The only exception is the generation of protobufs and code. We inverted that because we have a lot fewer people work on kernel & extension vs just the extension.
> 
> I personally use a nested `.vscode/settings.json` with:
> 
> ```
> {
>   "runme.server.binaryPath": "/Users/sourishkrout/Projects/stateful/oss/runme/runme",
> }
> ```

Excellent @sourishkrout ! I'll close this PR and add that ot the docs of `vscode-runme`.

PR #474, opened 2024-01-14
Implement the new command run-locally

adambabik:
The main advantage is that it supports executing interactive commands, including runme-in-runme, and pipes (`|`).

On top of that, there is the new package internal/command. It gives a better interface to create commands, as well as unifies the idea of a program config that now is defined in a proto file.

`runme run-locally` is not an equivalent of `runme run`. It's to be decided how these two can be merged. `run-locally` is hidden from the runme usage.

## `runme run-locally`

This is a new command. It does not support many flags known from `runme run`, but it can be used to test the new capabilities like piping and REPLs as showed below.

`run-locally [command1 command2 ...]` also changes the approach to selecting commands. You can pass multiple of them separated by space. Additionally, glob names are supported like `print-*` which will execute all commands starting with `print-`.

Piping can be tested with the following script:

````sh
```sh {"name": "pipe"}
cat - | tr a-z A-Z
```
````

And run like this: `echo test | go run . --filename test-run-locally.md run-locally pipe`.

You can also run Python repl:

````sh
```sh {"name": "python"}
python
```
````

`go run . --filename test-run-locally.md run-locally python`

And runme-in-runme:

````sh
```sh {"name": "runme"}
runme --filename test-run-locally.md ls
```
````

`go run . --filename test-run-locally.md run-locally runme`

## Config

The commands, either created programmatically or created in a service from user's input, are now based on a structure described in [config.proto](./internal/api/runme/runner/v2alpha1/config.proto). This config contains all details to create the runnable command the only variable is I/O.

## `internal/command` package

Firstly, it implements utilities around the command config, including parsing a code block to config.

Secondly, there are two implementations of commands which differ how they are execute:
* Native uses OS stdin, stderr, and stdout
* Virtual uses a pseudo terminal

## runmev2alpha1

There is also a new service registered on the same server. It has the same methods to manage sessions and `Execute()` to run commands. The difference is the request message, which now contains the `Config` message described above.

It also contains a new method `ResolveEnv()` which takes a script or a list of commands, detects environment variables in them, and tries to resolve them using project, session, or provided env in the request.

### CODE REVIEW

**Review by sourishkrout — APPROVED** [review]
Most minor comments. One thought I had was to rename `ResolveEnv` to `ResolveExport(s)` since they are ENVs but specifically the once that are exported. I think that would resolve overloading the ENV terminology. Wdyt @adambabik?

Outside of that, I'd suggest we get this merged before I try integrating it with the extension. Treat any issues/requests as separate issues after.

**Comment by sourishkrout** [comment]
> Most minor comments. One thought I had was to rename `ResolveEnv` to `ResolveExport(s)` since they are ENVs but specifically the once that are exported. I think that would resolve overloading the ENV terminology. Wdyt @adambabik?
> 
> Outside of that, I'd suggest we get this merged before I try integrating it with the extension. Treat any issues/requests as separate issues after.

On further examination, I understand now that `ResolveEnv`, in fact, returns all ENV vars, not just the exports inside of the script/command, which invalidates the point I was making.

File: internal/api/runme/runner/v1/runner.proto [hunk 1]
```diff
@@ -201,4 +254,11 @@ service RunnerService {
   // Subsequent "ExecuteRequest" should only contain "input_data" as
   // other fields will be ignored.
   rpc Execute(stream ExecuteRequest) returns (stream ExecuteResponse) {}
+
+  // ResolveEnv resolves environment variables from a script or a list of commands
+  // using the provided sources, which can be a list of environment variables,
+  // a session, or a project.
+  // The result contains all found environment variables. If the env is in any source,
+  // it is considered resolved. Otherwise, it is makred as unresolved.
```
sourishkrout, line 262: typo, no biggie

adambabik:
> > Most minor comments. One thought I had was to rename `ResolveEnv` to `ResolveExport(s)` since they are ENVs but specifically the once that are exported. I think that would resolve overloading the ENV terminology. Wdyt @adambabik?
> > Outside of that, I'd suggest we get this merged before I try integrating it with the extension. Treat any issues/requests as separate issues after.
> 
> On further examination, I understand now that `ResolveEnv`, in fact, returns all ENV vars, not just the exports inside of the script/command, which invalidates the point I was making.

No, it should not return all env. It should return only ones that are explicitly exported in the script or commands list.

With regard to your suggestion, I was thinking that maybe we can extend the functionality it in the future so that not only `export ENV` are recognised, but other variables usage like `${ENV}` as well. Still, I think you're right that `ResolveEnv` is not the best name. Maybe `ResolveVariables`?

sourishkrout:
> No, it should not return all env. It should return only ones that are explicitly exported in the script or commands list.
> 
> With regard to your suggestion, I was thinking that maybe we can extend the functionality it in the future so that not only `export ENV` are recognised, but other variables usage like `${ENV}` as well. Still, I think you're right that `ResolveEnv` is not the best name. Maybe `ResolveVariables`?

Gotcha. I'm playing around with this a bit more today/tmrw. Lemme get back to you on the name change. Perhaps `ResolveVariables` or `ResolveVars` is a better choice. I will let you know soon.

adambabik:
I added [an integration test](https://github.com/stateful/runme/pull/474/commits/3a6dfd4c09682c7dac1fd869324cf09fed686d38#diff-4b6c8026ba5048f41ee34fb3190e4374ba60949c506a1e491b2025ec86b5768eR43-R53) which validates that `ResolveEnv` does not return not exporter vars.

sourishkrout:
> > No, it should not return all env. It should return only ones that are explicitly exported in the script or commands list.
> > With regard to your suggestion, I was thinking that maybe we can extend the functionality it in the future so that not only `export ENV` are recognised, but other variables usage like `${ENV}` as well. Still, I think you're right that `ResolveEnv` is not the best name. Maybe `ResolveVariables`?
> 
> Gotcha. I'm playing around with this a bit more today/tmrw. Lemme get back to you on the name change. Perhaps `ResolveVariables` or `ResolveVars` is a better choice. I will let you know soon.

OK, so I think it'd like roll with `ResolveVars` (I'm also ok with `ResolveVariables`) if `vars` is too short, @adambabik, please.

PR #476, opened 2024-01-17
Fixes the language resolution from the FencedCodeBlock AST

pastuxso:
Closes: #475

This is a temporary solution, I think we can deal with it by extending the Parser with custom logic to handle the language resolution for this edge case and generate an AST according to our necessity.

Some examples of Goldmark extensions:

https://github.com/yuin/goldmark#list-of-extensions


adambabik (commented):
It would be great to add a unit test to `node_test.go`, similar to `TestCodeBlock_Intro`, which will verify that `Language()` and `Attributes()` return expected values for the problematic case.

The change itself LGTM!

sourishkrout (approved):
👍 LGTM. Let's merge.

PR #479, opened 2024-01-21
Remove unused connect protocol references

adambabik:
There are references to Buf connect protocol, which is not used and has not been maintained for new services.


pastuxso (approved):

sourishkrout (approved):
This is great 👍 . Been meaning to remove the unused Connect for some time now.

PR #480, opened 2024-01-24
Add deprecation notice to `runme branch` and aliases

sourishkrout:
Ahead of impending removal with the next major version.


duvanmonsa (approved):
LGTM

PR #486, opened 2024-02-02
Consolidate GRPC samples

sourishkrout:
Let's keep them in one place.

PS: Fixed a bug where empty options would cause a panic.


sourishkrout:
Quick review, please @pastuxso. Just consolidating some docs based on @adambabik's prev merge.

pastuxso (approved):

PR #488, opened 2024-02-03
Fix double quoted expression stmts

sourishkrout:
If it's not a literal, this used to panic. Please see diff.

### DIFF
--- a/examples/grpc-samples/deploy-helpers.sh
+++ b/examples/grpc-samples/deploy-helpers.sh
@@ -1,0 +1,42 @@
+export PROJECT_PATH="$( dirname $( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd ) )"
+export PROJECT_ID="$( gcloud config get-value project --format="value(.)" )"
+export REGION="us-central1"
+
+if [[ -z "$PROJECT_ID" ]]; then
+  echo "Set a gcloud project (via gcloud config or CLOUDSDK_CORE_PROJECT)" >&2
+  exit 1
+fi
+
+if ! python3 -c 'import cleverhans' &>/dev/null; then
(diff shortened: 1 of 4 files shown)


sourishkrout:
@adambabik came up while I was throwing "complex" scripts at the API: https://app.runme.dev/cell/cls6b7ufu000rdpzye7yv442h

examples/grpc-samples/deploy-helpers.sh
```diff
@@ -0,0 +1,42 @@
+export PROJECT_PATH="$( dirname $( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd ) )"
```
adambabik, line 1: @sourishkrout are you sure you want to keep this file?

internal/command/env_resolver_test.go
```diff
@@ -78,6 +78,13 @@ func TestEnvResolverParsing(t *testing.T) {
 				{Name: "TEST_VALUE_EXPR"},
 			},
 		},
+		{
+			name: "double quoted value expression",
+			data: `export TEST_DBL_QUOTE_VALUE_EXPR="$( echo -n 'value')"`,
```
adambabik, line 83:
```suggestion
			data: `export TEST_DBL_QUOTE_VALUE_EXPR="$(echo -n 'value')"`,
```

adambabik (approved):

examples/grpc-samples/deploy-helpers.sh
sourishkrout: Yes, otherwise https://github.com/stateful/runme/blob/expression-in-double-quotes/examples/grpc-samples/services.md#complex-script won't run.

PR #493, opened 2024-02-09
Fix .env which broke as part of moving cmds to the project service

### LINKED ISSUE #492: .env loading seems broken in CLI
Running `runme run hello1` with following files will not print "SOMETHING=this" as expected.

`.env`:
```
SOMETHING=this
```

`README.md`:
```
---
shell: bash
---

\`\`\`shell { name=hello1 }
echo SOMETHING=[${SOMETHING}]
\`\`\`
```

sourishkrout:
Please see #492 for details.

Before vs after:
```
❯ runme run hello1
SOMETHING1=[]
❯ ./runme run hello1
SOMETHING1=[this]
```

Fixes #492.

### DIFF
--- a/internal/runner/client/client_local.go
+++ b/internal/runner/client/client_local.go
@@ -47,6 +47,10 @@ func NewLocalRunner(opts ...RunnerOption) (*LocalRunner, error) {
 		return nil, err
 	}
 
+	if err := r.ApplyProjectEnvs(); err != nil {
+		return nil, err
+	}
+
 	if r.logger == nil {
 		r.logger = zap.NewNop()
 	}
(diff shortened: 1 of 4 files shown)


sourishkrout:
@adambabik, please lemme know if this fix works for you. Please ignore the code coverage, it's low due to integration vs unit test (not in coverage report).

internal/runner/client/client_local.go
```diff
@@ -95,7 +99,6 @@ func (r *LocalRunner) newExecutable(task project.Task) (runner.Executable, error
 		Logger:  r.logger,
 	}
 
-	// TODO(adamb): what about `r.envs`?
```
sourishkrout, line 98: @adambabik sounded like you had a hunch here

sourishkrout:
@adambabik, please do a retro-review. Going through a review with @pastuxso right now so we can ship a fix quickly.

pastuxso (approved):

adambabik:
@sourishkrout I think it's ok. The way the environment variables are handled in `internal/runner/client` is a bit unclear, but in the end all env go to a session which is passed to the runner via `ExecutableConfig` and from there the env are eventually taken.

I will send a follow up PR with clean up.

PR #494, opened 2024-02-09
Follow up to a fix for #492

adambabik:
I cleaned up env management. This PR also fixes experimental `runme run-locally`.

Relates to #492


internal/runner/client/client.go
```diff
@@ -191,7 +191,7 @@ func WithEnableBackgroundProcesses(enableBackground bool) RunnerOption {
 
 func WithEnvs(envs []string) RunnerOption {
 	return withSettings(func(rs *RunnerSettings) {
-		rs.envs = envs
+		rs.envs = append(rs.envs, envs...)
```
adambabik, line 194: @sourishkrout this is an only place that could cause some issues in the future. The rest in `internal/runner` is clean up.

sourishkrout (approved):
LGTM 👍.

PR #499, opened 2024-02-13
Fix bug where non-frontmatter is being mistaken as JSON

sourishkrout:
While at it replace NPM modules to discover markdown files for parser testing with GOPATH.

PS: Unrelated change: allow full path for `--filename` flag as long as it's absolute.


internal/cmd/common.go
```diff
@@ -44,7 +44,14 @@ func getProject() (*project.Project, error) {
 
 	if fFileMode {
 		var err error
-		proj, err = project.NewFileProject(filepath.Join(fChdir, fFileName), opts...)
+
+		filePath := filepath.Join(fChdir, fFileName)
+		_, err = os.Stat(fFileName)
+		if filepath.IsAbs(fFileName) && !os.IsNotExist(err) {
```
adambabik, line 50: Question: It's a bit odd that we will continue if `err != nil`. I guess the intention was to fallback to `NewFileProject`?

adambabik (approved):
Good idea with getting rid of `integration/subject` 👍

internal/cmd/common.go
sourishkrout: Yes, correct. I'll add a comment here.

PR #502, opened 2024-02-17
Fix exit code for run command

adambabik:
Consider the following Markdown file:

````sh
```sh { "name": "test-singleline" }
failhereplease && echo single
```

```sh { "name": "test-multiline" }
failhereplease
echo multi
```
````

When executing it with `runme run`, the exit code for both should not be zero. This PR makes sure this is the actual behaviour. 

The problem was in how we collect the environment variable. The solution was to switch to a clean up function instead of doing it inline, which could overwrite the original exit status.

Another problem, which is not covered in this PR, is to surface the actual shell exit code rather than any non-zero value convert to `1`. This is due to an explicit conversion in:
https://github.com/stateful/runme/blob/064e483cccc50a92e57c652021d517115595931a/main.go#L22-L25


sourishkrout (approved):
👍 LGTM. Let's create a follow-up issue for retaining the exit code as described.

PR #505, opened 2024-02-19
Implement configuration in runme.yaml

adambabik:
This PR introduces an implementation of file-based configuration in a form of `runme.yaml`. The main goal is to centralized the configuration in a single place and define it per workspaces rather than implicitly by each user using CLI flags or VS Code preferences.

Check out `experimental/runme.yaml` in order to learn the format of the config. It is described by `internal/api/runme/config/v1alpha1/config.proto`.

`beta run` is the only command that understands the file-based configuration. You can run it in the root directory of the runme project: 

```sh
go run . beta run echo
go run . beta run --category=a b
```

The implementation uses https://github.com/spf13/viper to read the config from a file. I also implement filters based on https://github.com/expr-lang/expr which on the one hand gives flexibility, on the other hand might be harder to reason about. The `beta run` implementation contains an example how a simple `--category` flag can be translated to a filter. From the implementation perspective, it's also non-trivial.

There are still things to do:

- [x] There is no validation. The proposal is to add rules directly in `internal/api/runme/config/v1alpha1/config.proto`.
- [x] Assess whether the filters is a good idea.
- [x] `internal/config/autoconfig` has no tests.
- [ ] Do not allow `runme.yaml` to reference directories and files up the tree relative to its base.
- [ ] Resolve `TODO`s.


internal/api/runme/config/v1alpha1/config.proto
```diff
@@ -0,0 +1,55 @@
...
+  message Filter {
+    string type = 1;
+
+    string condition = 2;
+  }
+
+  message Server {
+    string address = 1;
+
+    string ssl_dir = 2;
```
sourishkrout, line 43: `tls_dir`? SSL is so yesteryear :-)

sourishkrout (approved):
👍 Looks good for a merge. I will have to spend a bit more time with filters and the specific configuration options to have more concrete opinions but that should stop us from merging this PR and work of smaller increments.

Love the `beta` subcommand group 👍 .

Issue #504, opened 2024-02-19
`runme run` fails with error in default location (docker image)


sourishkrout:
Suspecting a fs permission issue not being correctly handled since we moved to the Project service. This should not error but instead display something like `could not execute command: no named code blocks, consider adding flag --allow-unnamed`.

```
$ docker run -it statefulhq/runme ls
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
could not execute command: lstat /proc/1/fd/11: no such file or directory
```

PR #507, opened 2024-02-20
breaking(docker): move runme to /opt/bin/

pastuxso:
Relocate the runme executable to the **`/opt/bin`** directory and designate **`/opt/var/runme`** as the working directory. This adjustment prevents crashes resulting from accessing the root filesystem (**`/`**) when the image is used outside a multi-stage build.

The right way to use the docker image outside multi-stage builds is the following:

* Use the default working directory:
  **`docker run -it --volume /your/runbooks:/opt/var/runme statefulhq/runme`** 

* Changing the working directory:
  **`docker run -it -w /runbooks --volume /your/runbooks:/runbooks statefulhq/runme`**

* Passing the **`--project`** flag:
  **`docker run -it --volume /your/runbooks:/runbooks statefulhq/runme --project /runbooks`** 
  

Sample:

image

Closes: #504

**For multi-stage dependants that are using the binary from **/runme** it will be a breaking change.**

### DIFF
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -3,11 +3,14 @@ LABEL maintainer="StatefulHQ <mail@stateful.com>"
 
 RUN apk --no-cache add ca-certificates
 RUN mkdir /newtmp && chown 1777 /newtmp
+RUN mkdir -p /opt/var/runme /opt/bin
 
 FROM scratch
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 COPY --from=build /newtmp /tmp
+COPY --from=build /opt /opt
 
-COPY runme /
+COPY runme /opt/bin/
+WORKDIR /opt/var/runme
 
-ENTRYPOINT ["/runme"]
+ENTRYPOINT ["/opt/bin/runme"]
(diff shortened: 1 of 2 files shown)


Dockerfile.alpine
```diff
@@ -8,6 +8,6 @@ FROM scratch
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 COPY --from=build /newtmp /tmp
 
-COPY runme /
-
-ENTRYPOINT ["/runme"]
+COPY runme /opt/
```
sourishkrout, line 11: What do you think about `/opt/runme/runme`? seems that's what people would expect locally on a Linux, no?

sourishkrout (changes requested):
@pastuxso see question

Dockerfile.alpine
pastuxso: @sourishkrout  I believe it's a good idea. In that case, if we aim for better POSIX compliance, the binary should reside in **/opt/bin/runme** and the working directory should be set to **/opt/var/runme**. However, considering we're using the _scratch_ image, I think it suffices to separate it only into /opt since we anticipate using volumes to mount the runbooks.

What do you think?

sourishkrout: We're still finding out how people want to use the image. We originally created it for a user who wanted to depend on it using a multi-build process (see Discord `#help` for context). Just because our docs give one example does not mean we want to limit ourselves to it.

So, since we're changing it, we should aim for the POSIX compliance you are describing, @pastuxso.

PR #508, opened 2024-02-20
[WIP] Expansion to "ResolveProgram"

sourishkrout:
Working on expanding `ResolveVars` into `ResolveProgram`. The idea is to "replace" the [resolveRunProgramExecution](https://github.com/stateful/vscode-runme/blob/main/src/extension/executors/runner.ts#L433-L439) function entirely with a service.

Other notable changes are:
- Service solely decides the resolution/prompt status of variables. Based on the `mode` mode passed by the client.
- Only top-level `export` vars are being considered. We don't want to interfere with anything tangled up in flow-control.
- Service rewrites scripts (returned as command sequence) to remove unresolved vars to be passed as part of ENV store.
- Likely removing option to pass `commands` since `ResolveProgram` will be responsible for turning shell scripts into commands sequence (skipped for shebang).

Progress to date https://app.runme.dev/cell/clsufpieq001flv0oggzrqqlb using the following request:

```json
{
    "script": "export NAKED=this is a value\nexport TEST=$( (pwd) )\nexport NAME='Sebastian Mustermann'\nexport PWD=`pwd`\necho \"Hello World, $NAME!\"\necho $TEST\n\ntest() {\n    export INNER=\"nested\"\n}\n\ntest",
    "mode": 0,
    "env": [
        "RUNME_ID=01HF7B0KK32HBQ9X4AAD3Z5V14",
        "NAME=Luna"
    ],
    "sessionId": "01HQ1Q3YGSHCXHHZ53GQVXJ70M",
    "sessionStrategy": 0
}
```

PS: Aiming to get full integration working before rewriting tests, porting to v2alpha1, and considering CLI integration.

PR #512, opened 2024-02-23
Evolve ResolveVars into ResolveProgram

sourishkrout:
Supersedes https://github.com/stateful/runme/pull/508. "Program" as in "Cell Shell Program".

Notable changes are:
- [x] Service solely decides the resolution/prompt status of variables. Based on the `mode` mode passed by the client.
- [x] Only top-level `export` vars are being considered. We don't want to interfere with anything tangled up in flow control.
- [x] Service rewrites scripts (returned as command sequence) to remove unresolved vars to be passed as part of ENV store.
- [x] Integrated with extension: https://github.com/stateful/vscode-runme/pull/1153
- [ ] Integration with CLI (intentionally pending).
- [ ] Removing option to pass `commands` since `ResolveProgram` (intentionally pending) will be responsible for turning shell scripts into commands sequence (skipped for shebang).

In a future version, a script to command normalization will be included, too. However, we can wait until further progress is made on the `runnerv2alpha1` refactor. Service is ported to runnerv2. However, the tests are somewhat incomplete (todos).

Example: https://us-central1.stateful.com/cell/clsyzdcv4000088eiccrten9s


sourishkrout:
I'm going to merge this on a successful build @adambabik. It'd still be interested in your review, please.

While it's not immediately apparent, this work is a precursor for what I call the "smart env store".

pastuxso (approved):
LGTM :rocket:

internal/command/program_resolver.go
```diff
@@ -0,0 +1,319 @@
...
+
+const (
+	ProgramResolverUnresolved ProgramResolverPrompt = iota
+	ProgramResolverResolved
+	ProgramResolverMessage
+	ProgramResolverPlaceholder
+)
+
+type ProgramResolverResult struct {
+	Prompt        ProgramResolverPrompt
```
adambabik, line 64: @sourishkrout this is hard to understand conceptually what `Prompt` really is. It seems like it's an indicator how the prompt in the VS Code should look like, based on the mode and whether an env is found in the sources. Is this correct?

If so, I would rename `ProgramResolverPrompt` to `ProgramResolverStatus` and change enum values to:
* `ProgramResolverStatusResolved` meaning that the env is resolved and no further action is needed.
* `ProgramResolverStatusUnresolvedWithMessage` meaning that the env is unresolved but there is a message to display.
* `ProgramResolverStatusUnresolvedWithPlaceholder` meaning that the env is unresolved but there is a placeholder/default value.

internal/command/program_resolver.go
```diff
@@ -0,0 +1,319 @@
...
+	return func() []string {
+		return env
+	}
+}
+
+// ProgramResolver uses a list of ProgramResolverSource to resolve environment variables
+// found in a shell program. The result contains all found environment variables.
+// If the env is in any source, it is considered resolved. Otherwise, it is marked
+// as unresolved.
+type ProgramResolver struct {
```
adambabik, line 37: Overall, I think the implementation would be easier to reason about if `ProgramResolver` didn't contain `mode`. Instead, there would be `ProgramResolverWithMode` which would look like this:

```go
struct ProgramResolverWithMode {
  *ProgramResolver
  Mode ProgramResolverMode
}
```

internal/runner/service.go
```diff
@@ -563,58 +563,51 @@ func runnerWinsizeToPty(winsize *runnerv1.Winsize) *pty.Winsize {
...
 		err = status.Error(codes.InvalidArgument, "either script or commands must be provided")
 	}
 	if err != nil {
 		return nil, err
 	}
 
-	response := &runnerv1.ResolveVarsResponse{}
+	response := &runnerv1.ResolveProgramResponse{
+		Commands: &runnerv1.ResolveProgramCommandList{
+			Lines: strings.Split(scriptRes.String(), "\n"),
```
adambabik, line 589: Using `\n` as a separator is not safe when we want to reconstruct the command list back. Within the script there might be escaped `\n` which here will not be properly handled. Likely the `Resolve` method need to work with a script and individual lines, which all should be valid shell commands, right?

internal/command/program_resolver.go
sourishkrout: Correct. I actually had it called `Status` before and renamed it to `Prompt` but can't remember the reason for it.

The renaming makes sense and is definitely more self-describing 👍 .

internal/runner/service.go
sourishkrout: Yep, you're right if, eg a heredoc was gonna be inside of the script the splitting would garble it.

The reason why I'd like to return a list of commands here is to make it explicit in the API that normalization/transformation from script->commands has already happened. However, I'm clearly not handling this well.

Feel free to propose a solution @adambabik.

sourishkrout: Interestingly, it actually works! Please see: https://us-central1.stateful.com/cell/clt0fkcs6000288eiz1b9nbnu

I believe the reason is because the splitting happens on a string-level as opposed to the bytes where `\\n` vs `\n` are unescaped appropriately. Wdyt @adambabik?

adambabik: This is an interesting case but I don't quite understand how Go processes escaped `\n` in this case. 

I was thinking about a bit simpler example, however, I am not sure if this is how the client will use this method. The current definition allows such usage, though. 

Check out [this test](https://github.com/stateful/runme/pull/513/files#diff-5c90f0532937e162b42092e036a66d065a0118edae2ce20350e53dfea2ebf52bR85-R131).

Basically, with a program as follows:
```
export FILE_NAME=default.txt
cat >"$FILE_NAME" <<EOF
Some content with\nnew line
EOF
```

I expect the output to be:
```
#
# FILE_NAME set in smart env store
# "export FILE_NAME=default.txt"

cat >"$FILE_NAME" <<EOF
Some content with\nnew line
EOF
```

Instead, I got (+/- new line at the end):
```
#
# FILE_NAME set in smart env store
# "export FILE_NAME=default.txt", 

cat >"$FILE_NAME" <<EOF
Some content with
new line
EOF
```

Also, the question is why `ResolveProgramResponse` does not contain the source in the same format as in the input but always return a list of commands. Should that be fixed?

sourishkrout: Yes, you're right. You managed to "reproduce" the problem. The only thing I can think of is that on the extension side `\n` is being escaped as `\\n`. I can x-check.

@adambabik perhaps it'd be better to use the [SingleLine](https://pkg.go.dev/mvdan.cc/sh/v3@v3.8.0/syntax#SingleLine) option and break the commands at `;` instead?

sourishkrout: image

@adambabik looks like the `\\n` is why it "works" ☝️

sourishkrout: Here's the same buf encoded as string printed with quotes (`%q`):

```
"cat >somefile <<EOF\nline1\nline2\\n\\n\nline3\nEOF\ncat somefile\n"
```

adambabik: In terms of the solution. Currently, the service accepts two kinds of program source: (1) a script as a single blob, (2) a list of commands. My first question would be whether we need both and in which cases the client uses one vs another.

The case of a script as a blob is easy, provided that the service returns it as a blob back. This is not the case now. Any reason for it?

For the case of a list of commands, the question is whether each line is a separate valid shell program. If not, then we need to concatenate it before processing and then split back. This is problematic. I will check if we can use something more complex than `\n` to do concatenate-and-split that our shell parser is ok with. We can also split it using a custom algorithm as we know the input and know how it can be modified, but I would prefer to avoid it.

PR #513, opened 2024-02-25
Refactor ProgramResolver

adambabik:
Follow up for https://github.com/stateful/runme/pull/512.


internal/command/program_resolver_test.go
```diff
@@ -2,205 +2,264 @@ package command
...
-			name: "no value",
-			data: `export TEST_NO_VALUE`,
+			name:    "no value",
+			program: `export TEST_NO_VALUE`,
 			result: &ProgramResolverResult{
-				Name: "TEST_NO_VALUE", Prompt: ProgramResolverMessage,
+				ModifiedProgram: true,
+				Variables: []ProgramResolverVarResult{
+					{
+						Status: ProgramResolverStatusUnresolved,
```
adambabik, line 26: @sourishkrout There are a few changes like this one in these tests, which I hope led to more consistent behaviour. 

All variables are unresolved here and:

* If a variable is empty (`export A`, `export A=`, `export A=""`) it is marked as `ProgramResolverStatusUnresolved`.
* If there is any unquoted value, it is marked as `ProgramResolverStatusUnresolvedWithMessage`. 
* If there is any quoted value, it is marked as `ProgramResolverStatusUnresolvedWithPlaceholder`.

internal/command/program_resolver_test.go
```diff
@@ -2,205 +2,264 @@ package command
...
-			runProgramResolverFixtureTestCase(tc, ProgramResolverModeSkip),
-		)
+		t.Run("ProgramResolverModeSkip_"+tc.name, func(t *testing.T) {
+			r := NewProgramResolver(ProgramResolverModeSkipAll)
+			buf := bytes.NewBuffer(nil)
+			got, err := r.Resolve(strings.NewReader(tc.program), buf)
+			require.NoError(t, err)
+			assert.EqualValues(t, tc.modifiedProgram, buf.String())
+			// In skip mode, all variables will be resolved.
+			if tc.result != nil {
```
adambabik, line 184: @sourishkrout I hope I figured this out correctly: when `ProgramResolverModeSkipAll`, then all variables are marked as resolved and their resolved value is equal to the original value.

adambabik (commented):
@sourishkrout watch out for field names changed in `runner.proto`. There is no new enum values though or deleted/new fields. `Status` enum is not backward compatible, as `STATUS_RESOLVED` is the last one.

internal/command/program_resolver_test.go
sourishkrout: That's fair. In any case, the extension will likely choose to prompt for `ProgramResolverStatusUnresolved` with a generic message. 👍

sourishkrout (approved):
👍 Thank you!

I will change the integration code on the extension-side. Do some acceptance testing and if successful merge this branch.

PR #517, opened 2024-02-27
514 bug go install command installs older version v182 instead of latest one

sourishkrout:
Fix `go install github.com/stateful/runme@latest` to install actual `latest` aka `v3` (right now).


pastuxso (approved):
LGTM :rocket:

PR #523, opened 2024-03-07
No implementation of `--filter` anywhere available - removing

sourishkrout:
The `--filter` flag was never used/implemented. Let's remove it.


pastuxso (approved):
LGTM :rocket:

PR #524, opened 2024-03-09
Increases tests coverage for env variables parsing

pastuxso:
Bumps the version of **stateful/godotenv** to the latest one to support parsing naked environment variables like `FOO= # comment`.


sourishkrout (approved):
👍 LGTM

Issue #528, opened 2024-03-13
Add `runme env store snapshot` subcommand


sourishkrout:
The command should print a tabular TUI akin to `runme list` with data returned from the `MonitorEnvStore` API.

Example client here: https://github.com/stateful/runme/blob/main/examples/grpc-client/main.go#L122-L145

It'd be great to run it where it exits after the first snapshot is received and alternatively with a `--watch` / `-w` flag just continuously refreshes. However, let's focus on building the non-refreshing version first.

PS: Please see the session strategy stuff to infer the sessionID. I believe having `RUNME_SESSION` in the ENV should work as long as it's being used inside of VS Code.

sourishkrout:
Done.

PR #532, opened 2024-03-14
Adds beta command `runme env store snapshot`

pastuxso:
This PR supports interacting with the **MonitorEnvStore** in the CLI.

Usage:

`runme env store snapshot --tlsdir=<RUNME TLS CERTIFICATES> --address=<RUNME SERVER ADDRESS> --session=<RUNME SESSION ID>`

The following variables will be used by default if you don't pass arguments:
 
`RUNME_SESSION`
`RUNME_SERVER_ADDR`
`RUNME_TLS_DIR`

Example:

`runme env store snapshot`

image

Issue: https://github.com/stateful/runme/issues/528


internal/runner/session.go
```diff
@@ -41,6 +41,9 @@ func NewSession(envs []string, proj *project.Project, logger *zap.Logger) (*Sess
 func NewSessionWithStore(envs []string, proj *project.Project, owlStore bool, logger *zap.Logger) (*Session, error) {
 	sessionEnvs := []string(envs)
 
+	id := ulid.GenerateID()
+	sessionEnvs = append(sessionEnvs, fmt.Sprintf("RUNME_SESSION=%s", id))
```
pastuxso, line 45: I added the RUNME_SESSION variable to infer its value later.

image

internal/cmd/environment.go
```diff
@@ -20,12 +29,103 @@ func environmentCmd() *cobra.Command {
...
+				}
+
+				return errors.Wrap(table.Render(), "failed to render")
+			}
+
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&addr, "address", "", "The Server address to connect to, i.e. 127.0.0.1:7865")
```
pastuxso, line 122: I'm wondering how to infer the connection when you run the command inside a notebook.

internal/runner/service.go
```diff
@@ -86,6 +91,8 @@ func (r *runnerService) CreateSession(ctx context.Context, req *runnerv1.CreateS
 
 	envs := make([]string, len(req.Envs))
 	copy(envs, req.Envs)
+	envs = append(envs, fmt.Sprintf("RUNME_RUNNER_ADDRESS=%s", r.address))
+	envs = append(envs, fmt.Sprintf("RUNME_RUNNER_TLS_DIR=%s", r.tlsDir))
```
pastuxso, line 95: It allows us to infer the server params to perform the runme env store snapshot

image

internal/cmd/environment.go
```diff
@@ -20,12 +29,103 @@ func environmentCmd() *cobra.Command {
...
+
+func storeSnapshotCmd() *cobra.Command {
+	var (
+		addr      string
+		tlsDir    string
+		sessionID string
+	)
+
+	cmd := cobra.Command{
+		Use:   "snapshot",
```
sourishkrout, line 59: @pastuxso there's a way to set a `Hidden: true` or something like it. Let's do that please as long as it experimental.

internal/runner/service.go
```diff
@@ -803,8 +816,8 @@ func convertToMonitorEnvStoreResponse(msg *runnerv1.MonitorEnvStoreResponse, sna
 			OriginalValue: item.Value.Original,
 			ResolvedValue: item.Value.Resolved,
 			Status:        status,
-			CreateTime:    item.Var.Created.String(),
-			UpdateTime:    item.Var.Updated.String(),
+			CreateTime:    item.Var.Created.Format(time.RFC3339),
```
pastuxso, line 819: @sourishkrout let's use the RFC3339 which is more standard for sharing dates through systems and compatible with ISO8601, It was tested with the extension and is working fine.

internal/env/env.go
```diff
@@ -13,3 +18,22 @@ func ConvertMapEnv(mapEnv map[string]string) []string {
 
 	return result
 }
+
+func CleaSessionVars(envs []string) []string {
```
sourishkrout, line 22: Clean?

internal/cmd/environment.go
```diff
@@ -20,12 +30,82 @@ func environmentCmd() *cobra.Command {
...
+	var (
+		addr      string
+		tlsDir    string
+		sessionID string
+	)
+
+	cmd := cobra.Command{
+		Hidden: true,
+		Use:    "snapshot",
+		Short:  "Dump environment variables to stdout",
```
sourishkrout, line 63: Copy & paste description leftover?

internal/cmd/server.go
```diff
@@ -104,7 +104,7 @@ The kernel is used to run long running processes like shells and interacting wit
 			parserv1.RegisterParserServiceServer(server, editorservice.NewParserServiceServer(logger))
 			projectv1.RegisterProjectServiceServer(server, projectservice.NewProjectServiceServer(logger))
 			if enableRunner {
-				runnerServicev1, err := runner.NewRunnerService(logger)
+				runnerServicev1, err := runner.NewRunnerService(logger, addr, tlsDir)
```
sourishkrout, line 107: don't quite understand why we need this

internal/cmd/environment.go
```diff
@@ -49,6 +129,46 @@ func environmentDumpCmd() *cobra.Command {
...
+		}
+
+		table.AddField(env.Name)
+		table.AddField(value)
+		table.AddField(env.Spec)
+		table.AddField(env.Origin)
+
+		t, err := time.Parse(time.RFC3339, env.UpdateTime)
+		if err == nil {
+			table.AddField(t.Format(time.RFC1123Z))
```
sourishkrout, line 161: maybe just `t.Format(time.DateTime)` here? on the client it's okay to show "local time"

sourishkrout (changes requested):
Can we "remove" the session injection for now? Don't need it for the demo and it needs more thought.

internal/cmd/server.go
pastuxso: They were added to be available later as environment variables when a new session is created.

pastuxso: https://github.com/stateful/runme/pull/532/files#diff-50b92da422aa4c0c738d7a625064364d14c03cafd8b1048eefd6fb9ae10227d9R96

pastuxso: NVM, Now I understand that variables are being injected in the extension.

image

pastuxso: Removing, you're right.

pastuxso:
> Can we "remove" the session injection for now? Don't need it for the demo and it needs more thought.

Yes, removed.

sourishkrout (approved):
👍 Works. Added some flags.

Likely needs a refactor to use the Runner client interface for non-hidden release which we can do later.

Issue #533, opened 2024-03-15
Experimental `runme env store check`


sourishkrout:
The idea would be to call the `rpc CreateSession(CreateSessionRequest) returns (CreateSessionResponse)` and if it does not error exit with "success" and exit code zero. Otherwise, return the error message handed down by the API.

If, for an instance, an Spec was misspelled, e.g. `Pllain!` and unknown, the API should complain. See screenshot below which is just surfacing the error returned by the gRPC API.

![Image](https://github.com/stateful/runme/assets/250527/bfdb4be9-6819-46ef-acc5-8112b3fde489)

PR #537, opened 2024-03-17
Experimental Command `runme env store check`

pastuxso:
Adds an experimental command to check the smart store, it creates a new session, if the session creation fails, it prints the founded errors.

Issue: #533


sourishkrout (approved):
👍 LGTM for the demo

PR #544, opened 2024-03-28
Fix handling non-existing dir in LoadOrGenerateConfigFromDir

adambabik:


sourishkrout (approved):
👍 LGTM. Fixes the issue.

Great to have a test now, too.

PR #548, opened 2024-04-07
Implement command execution in Docker

adambabik:
This PR introduces a `DockerCommand` which allows to execute code blocks in Docker. It's equivalent to `NativeCommand` and `VirtualCommand`.

This PR contains also a refactor of the command package. The highlight is to introduce `command.Command` interface as a generalization of all types of commands. It requires introducing `command.Kernel` too, which is an abstraction over the underlying OS.

`DockerCommand` is functional, but there might be some rough edges still.

All changes are limited to the `beta` commands.


sourishkrout:
@adambabik is `$ TAGS="test_with_docker" make test` how I would best run this locally?

adambabik:
@sourishkrout exactly. The CI pipeline should be configured in that way in this PR.

sourishkrout (approved):
👍 LGTM

Issue #543, opened 2024-03-27
Replicate Environment Variable Resolution/Prompting like the current integration with the Extension


pastuxso:
Currently, there is redundancy in our approach to environment variable resolution, evading kernel APIs such as ResolveProgram and the Owl Store. This duplication of logic interferes with consistency between the CLI and the Extension.

- [x] Refactor client_local.go and client_remote.go to call the ResolveProgram
- [x] Adds support to handle the values of the cell attribute `promptEnv`
- [ ] Update tests

PR #550, opened 2024-04-08
Replicate Environment Variable Resolution/Prompting like the current integration with the Extension

pastuxso:
TODO: Desecription

Closes: #543

### DIFF
--- a/internal/cmd/common.go
+++ b/internal/cmd/common.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strconv"
 	"strings"
 
@@ -14,6 +13,7 @@ import (
 	"github.com/spf13/cobra"
 	"go.uber.org/zap"
 
+	"github.com/stateful/runme/v3/internal/command"
 	"github.com/stateful/runme/v3/internal/document"
 	"github.com/stateful/runme/v3/internal/document/identity"
 	runnerv1 "github.com/stateful/runme/v3/internal/gen/proto/go/runme/runner/v1"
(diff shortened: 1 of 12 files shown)


internal/cmd/run.go
```diff
@@ -203,6 +199,8 @@ func runCmd() *cobra.Command {
 					return err
 				}
 
+				defer localRunner.Stop()
```
pastuxso, line 202: It's required to stop the Ad-Hoc gRPC server

internal/cmd/common.go
```diff
@@ -350,63 +350,56 @@ func promptEnvVars(cmd *cobra.Command, envs []string, tasks ...project.Task) err
 	for _, task := range tasks {
 		block := task.CodeBlock
 
-		if block.PromptEnv() {
-			varPrompts := getCommandExportExtractMatches(block.Lines())
-			for _, ev := range varPrompts {
-				if slices.Contains(keys, ev.Key) {
-					block.SetLine(ev.LineNumber, "")
+		if !block.PromptEnv() {
```
pastuxso, line 353: Refactor in favor of an early return instead of a nested block

internal/runner/client/client_local.go
```diff
@@ -218,6 +171,43 @@ func shellID() (int, bool) {
 	return i, true
 }
 
+func (r *LocalRunner) setupAdHocServer() (
```
sourishkrout, line 174: We want to extend the existing [Runner interface](https://github.com/stateful/runme/blob/main/internal/runner/client/client.go#L55-L66) that separates Local and Remote implementations, not make the Local implementation Remote.

sourishkrout (changes requested):
See in-code comment. We also don't want to redo session mgmt for the clients either.

PR #553, opened 2024-04-12
Refactor of experimental config aka config+

sourishkrout:
Stopped short of aligning the internal implementations naming/shape yet until we get from `alpha` to something like `beta`.

The idea was to separate configuration concerns by `Kernel` (anything that's core and settings that are shared between clients; still need to pull them in from primarily the extension) and `Project` which describes how to handle the project inside repo the config resides in.

What I still need to add is `Document` aka Frontmatter which is either meant to be set globally at root or allows overwrites via nesting per directory (e.g. `cwd: ../..`).

Post PR early to allow work in progress feedback.


sourishkrout:
Obsolete due to https://github.com/stateful/runme/pull/584.

Issue #552, opened 2024-04-09
Interpreter detection does not work with Python venv


sourishkrout:
The problem is that the lookup is performed on the system ENV, not the runner's current session which accounts for changes to vars such as `PATH`.

Here's the problem:

### Requirements

- Ensure you have python installed
- Create a virtual env
- Install the docker and prettytable packages

```sh {"id":"01HV1QY5V8NFJ1FQQ5578KV1MX","terminalRows":"20"}
python3 -m venv .venv
source .venv/bin/activate
pip3 install docker prettytable

# Ran on 2024-04-09 11:37:38-04:00 for 1.921s exited with 0
Requirement already satisfied: docker in ./.venv/lib/python3.12/site-packages (7.0.0)
Requirement already satisfied: prettytable in ./.venv/lib/python3.12/site-packages (3.10.0)
Requirement already satisfied: packaging>=14.0 in ./.venv/lib/python3.12/site-packages (from docker) (24.0)
Requirement already satisfied: requests>=2.26.0 in ./.venv/lib/python3.12/site-packages (from docker) (2.31.0)
Requirement already satisfied: urllib3>=1.26.0 in ./.venv/lib/python3.12/site-packages (from docker) (2.2.1)
Requirement already satisfied: wcwidth in ./.venv/lib/python3.12/site-packages (from prettytable) (0.2.13)
Requirement already satisfied: charset-normalizer<4,>=2 in ./.venv/lib/python3.12/site-packages (from requests>=2.26.0->docker) (3.3.2)
Requirement already satisfied: idna<4,>=2.5 in ./.venv/lib/python3.12/site-packages (from requests>=2.26.0->docker) (3.6)
Requirement already satisfied: certifi>=2017.4.17 in ./.venv/lib/python3.12/site-packages (from requests>=2.26.0->docker) (2024.2.2)
```

## Bug

We've got a bug here where the interpreter detection does not use the Runme session's PATH.

```py {"id":"01HV1QY5V8NFJ1FQQ5593EASDA"}
import docker
from prettytable import PrettyTable

def list_running_containers():
    client = docker.from_env()
    containers = client.containers.list()

    if containers:
        table = PrettyTable(["Container ID", "Name", "Image", "Status"])
        for container in containers:
            table.add_row([container.id[:12], container.name, container.attrs['Config']['Image'], container.status])
        print("Running containers:")
        print(table)
    else:
        print("No running containers found.")

if __name__ == "__main__":
    list_running_containers()


[...]

PR #555, opened 2024-04-15
Use PATH from session to look for programs

adambabik:
If `PATH` is present in the session, use it to look for the program paths. This is especially important for situation when within a session a virtual env is created and consecutive code blocks should be executed within it.

Fixes https://github.com/stateful/runme/issues/552

Port this change to https://github.com/stateful/runme/pull/548.

### DIFF
--- a/internal/runner/session.go
+++ b/internal/runner/session.go
@@ -3,6 +3,7 @@ package runner
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 
 	lru "github.com/hashicorp/golang-lru/v2"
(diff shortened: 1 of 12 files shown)


internal/runner/session.go
```diff
@@ -309,6 +314,11 @@ func (es *owlEnvStorer) addEnvs(envs []string) error {
 	return nil
 }
 
+func (es *owlEnvStorer) getEnv(name string) string {
+	// TODO(adamb): implement this
```
adambabik, line 318: @sourishkrout I will need some help here. Is the owl store designed in a way that it's possible to get a single env value by name?

sourishkrout: Not yet but I can add it later. I'd suggest to use a snapshot for now and leave a todo for me. Something roughly like below should work:

```go
	snapshot, err := es.owlStore.Snapshot()
	for _, item := range snapshot {
		if item.Var.Name == "PATH" {
			...
			break
		}
	}
```

sourishkrout (approved):
👍 nice work!

PR #556, opened 2024-04-15
Use pointers internally for frontmatter

sourishkrout:
Changes to futher relax the requirement that Session Outputs always require document-level lifecycle identity.

1. The lifecycle identity (frontmatter) is a deserialization option. The best way for serialization is to check if lifecycle identity is present.
2. Reduce dependency of the test suite to particular version number. Makes it easier to run when build/test flags are set differently.
3. When JSON frontmatter was being used, the `---` trailer didn't start with a new line. Fixed.

### DIFF
--- a/internal/project/project.go
+++ b/internal/project/project.go
@@ -492,7 +492,7 @@ func getCodeBlocks(data []byte) (document.CodeBlocks, error) {
 	identityResolver := identity.NewResolver(identity.DefaultLifecycleIdentity)
 	d := document.New(data, identityResolver)
 
-	if f, err := d.Frontmatter(); err == nil && f != nil && !f.Runme.IsEmpty() && f.Runme.Session != nil && f.Runme.Session.ID != "" {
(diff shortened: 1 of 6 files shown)


sourishkrout:
@pastuxso when you have a moment, please.

internal/project/project.go
```diff
@@ -492,7 +492,7 @@ func getCodeBlocks(data []byte) (document.CodeBlocks, error) {
 	identityResolver := identity.NewResolver(identity.DefaultLifecycleIdentity)
 	d := document.New(data, identityResolver)
 
-	if f, err := d.Frontmatter(); err == nil && f != nil && f.Runme.Session.ID != "" {
+	if f, err := d.Frontmatter(); err == nil && f != nil && !f.Runme.IsEmpty() && f.Runme.Session != nil && f.Runme.Session.ID != "" {
```
pastuxso, line 495: What do you think about encapsulating this condition  `!f.Runme.IsEmpty() && f.Runme.Session != nil && f.Runme.Session.ID != ""` into a method `func (m *RunmeMetadata) IsFooBarBaz()`?

PR #558, opened 2024-04-18
Fix intermediate state where newly gen frontmatter wasn't saved yet

sourishkrout:
Also simplifies logic when document-level lifecycle identity (i.e. doc ID) is turned off.

### DIFF
--- a/pkg/document/document_test.go
+++ b/pkg/document/document_test.go
@@ -88,9 +88,10 @@ First paragraph
 
 		frontmatter, err := doc.Frontmatter()
 		require.NoError(t, err)
-		marshaledFrontmatter, err := frontmatter.Marshal(testIdentityResolver.DocumentEnabled())
+		marshaledFrontmatter, docID, err := frontmatter.Marshal(testIdentityResolver.DocumentEnabled())
 		require.NoError(t, err)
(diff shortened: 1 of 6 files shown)


internal/document/frontmatter.go
```diff
@@ -87,19 +87,21 @@ func newFrontmatter() *Frontmatter {
 	}
 }
 
-// Marshal returns a marshaled frontmatter including triple-dashed lines.
-// If the identity is required, but Frontmatter is nil, a new one is created.
-func (f *Frontmatter) Marshal(requireIdentity bool) ([]byte, error) {
+// Marshal returns a marshaled frontmatter including triple-dashed lines and a document ID.
+// If the identity is required, but Frontmatter is nil, a new one including ID is created.
+func (f *Frontmatter) Marshal(requireIdentity bool) ([]byte, string, error) {
```
adambabik, line 92: Remark: In the ideal world we would strive for following signature: `Marshal() ([]byte, error)`.

adambabik (approved):

internal/document/frontmatter.go
sourishkrout: I know, but it would require me to rethink/rewrite a large part of the prepopulating frontmatter scenario which makes other things super complicated.

sourishkrout: Still need to do more integration testing with the extension before committing to this change, fwiw.

sourishkrout:
Obsolete as of https://github.com/stateful/runme/pull/677.

PR #560, opened 2024-04-19
Unlock shebang++

sourishkrout:
Multiple things:

1. Allow cells to be store in ENV as long as their name conforms to an opinionated layout (we can give visual feedback in the notebook UI).
2. If a language is not executable and no interpreter/program is set, run `cat` to stash the input cell in the output.

This will enable use cases where e.g. you have a SQL query you're live-editing in one cell and running it with `$ bq query` (BigQuery CLI) in another cell. Enables better interplay between cells than just running them back-to-back (`$__`).

@adambabik wondering if we should deny overwriting variables already contained in the ENV? However, any `export` statement will let you do that 🤷. Also, I don't want to overload the Runner with too much "cell/block"-level terminology so I introduced `known_id` & `known_name` which are as below. The idea being that if a name or id is known, clients can specify them here. Otherwise they are "unknown" since they are not required to *run* programs. Open to suggestion to model this differently.

```
` ` `sql {"id":"01HT37XG04CWS4CQS2WS7P1MX2","name":"experiments_sql"}
```

PS: Also renamed `StoreLastOutput` to `StoreEnvVars` only in `runnerv2` as a cleanup.


internal/command/config_path_normalizer.go
```diff
@@ -52,6 +52,14 @@ func pathNormalizer(cfg *Config) (*Config, func() error, error) {
 		}
 	}
 
+	// default to "cat"
+	cat, err := exec.LookPath("cat")
```
sourishkrout, line 56: @adambabik should this be using the new `sys` abstraction?

sourishkrout:
Also just realized that runnerv2's command packages does not seem to have any `TEMP_FILE` test cases, no @adambabik?

internal/api/runme/runner/v1/runner.proto
```diff
@@ -177,6 +177,12 @@ message ExecuteRequest {
 
   // file extension associated with script
   string file_extension = 26;
+
+  // optional well known id for cell/block
+  string known_id = 27;
```
adambabik, line 182: It seems that it is not used. We can keep it for the completeness reason.

The most important for me is to provide more detailed description. What does "known" mean? What guarantees does it give? Maybe given an example how client would use this.

Finally, if you take a look at `config/v1alpha1/config.proto`, we have a way to validate field values in the proto definition. This is a bit tricky and not used anywhere else, so it might be a better idea for a separate PR.

adambabik:
> Also just realized that runnerv2's command packages does not seem to have any `TEMP_FILE` test cases, no @adambabik?

This is implemented in `internal/command` which is used in runnerv2. Check out `command_args_normalizer.go` which handles `runnerv2alpha1.CommandMode_COMMAND_MODE_FILE`. It is also tested (I checked code coverage) but implicitly.

internal/api/runme/runner/v1/runner.proto
sourishkrout: Will add an example. The reason why I added `id` (unused) was because I have near-term projects in mind where I need it. I'll be sure to make the description better.

PR #561, opened 2024-04-22
Adds support to invoke ResolveProgram for runme run

pastuxso:
Currently, there is redundancy in our approach to environment variable resolution, evading kernel APIs such as ResolveProgram and the Owl Store. This duplication of logic interferes with consistency between the CLI and the Extension.

- [x] Refactor client_local.go and client_remote.go to call the ResolveProgram
- [x] Adds support to handle the values of the cell attribute `promptEnv`
- [x] Update tests

Closes: #543

### DIFF
--- a/examples/envs/README.md
+++ b/examples/envs/README.md
@@ -5,21 +5,25 @@ runme:
 ---
 
 ```sh {"id":"01HW167A4SWNJ0AQZGNZ1Z38XZ","name":"vars","promptEnv":"auto"}
-export VAR_NAME='Placeholder 1'
-export VAR_NAME2='Placeholder 2'
+export VAR_NAME1='Placeholder 1'
+export VAR_NAME2="Placeholder 2"
 export VAR_NAME3=""
+export VAR_NAME4=Message
 
-echo "1. ${VAR_NAME}"
+echo "1. ${VAR_NAME1}"
 echo "2. ${VAR_NAME2}"
 echo "3. ${VAR_NAME3}"
+echo "4. ${VAR_NAME4}"
 ```
 
 ```sh {"id":"01HW1B541A9P8BVBJZJ4E1XV1F","name":"vars2","promptEnv":"no"}
-export VAR2_NAME='Placeholder 3'
-export VAR2_NAME2='Placeholder 4'
+export VAR_NAME5='Placeholder 5'
+export VAR_NAME6='Placeholder 6'
 
-echo "1. ${VAR_NAME}"
+echo "1. ${VAR_NAME1}"
 echo "2. ${VAR_NAME2}"
-echo "3. ${VAR2_NAME}"
-echo "4. ${VAR2_NAME2}"
+echo "3. ${VAR_NAME3}"
+echo "4. ${VAR_NAME4}"
+echo "5. ${VAR_NAME5}"
+echo "6. ${VAR_NAME6}"
 ```
--- a/examples/envs/README.md
+++ b/examples/envs/README.md
@@ -1,0 +1,23 @@
+---
+runme:
+  id: 01HW1677G0SDVNNYEDJS3PBKS4
+  version: v3
+---
+
+```sh {"id":"01HW167A4SWNJ0AQZGNZ1Z38XZ","name":"vars"}
+export VAR_NAME="Placeholder 1"
+export VAR_NAME2="Placeholder 2"
+
+echo "1. ${VAR_NAME}"
+echo "2. ${VAR_NAME2}"
+```
+
+```sh {"id":"01HW1B541A9P8BVBJZJ4E1XV1F","name":"vars2"}
+export VAR2_NAME="Placeholder 3"
+export VAR2_NAME2="Placeholder 4"
+
+echo "1. ${VAR_NAME}"
+echo "2. ${VAR_NAME2}"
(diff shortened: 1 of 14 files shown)


examples/envs/README.md
```diff
@@ -0,0 +1,23 @@
+---
+runme:
+  id: 01HW1677G0SDVNNYEDJS3PBKS4
+  version: v3
+---
+
+```sh {"id":"01HW167A4SWNJ0AQZGNZ1Z38XZ","name":"vars"}
+export VAR_NAME="Placeholder 1"
```
sourishkrout, line 8: fyi, this is being treated like a `MESSAGE` not a `PLACEHOLDER` right now. I assume that's because the prompt mode is not accounted for yet (code further below). Just calling it out.

- Message uses no quotes
- Placeholder uses single or double quotes

internal/cmd/common.go
```diff
@@ -336,90 +334,81 @@ type runFunc func(context.Context) error
...
-		for _, match := range test.FindAllStringSubmatch(line, -1) {
-			e := match[0]
-
-			parts := strings.SplitN(strings.TrimSpace(e)[len("export "):], "=", 2)
-			if len(parts) == 0 {
-				continue
-			}
-			key := parts[0]
-			ph := strings.TrimSpace(parts[1])
+func prompter(cmd *cobra.Command, variable *runnerv1.ResolveProgramResponse_VarResult) (string, error) {
```
sourishkrout, line 385: More descriptive, i.e. longer, function name. The `cmd` package's namespace is pretty wide. Take inspiration from other function names.

internal/api/runme/runner/v1/runner.proto
```diff
@@ -362,6 +362,8 @@ service RunnerService {
   rpc GetSession(GetSessionRequest) returns (GetSessionResponse) {}
   rpc ListSessions(ListSessionsRequest) returns (ListSessionsResponse) {}
   rpc DeleteSession(DeleteSessionRequest) returns (DeleteSessionResponse) {}
+  // TODO(cepeda): Proposal
+  // rpc SyncSession(SyncSessionRequest) returns (SyncSessionResponse) {}
```
sourishkrout, line 366: Not sure why this is proposed/required, pinged you out of band to discuss.

internal/runner/client/client_local.go
```diff
@@ -220,6 +284,30 @@ func shellID() (int, bool) {
 	return i, true
 }
 
+// TODO(cepeda): Session is not working
```
sourishkrout, line 287: fyi, one of the big difference between local and remote runner impls is that session are ad-hoc versus "persistent". the only exception is the `tui`. however, that's a long-running process.

internal/runner/client/client_local.go
```diff
@@ -58,17 +61,31 @@ func NewLocalRunner(opts ...RunnerOption) (*LocalRunner, error) {
...
+	ctx := context.Background()
+	err = r.setupSession(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return r, nil
+}
+
+func (r *LocalRunner) setupSession(ctx context.Context) error {
```
sourishkrout, line 79: Consider removing `ctx` since it's not being used

internal/runner/client/client_local.go
```diff
@@ -91,9 +108,9 @@ func (r *LocalRunner) newExecutable(task project.Task) (runner.Executable, error
 		Tty:     block.InteractiveLegacy(),
 		Stdout:  r.stdout,
 		Stderr:  r.stderr,
-		Session: r.session,
 		System:  system.Default,
 		Logger:  r.logger,
+		Session: r.session,
```
sourishkrout, line 113: Not sure why this "had" to move down a few lines - not wrong but no change should be no diff?

sourishkrout (changes requested):
@pastuxso Overall looking great 👍; few comments in line. The one scenario that's not covered is having "shell prompts", aka `$` inside of scripts (see diff below). See below. Currently normalization for this is done on the client. Been meaning to move this on the API-side, however, don't want to do it haphazardly since it requires broad and focused testing. This is the relevant call: https://github.com/stateful/vscode-runme/blob/7c88cff0787111243e45bdc0b1f8ac5d67b1d795/src/extension/executors/runner/index.ts#L516-L517

```
diff --git a/examples/envs/README.md b/examples/envs/README.md
index d3ebcc1..5a06039 100644
--- a/examples/envs/README.md
+++ b/examples/envs/README.md
@@ -5,10 +5,10 @@ runme:
 ---
 
 ```sh {"id":"01HW167A4SWNJ0AQZGNZ1Z38XZ","name":"vars","promptEnv":"auto"}
-export VAR_NAME1='Placeholder 1'
-export VAR_NAME2="Placeholder 2"
-export VAR_NAME3=""
-export VAR_NAME4=Message
+$ export VAR_NAME1='Placeholder 1'
+$ export VAR_NAME2="Placeholder 2"
+$ export VAR_NAME3=""
+$ export VAR_NAME4=Message
 
 echo "1. ${VAR_NAME1}"
 echo "2. ${VAR_NAME2}"
```

sourishkrout (approved):
LGTM 👍. Nice work, @pastuxso.

Issue #551, opened 2024-04-09
Missing piece in Runme terminal prototype: rolling up environment variables


sourishkrout:
While not fully baked this VS Code terminal integration with already works pretty well. We likely need a new [CommandMode](https://github.com/stateful/runme/blob/main/internal/api/runme/runner/v1/runner.proto#L74-L78) that runs ENV snapshotting before the process quits. However, perhaps there are alternatives?

Here's the simple integration; https://github.com/stateful/vscode-runme/blob/main/src/extension/kernel.ts#L826-L831

PR #563, opened 2024-04-23
Implement TerminalCommand and Terminal command mode

adambabik:
This change implements a new mode called `TERMINAL`. In this mode, the server, right after starting the script, writes initial commands that are responsible for collecting environment variables, which will be added or deleted to the current session.

The `TERMINAL` mode is supported only in the `v2alpha1` runner.

Fixes https://github.com/stateful/runme/issues/551


sourishkrout:
@adambabik probably makes sense to disable that the failing tests on Windows?

adambabik:
@sourishkrout ah yes, sorry I missed that. Tests using `pty` do not work on Windows. I rebased on main and force pushed.

sourishkrout (approved):
LGTM 👍. Another good reason to move to `runnerv2`.

PR #564, opened 2024-05-01
internal/config: replace viper with custom config loader

adambabik:
This PR implements a generic config loader. The main purpose is to find all relevant `runme.yaml` starting from a Markdown file location. The algorithm goes up the directory tree until the project root. The next step is to merge all found `runme.yaml` configs into a single one that will be used by the execution layer.


sourishkrout (approved):
👍 LGTM

PR #573, opened 2024-05-14
Define AI GRPC service to support using AI to generate cells

jlewi:
We'd like to use AI to generate cells. In particular, we'd like to support using [Foyle](foyle.io) to add AI capabilities.

The initial plan is to use a gRPC service to allow RunMe to communicate with Foyle (or potentially other AI services). This proto needs to define that gRPC service.


jlewi:
@sourishkrout PTAL

sourishkrout:
@jlewi main line is merged. you'll see that now everything API lives under `pkg/api`. the definitions specifically under `pkg/api/proto` and generated code under `pkg/api/gen`.

sourishkrout:
> The PR currently defines a new buf module for the AI service. I'm not sure if that's required or advisable.

I'd suggest to just make it part of Runme's buf module for now. We can work on untangling once we have a working "prototype".

jlewi:
@sourishkrout Thanks for the quick refactor. I've updated the PR to move the ai protos into the pkg/api directory.

jlewi:
The SONAR failure seems unrelated to my changes. The error is
`Set the SONAR_TOKEN env variable`.

sourishkrout (approved):
👍 LGTM

sourishkrout:
> The SONAR failure seems unrelated to my changes. The error is `Set the SONAR_TOKEN env variable`.

Something with repo secrets permissions since your repos is a fork. Not to worried about this. Sonar is a way for us to monitor code health from a mile high view.

PR #575, opened 2024-05-14
Move `api` & `gen` from internal under pkg

sourishkrout:
This refactor will allow linking against Runme's proto API definitions.

Once merged let's reconcile with https://github.com/stateful/runme/pull/573.


sourishkrout:
Merging. Received sign off ✅  out of band on Discord.

Issue #566, opened 2024-05-06
Interactive inputs do not produce "echo"


sourishkrout:
Simple example is running `read`. Unclear what needs to toggle whether or not echoing is enabled. Might have to do some digging in Runnerv1.

```
config:{program_name:"/bin/zsh" directory:"/Users/sourishkrout/Projects/stateful/oss/vscode-runme/examples" env:"RUNME_ID=01HX7QEZTB94V7A28K766WNRG3" env:"TERM=xterm-256color" commands:{items:"read" items:""} interactive:true mode:COMMAND_MODE_INLINE known_id:"01HX7QEZTB94V7A28K766WNRG3"} winsize:{rows:10 cols:177} session_id:"01HX7QGJ4VNB9YYEC3NCHD908M" store_stdout_in_env:true
```

https://github.com/stateful/runme/blob/e163a3a9a20284457c8c518b753b62257e66a800/internal/command/command_virtual.go#L98-L100

adambabik:
It's disabled by default: https://github.com/stateful/runme/blob/e163a3a9a20284457c8c518b753b62257e66a800/internal/command/command_virtual.go#L98-L100

Should it be enabled or configurable?

sourishkrout:
> Should it be enabled or configurable?

I wish I could remember the logic because it wasn't just always on or always off. Do you see anything in the runnerv1 implementation that's revelatory?

adambabik:
> I wish I could remember the logic because it wasn't just always on or always off. Do you see anything in the runnerv1 implementation that's revelatory?

In the `runnerv1` service, echo is always enabled. When using the CLI, executing a program which is a file or shell, and the block is marked as interactive, which is a default, then echo is disabled. I am not sure if it was well thought out, but the fact is that in `runnerv2` echo is always disabled. Should I flip it to stay compatible?

> @adambabik when you search for `DisableEcho` it appears that the echo is enabled when a TTY is attached which is always true for a VirtualCommand, no?

This is true and `VirtualCommand` is used when the block is interactive, which is a default.

sourishkrout:
@sourishkrout yes, let's go ahead and flip it please. Being "incompatible" otherwise means the integration with the notebook terminal stops working as expected.

adambabik:
@sourishkrout I merged #576. Let me know if there is anything else that should be done as a part of this ticket.

sourishkrout:
> @sourishkrout I merged #576. Let me know if there is anything else that should be done as a part of this ticket.

[...]

PR #576, opened 2024-05-14
Refactor internal/command by removing normalizers

adambabik:
Remove the concept of normalizers from `internal/command`. Introduce hierarchical commands and `command.Factory`, instead.

This PR makes it more explicit how various command types are created. 

Before there were really two types of commands, namely native and virtual. The latter was using pty under the hood. Whether a particular program was executed inline or through a script file was decided in normalizers that modified `command.ProgramConfig` (known `command.Config` earlier).

With this change, each command type is a separate struct. The commands are hierarchical, i.e. there is a `command.base` and higher-level commands which embed `command.internalCommand`.

All commands creation is managed by `command.Factory` so a caller does not need to worry about which command type to create. The caller will typically write the code like this:
```go
factory := NewFactory(cfg, kernel, logger)
cmd := factory.Build(
	&ProgramConfig{
		ProgramName: "echo",
		Arguments:   []string{"-n", "test"},
		Mode:        runnerv2alpha1.CommandMode_COMMAND_MODE_INLINE,
	}
	Options{
		Session:     session,
		Stdout:       stdout
	},
)
```


sourishkrout (approved):
👍 LGTM, just one typo in comments.

sourishkrout:
One issue I'm noticing is that `disableEcho` is enabled for the notebook-integrated terminal in the extension which makes any input invisible.

Relevant comment: https://github.com/stateful/runme/issues/566#issuecomment-2118354869

adambabik:
> One issue I'm noticing is that `disableEcho` is enabled for the notebook-integrated terminal in the extension which makes any input invisible.
> 
> Relevant comment: [#566 (comment)](https://github.com/stateful/runme/issues/566#issuecomment-2118354869)

This should be addressed in https://github.com/stateful/runme/pull/576/commits/d1d7e2f976bdd4be8f8b281f09e1ce54ce1d1ed6.

Issue #567, opened 2024-05-07
Null pointer panic when a session is being deleted


sourishkrout:
Not a difficult fix but I'll let you take a look first, @adambabik since you refactored the v1 code.

```
2024-05-07T09:19:50.613-0400	INFO	runnerv2service/service_sessions.go:95	running DeleteSession in runnerService
panic: runtime error: index out of range [-1]

goroutine 11 [running]:
github.com/stateful/runme/v3/internal/command.(*SessionList).Delete(0x1400000fcb0, {0x140007c6500, 0x1a})
	/Users/sourishkrout/Projects/stateful/oss/runme/internal/command/session.go:105 +0x210
github.com/stateful/runme/v3/internal/runnerv2service.(*runnerService).DeleteSession(0x14000517be0, {0x10258f820, 0x140009c43c0}, 0x140008a44c0)
	/Users/sourishkrout/Projects/stateful/oss/runme/internal/runnerv2service/service_sessions.go:97 +0x90
github.com/stateful/runme/v3/internal/gen/proto/go/runme/runner/v2alpha1._RunnerService_DeleteSession_Handler({0x1024dd9e0, 0x14000517be0}, {0x10258f820, 0x140009c43c0}, 0x14000b9d800, 0x0)
	/Users/sourishkrout/Projects/stateful/oss/runme/internal/gen/proto/go/runme/runner/v2alpha1/runner_grpc.pb.go:289 +0x2a4
google.golang.org/grpc.(*Server).processUnaryRPC(0x140003f7000, {0x10258f820, 0x140009c4300}, {0x102598058, 0x14000028000}, 0x14000b9b200, 0x14000285620, 0x10303d240, 0x0)
	/Users/sourishkrout/go/pkg/mod/google.golang.org/grpc@v1.62.1/server.go:1386 +0x1398
google.golang.org/grpc.(*Server).handleStream(0x140003f7000, {0x102598058, 0x14000028000}, 0x14000b9b200)
	/Users/sourishkrout/go/pkg/mod/google.golang.org/grpc@v1.62.1/server.go:1797 +0xd00
google.golang.org/grpc.(*Server).serveStreams.func2.1()
	/Users/sourishkrout/go/pkg/mod/google.golang.org/grpc@v1.62.1/server.go:1027 +0x144
created by google.golang.org/grpc.(*Server).serveStreams.func2 in goroutine 40
[...]

PR #579, opened 2024-05-17
Fix handling sessions in runnerv2 service

adambabik:
Closes https://github.com/stateful/runme/issues/567


sourishkrout (approved):
Works 👍

PR #582, opened 2024-05-21
Remove kernel from runme.yaml and add explicit docker support

adambabik:
This change removes complicated kernels and introduces an explicit support for Docker:

```yaml
docker:
  enabled: true
  image: runme-runtime:latest
  build:
    context: ./experimental/docker
    dockerfile: Dockerfile
```

Running code blocks in Docker is still not fully implemented.


sourishkrout (approved):
LGTM 👍

Issue #574, opened 2024-05-14
Support using Foyle's AI capabilities as a plugin


jlewi:
Per the discussion in discord; we'd like to experiment with integrating RunMe with [Foyle's AI](https://foyle.io/) capabilities. Foyle aims to be a DevOps copilot. One of the key problems its tackling is trying to train an AI copilot to be an expert in your infrastructure. Foyle relies on a notebook like experience to collect implicit human feedback which is used to train the AI ([blog post](https://foyle.io/docs/blog/learning/)). Like RunMe, Foyle is also built using VSCode Notebooks. RunMe is far more mature and has more features than Foyle. So it likely makes sense to reuse RunMe as Foyle's notebook rather than rebuilding a separate stack.

Towards that end, this issue is tracking the minimal implementation to get RunMe to support calling out to Foyle and surfacing the results. The current thinking is they would be two separate extensions to start with and RunMe would use gRPC to communicate with Foyle. A more detailed design is provided in [Foyle Tech Note 004](https://github.com/jlewi/foyle/blob/main/tech_notes/tn004_runme.md).

 The first step is designing a shared gRPC service to allow the services to communicate; #573 

- [x] Create a minimal runme code to generate completions using foyle (#1356)
- [x] Ensure Runme logs data necessary for retraining Foyle
- [ ] Related to jlewi/foyle#110

jlewi:
Here's a more detailed look at what we need to do for logging

# Gaps In Current RunMe Logging

Here are the current limitations of logging in RunMe for the purposes of supporting Foyle retraining

* Server logging isn't always enabled when using VSCode
* When server logs are enabled, server logs aren't persisted to files; they are echoed to an output terminal in VSCode
   * This means they are ephemeral and aren't easily accessible by Foyle for retraining
* Server Logs don't record the actual executed command at info level
  * request is logged at debug level [here](https://github.com/stateful/runme/blob/3d637c0238b8bd6b465118181efdbadd054fe3b3/internal/runner/service.go#L219)
  * However, I don't think it would be logged using the JSON serialization format of the proto which would make deserializing it difficult
    * Per [StackOverflow](https://stackoverflow.com/questions/68411821/correctly-log-protobuf-messages-as-unescaped-json-with-zap-logger)
      it won't be seralized as the proto JSON format

# Proposed Changes to RunMe Code Base

## Server Changes

* Server should setup a second logger to always log to a JSON file 
[...]

PR #585, opened 2024-05-23
Create machine logs to support AI retraining

jlewi:
To support collecting implicit feedback and retraining an AI we need to create machine readable logs that contain the actual command executed.

The design is described in #574.

With this change and ai logs enabled here is what the logs look like.

On the console a sample log message looks like the following

1.716425780157274e+09	info	runner/service.go:233	received initial request	{"_id": "01HYHF5TXWTPV0ZGSQP9WKMZP9", "req": "{\"commands\":[\"ls\"]}"}

In the JSON log we get a log message like the following

{"severity":"info","time":1716425780.157274,"caller":"runner/service.go:233","function":"github.com/stateful/runme/v3/internal/runner.(*runnerService).Execute","message":"received initial request","_id":"01HYHF5TXWTPV0ZGSQP9WKMZP9","req":"{\"commands\":[\"ls\"]}"}

Related to #574

### DIFF
--- a/internal/cmd/server.go
+++ b/internal/cmd/server.go
@@ -35,6 +35,7 @@ func serverCmd() *cobra.Command {
 		devMode      bool
 		enableRunner bool
 		tlsDir       string
+		enableAILogs bool
 	)
 
 	cmd := cobra.Command{
@@ -56,7 +57,7 @@ The kernel is used to run long running processes like shells and interacting wit
 			}
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			logger, err := getLogger(devMode)
+			logger, err := getLogger(devMode, enableAILogs)
 			if err != nil {
 				return err
 			}
@@ -133,8 +134,9 @@ The kernel is used to run long running processes like shells and interacting wit
 	cmd.Flags().StringVarP(&addr, "address", "a", defaultAddr, "Address to create unix (unix:///path/to/socket) or IP socket (localhost:7890)")
 	cmd.Flags().BoolVar(&devMode, "dev", false, "Enable development mode")
 	cmd.Flags().BoolVar(&enableRunner, "runner", true, "Enable runner service (legacy, defaults to true)")
+	cmd.Flags().BoolVar(&enableAILogs, "ai-logs", false, "Enable logs to support training an AI")
 	cmd.Flags().StringVar(&tlsDir, "tls", defaultTLSDir, "Directory in which to generate TLS certificates & use for all incoming and outgoing messages")
-
+	cmd.Flags().StringVar(&configDir, configDirF, GetDefaultConfigHome(), "If ai logs is enabled logs will be written to ${config-dir}/logs")
 	_ = cmd.Flags().MarkHidden("runner")
 
 	return &cmd
--- a/internal/runner/service.go
+++ b/internal/runner/service.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"google.golang.org/protobuf/encoding/protojson"
 	"io"
 	"os"
 	"regexp"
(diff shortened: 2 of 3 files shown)


jlewi:
@sourishkrout This is ready for you.

internal/cmd/server.go
```diff
@@ -133,8 +134,9 @@ The kernel is used to run long running processes like shells and interacting wit
 	cmd.Flags().StringVarP(&addr, "address", "a", defaultAddr, "Address to create unix (unix:///path/to/socket) or IP socket (localhost:7890)")
 	cmd.Flags().BoolVar(&devMode, "dev", false, "Enable development mode")
 	cmd.Flags().BoolVar(&enableRunner, "runner", true, "Enable runner service (legacy, defaults to true)")
+	cmd.Flags().BoolVar(&enableAILogs, "ai-logs", false, "Enable logs to support training an AI")
 	cmd.Flags().StringVar(&tlsDir, "tls", defaultTLSDir, "Directory in which to generate TLS certificates & use for all incoming and outgoing messages")
-
+	cmd.Flags().StringVar(&configDir, configDirF, GetDefaultConfigHome(), "If ai logs is enabled logs will be written to ${config-dir}/logs")
```
sourishkrout, line 139: Any reason not to reuse `fLogFilePath` as base? The config home is a bit weird on `darwin` (and frankly, tentative). Also, as part of the `beta` commands (incomplete alpha) we are migrating config away from a central location to a `runme.yaml` that resides inside a project/repository.

internal/runner/service.go
```diff
@@ -216,7 +218,20 @@ func (r *runnerService) Execute(srv runnerv1.RunnerService_ExecuteServer) error
 		return errors.WithStack(err)
 	}
 
-	logger.Debug("received initial request", zap.Any("req", req))
+	reqJSON, err := protojson.Marshal(req)
+	if err != nil {
+		logger.Error("failed to marshal request", zap.Error(err))
+	}
+
+	// We want to always log the request because it is used for AI training.
```
sourishkrout, line 226: We should try to add a tests for his in https://github.com/jlewi/runme/blob/jlewi/logs/internal/runner/service_test.go - the reason being is that I'm currently in the middle of charting a migration from `runnerv1` (this impl) to `runnerv2` (alpha; future) and the tests serve as a spec to not have things fall thru the cracks.

`runnerv2` lives over here https://github.com/jlewi/runme/tree/jlewi/logs/internal/runnerv2service but instead of syncing the service impl, I'd rather have a comprehensive test suite in `runnerv1` as guard rails for filling in all the gaps.

I'm happy to take a stab at adding a test case if you need a template.

sourishkrout (changes requested):
This makes sense 👍; I've added some comments I'd like to address before merging.

internal/cmd/server.go
jlewi: So I think we can break this down into separate questions

1. What's the default location for logs? e.g. is it ${CONFIGDIR}/logs?
2. What flags control the location of the logs?

Regarding the first question.

* I don't think we want the default value to be a temporary directory because the logs are too valuable to be lost when the temporary directory gets cleaned up

> we are migrating config away from a central location to a runme.yaml that resides inside a project/repository

Are you suggesting the logs get stored relative to the project/repo specific runme.yaml? Or is this just an indication that the runme user configuration is going away in favor of per project/repository configuration.

Either way; I think of log location as being configured per user not per project/repository. Which implies we need some per user directory.

> The config home is a bit weird on darwin

What about ${HOME}/.runme for  runme user configuration and ${HOME}/.runme/logs

Two; I think once we agree on where logs should be stored we can decide if it makes sense to change fLogFilePath to use it and if so use it to compute the location of the AI Logs.

internal/runner/service.go
jlewi: Let me take a look at this.

jlewi: @sourishkrout added a test. The test passes locally when I run it individually and the tests appear to be green in CI. However, as noted in #587 I'm seeing test failures when running locally

internal/cmd/server.go
sourishkrout: I see your point here, especially that it's a `user` vs a `server` log. That really answers both questions 1. & 2. for me. Here's my suggestion, we roll with your code change as is, however, we rename `GetDefaultConfigHome` to `GetUserConfigHome` to make the difference in base directories more apparent and future proof @jlewi?

Ideally `GetDefaultConfigHome` / `GetUserConfigHome` would actually be rooted in the home directory (on all platforms) but we can fix that at a later time. I'll create a bug ticket for it.

jlewi: Done.

jlewi:
@sourishkrout PTAL; I did the rename of the logging directory.

internal/runner/service.go
sourishkrout: Apologies for the #587 expedition. After rerunning the C/I job it looks like it's indeed the flaking issue we are working on.

sourishkrout (approved):
LGTM ✅

PR #588, opened 2024-05-24
move document package from internal to pkg

jlewi:
The purpose of this PR is to make it possible to reuse runme's Markdown -> Notebook code.
In particular I'd like to use it to parse ChatGPT output into a notebook so that code blocks can be rendered in RunMe notebooks. 

In Foyle, there is currently a bug with parsing out code blocks nested inside lists jlewi/foyle#114. Runme's code properly handles this use case. This will allow us to reuse RunMe's code to fix this problem.


jlewi:
Looks like the test failure was a flake because it passed at [8c1e584](https://github.com/stateful/runme/pull/588/commits/8c1e584ab105d94211e64833cf06c227fb1341d8)

PR #591, opened 2024-05-25
Disable SONAR scan in PRs created from forks

### LINKED ISSUE #587: UnitTest failures on main
Running on main I'm seeing test failures.

```
git log
commit 61fc8785ab8adb4cec5f1135d278a9e8895f327c (HEAD, upstream/main, upstream/HEAD)
Author: Sebastian Tiedtke <sebastiantiedtke@gmail.com>
Date:   Thu May 23 16:41:17 2024 -0400

    Is python3 more likely than python? (#586)
    
    `python` is an alias for `python3` on my BMP which makes the test fail.
```

Here are the test failures

```
 logger.go:146: 2024-05-23T23:09:03.531Z	INFO	stream canceled after the process finished; ignoring	{"id": "01HYKVE2NV87J6ZQTMY69E6BYT"}
    --- FAIL: TestRunnerServiceServerExecute_WithInput/SimulateCtrlC (1.52s)
        service_execute_test.go:643: 
            	Error Trace:	/Users/jlewi/git_runme/internal/runnerv2service/service_execute_test.go:643
            	Error:      	"rpc error: code = Unknown desc = exit status 1" does not contain "exit status 130"
            	Test:       	TestRunnerServiceServerExecute_WithInput/SimulateCtrlC
        service_execute_test.go:644: 
            	Error Trace:	/Users/jlewi/git_runme/internal/runnerv2service/service_execute_test.go:644
            	Error:      	Not equal: 
            	            	expected: 130
            	            	actual  : 1
            	Test:       	TestRunnerServiceServerExecute_WithInput/SimulateCtrlC
```

```
    --- FAIL: TestRunnerServiceServerExecute_WithInput/ContinuousInput (0.04s)
        service_execute_test.go:600: 
            	Error Trace:	/Users/jlewi/git_runme/internal/runnerv2service/service_execute_test.go:600
            	Error:      	Not equal: 
            	            	expected: "a\r\nb\r\nc\r\nd\r\nA\r\nB\r\nC\r\nD\r\n"
            	            	actual  : "a\r\nb\r\nA\r\nB\r\nc\r\nC\r\nd\r\nD\r\n"
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -2,7 +2,7 @@
            	            	 b
            	            	-c
            	            	-d
            	            	 A
            	            	 B
            	            	+c
            	            	 C
            	            	+d
            	            	 D
            	Test:       	TestRunnerServiceServerExecute_WithInput/ContinuousInput
```

```
2024-05-23T23:09:03.740Z	DEBUG	runner/command.go:404	finished copying from pty to stdout	{"count": 32}
--- FAIL: Test_command (0.00s)
    --- FAIL: Test_command/JavaScript (0.03s)
        command_test.go:107: 
            	Error Trace:	/Users/jlewi/git_runme/internal/runner/command_test.go:107
            	Error:      	Not equal: 
            	            	expected: ""
            	            	actual  : "(node:24760) ExperimentalWarning: The Node.js specifier resolution flag is experimental. It could change or be removed at any time.\n(Use `node --trace-warnings ...` to show where the warning was created)\n"
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1 +1,3 @@
            	            	+(node:24760) ExperimentalWarning: The Node.js specifier resolution flag is experimental. It could change or be removed at any time.
            	            	+(Use `node --trace-warnings ...` to show where the warning was created)
            	            	 
            	Test:       	Test_command/JavaScript
    --- FAIL: Test_command/JavaScriptEnv (0.03s)
        command_test.go:135: 
            	Error Trace:	/Users/jlewi/git_runme/internal/runner/command_test.go:135
            	Error:      	Not equal: 
            	            	expected: ""
            	            	actual  : "(node:24765) ExperimentalWarning: The Node.js specifier resolution flag is experimental. It could change or be removed at any time.\n(Use `node --trace-warnings ...` to show where the warning was created)\n"
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1 +1,3 @@
            	            	+(node:24765) ExperimentalWarning: The Node.js specifier resolution flag is experimental. It could change or be removed at any time.
            	            	+(Use `node --trace-warnings ...` to show where the warning was created)
            	            	 
            	Test:       	Test_command/JavaScriptEnv
FAIL
FAIL	github.com/stateful/runme/v3/internal/runner	4.624s
ok  	github.com/stateful/runme/v3/internal/runner/client	0.255s
--- FAIL: TestRunnerServiceServerExecute_WithInput (1.61s)
```

Tests appear to have passed on main at this commit #586 .

Are others seeing test failures when running locally or is it just me?

jlewi:
* Disable the SONAR scan in PRs created from forks. It won't be able to run because the GHA can't get the secrets for PRs created from a fork.
  * The failure creates noise in various dashboards because we see the red "x" by the PR.

~~* A lot of unittests aren't running on CI because the GHA currently ~~

~~  * Runs all tests but only on windows and a bunch of tests have the build tag !windows~~
~~  * On docker ubuntu only the tests with the dag docker ubuntu run~~

~~* The fix is to add a GHA step that runs `make test` to run on all OSS~~
~~  * This rule doesn't use any build tags so it will run all tests.~~
~~Related to #587~~

* Mark some locally flaky tests to be skipped

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,12 +57,6 @@ jobs:
         run: |
           go build -o runme main.go
           ./runme --version
-      - name: Test All
-        run: |
-          export SHELL=/bin/bash
-          export TZ=UTC
-          make test
-        if: ${{ matrix.os == 'ubuntu-latest' }}
       - name: Test
         run: |
           export SHELL=/bin/bash
@@ -70,10 +64,7 @@ jobs:
           TAGS="test_with_docker" make test/coverage
           make test/coverage/func
         if: ${{ matrix.os == 'ubuntu-latest' }}
-      # make test should run on all os's because
-      # on windows only a subset of the tests run.
       - name: Test
-        # N.B. Set-Timezone is a windows command
         run: |
           Set-Timezone -Id "UTC" -PassThru
           make test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,7 @@ jobs:
         run: |
           export SHELL=/bin/bash
           export TZ=UTC
-          make test          
+          make test
         if: ${{ matrix.os == 'ubuntu-latest' }}
       - name: Test
         run: |
@@ -69,15 +69,15 @@ jobs:
           export TZ=UTC
           TAGS="test_with_docker" make test/coverage
           make test/coverage/func
-        if: ${{ matrix.os == 'ubuntu-latest' }}      
+        if: ${{ matrix.os == 'ubuntu-latest' }}
       # make test should run on all os's because
       # on windows only a subset of the tests run.
       - name: Test
-        # N.B. Set-Timezone is a windows command 
+        # N.B. Set-Timezone is a windows command
         run: |
           Set-Timezone -Id "UTC" -PassThru
           make test
-        if: ${{ matrix.os == 'windows-latest' }}        
+        if: ${{ matrix.os == 'windows-latest' }}
       - name: Upload coverage
         uses: actions/upload-artifact@v4
         if: ${{github.actor != 'dependabot[bot]' && matrix.os == 'ubuntu-latest'}}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,6 +134,10 @@ jobs:
 
       - name: SonarCloud Scan
         uses: SonarSource/sonarcloud-github-action@v2.1.1
+        # Skip this test if the PR is created from a fork.
+        # If its created from a fork the PR won't be able to fetch the secrets so
+        # the step will fail.
+        if: github.event.pull_request.head.repo.full_name == github.repository
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # Needed to get PR information, if any
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,11 +64,12 @@ jobs:
           TAGS="test_with_docker" make test/coverage
           make test/coverage/func
         if: ${{ matrix.os == 'ubuntu-latest' }}
+      # make test should run on all os's because
+      # on windows only a subset of the tests run.
       - name: Test
         run: |
           Set-Timezone -Id "UTC" -PassThru
-          make test
-        if: ${{ matrix.os == 'windows-latest' }}
+          make test        
       - name: Upload coverage
         uses: actions/upload-artifact@v4
         if: ${{github.actor != 'dependabot[bot]' && matrix.os == 'ubuntu-latest'}}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,19 +57,27 @@ jobs:
         run: |
           go build -o runme main.go
           ./runme --version
+      - name: Test All
+        run: |
+          export SHELL=/bin/bash
+          export TZ=UTC
+          make test          
+        if: ${{ matrix.os == 'ubuntu-latest' }}
       - name: Test
         run: |
           export SHELL=/bin/bash
           export TZ=UTC
           TAGS="test_with_docker" make test/coverage
           make test/coverage/func
-        if: ${{ matrix.os == 'ubuntu-latest' }}
+        if: ${{ matrix.os == 'ubuntu-latest' }}      
       # make test should run on all os's because
       # on windows only a subset of the tests run.
       - name: Test
+        # N.B. Set-Timezone is a windows command 
         run: |
           Set-Timezone -Id "UTC" -PassThru
-          make test        
+          make test
+        if: ${{ matrix.os == 'windows-latest' }}        
       - name: Upload coverage
         uses: actions/upload-artifact@v4
         if: ${{github.actor != 'dependabot[bot]' && matrix.os == 'ubuntu-latest'}}
--- a/internal/runner/service_test.go
+++ b/internal/runner/service_test.go
@@ -347,7 +347,7 @@ func Test_runnerService(t *testing.T) {
 	})
 
 	t.Run("Input", func(t *testing.T) {
-		if os.Getenv("RUN_FLAKY") == "" {
+		if skip, err := strconv.ParseBool(os.Getenv("SKIP_FLAKY")); err == nil && skip {
 			t.Skip("skipping flaky test")
 		}
 		t.Parallel()
--- a/internal/runner/service_test.go
+++ b/internal/runner/service_test.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
+	"strconv"
 	"strings"
 	"syscall"
 	"testing"
--- a/internal/runner/service_test.go
+++ b/internal/runner/service_test.go
@@ -347,6 +347,9 @@ func Test_runnerService(t *testing.T) {
 	})
 
 	t.Run("Input", func(t *testing.T) {
+		if os.Getenv("RUN_FLAKY") == "" {
+			t.Skip("skipping flaky test")
+		}
 		t.Parallel()
 
 		stream, err := client.Execute(context.Background())


adambabik:
The first bullet points are not true. There is this step:

```
      - name: Test
        run: |
          export SHELL=/bin/bash
          export TZ=UTC
          TAGS="test_with_docker" make test/coverage
          make test/coverage/func
```

It runs all tests thanks to `TAGS="test_with_docker" make test/coverage`, including generating a coverage report.

Not all tests can run on Windows because they are Linux specific. At the current state, runme does not provide full Windows support. Windows with WSL is recommended.

With regard to the Sonar support and PRs from forks, let's wait for @sourishkrout's comment.

jlewi:
@adambabik Thank you. You're correct and setting TAGS=tags_with_docker should run all tests. In which case, do you know what's going on with respect to service_test.go is running CI?
You can see from my previous run of test all
https://github.com/stateful/runme/actions/runs/9231990310/job/25402625003

That service_test failed in the test_all rule but not on main.
Based on the coverage report (https://github.com/stateful/runme/actions/runs/9229617830/job/25396164592) it does seem like service_test is running on maining. 

It looks like test all sets `-race=false` whereas the test with docker rule doesn't.

sourishkrout:
> With regard to the Sonar support and PRs from forks, let's wait for @sourishkrout's comment.

Makes sense to disable Sonar for forks since we don't want to get into secrets sharing.

adambabik:
> In which case, do you know what's going on with respect to service_test.go is running CI?

It is run on Linux but it's omitted on Windows.

> You can see from my previous run of test all https://github.com/stateful/runme/actions/runs/9231990310/job/25402625003

This test is simply flaky. It is run in main and on PRs but it sometimes fails due to race condition. runnerv2 is more stable in this respect.

.github/workflows/ci.yml
```diff
@@ -57,14 +57,23 @@ jobs:
         run: |
           go build -o runme main.go
           ./runme --version
+      - name: Test All
```
adambabik, line 60: This is duplication of what's below (`name: Test`) but without coverage. It runs exactly the same set of tests.

We can argue if it should be `TAGS="test_with_docker" make test-coverage` (notice instead of `-` which is a different Makefile target) in the step below to bring it closer to `make test`.

.github/workflows/ci.yml
```diff
@@ -57,14 +57,23 @@ jobs:
...
         run: |
           export SHELL=/bin/bash
           export TZ=UTC
           TAGS="test_with_docker" make test/coverage
           make test/coverage/func
         if: ${{ matrix.os == 'ubuntu-latest' }}
+      # make test should run on all os's because
+      # on windows only a subset of the tests run.
       - name: Test
+        # N.B. Set-Timezone is a windows command
```
adambabik, line 76: I don't understand this comment. It is set before calling `make test`: `Set-Timezone -Id "UTC" -PassThru`.

internal/runner/service_test.go
```diff
@@ -347,6 +347,9 @@ func Test_runnerService(t *testing.T) {
 	})
 
 	t.Run("Input", func(t *testing.T) {
+		if os.Getenv("RUN_FLAKY") == "" {
```
adambabik, line 350: This is not the way to go. We should make such tests more robust instead of marking them as flaky. Otherwise, they will never be fixed. I am not sure if this test is any more flaky than other ones.

adambabik (changes requested):

adambabik (changes requested):
My recommendation to move forward with this PR is as follows:
- Change `TAGS="test_with_docker" make test/coverage` to `TAGS="test_with_docker" make test-coverage` to make it closer to `make test` on Windows.
- Keep the change related to skipping SonarCloud scan.
- Remove other changes.

internal/runner/service_test.go
sourishkrout: > This is not the way to go. We should make such tests more robust instead of marking them as flaky. Otherwise, they will never be fixed. I am not sure if this test is any more flaky than other ones.

@adambabik I don't disagree with you on the principle but it's beyond Jeremy's involvement to fix tests in the existing test suite beyond a reasonable effort (which he's already making). I think it's fair marking them so we can continue to target them while it removes ambiguity whether test failures are connected or disconnect to unrelated changes.

sourishkrout:
> * Remove other changes.

Please see my other comment.

internal/runner/service_test.go
adambabik: I noticed that now all tests passed, so that single test isn't more flaky than others. But it's a good point that if any test is blocking a merge then this approach seems good, or commenting it out entirely plus creating an issue is also fine.

Unless this is also a problem locally then I would be happy to discuss it on Discord in details, @jlewi.

.github/workflows/ci.yml
sourishkrout: I believe it's just a note to make clear that `Set-Timezone` is a Windows PowerShell command; for folks who are not familiar with Windows details. Please confirm @jlewi.

jlewi: That's right. Happy to remove it if you prefer less comments.

jlewi:
Thanks @adambabik and @sourishkrout and sorry for the slow response.
So I'm unsure what to do with this PR because. The original motivation for this PR was 

* I observed locally that several tests were flaky
* On CI the tests were passing consistently on main.

This lead me to conclude that the problem was that some tests were not running under CI. This conclusion was based on a misunderstanding about how tags worked. I assumed that setting tag "test_with_docker" only ran those tests as opposed to all tests + tests with that tag.

As @adambabik points out that's incorrect. So `TestRunnerServiceServerExecute_WithInput/SimulateCtrlC` is running and presumably passing fairly reliably on CI. I tried again locally on main and its failing for me locally.

```
    --- FAIL: TestRunnerServiceServerExecute_WithInput/SimulateCtrlC (1.52s)
        service_execute_test.go:724: 
            	Error Trace:	/Users/jlewi/git_runme/internal/runnerv2service/service_execute_test.go:724
            	Error:      	"rpc error: code = Unknown desc = exit status 1" does not contain "exit status 130"
            	Test:       	TestRunnerServiceServerExecute_WithInput/SimulateCtrlC
        service_execute_test.go:725: 
            	Error Trace:	/Users/jlewi/git_runme/internal/runnerv2service/service_execute_test.go:725
            	Error:      	Not equal: 
            	            	expected: 130
            	            	actual  : 1
            	Test:       	TestRunnerServiceServerExecute_WithInput/SimulateCtrlC
```

So at this point; I don't know why CI is more reliable than local. As a result, I don't know how to fix GHA to make it not mask flakiness. However, this means all my CI changes (except for the sonar one) are solving a non-existent problem so I reverted them.

I'm uncertain what to do about skipping SimulateCtrlC. So I'm good with whatever you want. Skipping the test made more sense if we could reliably reproduce the failure in CI and didn't want to block PRs while we fixed it. But since the failures aren't reliably reproducing in CI the rationale doesn't really hold.

I've left the Sonar change in because I think that makes sense given reasons mentioned previously.

Closing this PR is also a completely reasonable outcome.

adambabik:
@jlewi thanks for the detailed explanation! There could be several reasons why this test fails for you locally, such as bash version, OS (including different Linux distros), terminal settings, and so on. 

One of the options we could introduce is to run tests locally in a Docker container. Then we could ensure more reliable behaviour for all contributors, without them having to guess what to install in which version or delaying feedback and waiting for the CI result. However, the fact is that the nature of runme and its ability to execute commands on the host OS will keep  it open for edge cases.

internal/runner/service_test.go
adambabik: @jlewi we can introduce this, but I suggest changing it slightly:

```suggestion
		if skip, err := strconv.ParseBool(os.Getenv("SKIP_FLAKY")); err == nil && skip {
```

It will be more explicit what's the intention: if the `SKIP_FLAKY` env is present and set to [truthy value](https://pkg.go.dev/strconv#ParseBool) then skip this flaky test.

adambabik (commented):
One suggestion, otherwise LGTM. Thanks!

jlewi:
Thanks @adambabik ; I had no idea that function was part of the stdlib. I've applied the suggestion.

sourishkrout (approved):
👍 LGTM

sourishkrout:
@jlewi are the pre commit hooks running for you? Wondering if our contribution guidelines are incomplete

jlewi:
@sourishkrout That was my mistake. I just accepted the suggestion through the GitHub UI and didn't check back to make sure the CI was passing. I'd have to double check that precommits are running on stateful/runme; they run on stateful/runme-vscode.

Issue #571, opened 2024-05-10
Port path lookup from runnerv1 to take session PATH into account


sourishkrout:
Basically to mirror https://github.com/stateful/runme/pull/555.

PR #594, opened 2024-05-27
Use PATH from session in runnerv2 and beta commands

adambabik:
This PR additionally contains some refactoring in the `command` and `system` packages.

Closes https://github.com/stateful/runme/issues/571


internal/runner/service_test.go
```diff
@@ -525,6 +525,56 @@ func Test_runnerService(t *testing.T) {
 		)
 	})
 
+	t.Run("ExecuteWithPathEnvInSession", func(t *testing.T) {
```
adambabik, line 528: This test was missing in the original PR about using `PATH` from the session.

sourishkrout (approved):
LGTM 👍

Issue #580, opened 2024-05-19
Document-level options "category"


ruzickap:
Hello 

The new feature introduced in https://github.com/stateful/runme/issues/369 - 'Document-level options "category"' doesn't work:

Example - let's have the following file:

* `install-manual-solution-2.md`
<pre>
---
shell: bash
cwd: /tmp
category: solution-2
---

Installation steps for Solution 2:

```sh
echo "[1] Install command in $PWD"
```

```sh
echo "[2] Install second command"
```

```sh { "category":"delete-solution" }
echo "Delete solution"
```

```sh { "category":"delete-solution" }
echo "Delete solution - second command"
```
</pre>

* Issue 1 - If I run `runme run --all --skip-prompts --category solution-2 --dry-run` I got only first command in markdown document "processed":

  ```
  #!/opt/homebrew/bin/bash

  // run in "/tmp"

  set -e -o pipefail
  echo "[1] Install command in $PWD"
  ```

  -> Second command `[2] Install second command` is skipped / ignored ???

* Issue 2 - The `runme run --all --skip-prompts --category delete-solution --dry-run` is not working at all and `delete-solution` category is not "recognized":
  ```
  could not execute command: No tasks to execute with the category provided
  ```

It seems like the feature is buggy.

Thank you for looking at it...

sourishkrout:
Could you please verify/repro @pastuxso when you get a chance? Thanks.

pastuxso:
Hi @ruzickap, Thank you for reporting this issue. I'll review it and get back to you as soon as I have news.

pastuxso:
Hey @ruzickap, I've identified the cause of the issue and I'll fix it. In the meantime, you can use a workaround by adding a name to the cells you want to include in the global category and removing them from the cell with the category. Here's an example:


````md
---
shell: bash
cwd: /tmp
category: solution-2
---

Installation steps for Solution 2:

```sh {"name":"one"}
echo "[1] Install command in $PWD"
```

```sh {"name":"two"}
echo "[2] Install second command"
```

```sh {"category":"delete-solution"}
echo "Delete solution"
```

```sh {"category":"delete-solution"}
echo "Delete solution - second command"
```


````

sourishkrout:
@ruzickap apologies for the named vs unnamed oversight. We fixed the issue and added more test coverage. Thanks @pastuxso!

pastuxso:
[...]

PR #598, opened 2024-06-02
Fixes category filtering for unnamed cells

pastuxso:
* Enhances readability
* Fixes category filtering
* Adds more **textar** test cases

Closes: #580

### DIFF
--- a/internal/runner/service.go
+++ b/internal/runner/service.go
@@ -648,7 +648,7 @@ func readLoop(
 func runnerWinsizeToPty(winsize *runnerv1.Winsize) *pty.Winsize {
 	if winsize == nil {
 		// sane default
-		return &pty.Winsize{Cols: 80}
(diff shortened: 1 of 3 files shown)


internal/runner/service.go
```diff
@@ -648,7 +648,7 @@ func readLoop(
 func runnerWinsizeToPty(winsize *runnerv1.Winsize) *pty.Winsize {
 	if winsize == nil {
 		// sane default
-		return &pty.Winsize{Cols: 80}
+		return &pty.Winsize{Rows: 24, Cols: 80}
```
pastuxso, line 651: It helps to resolve flaky service tests on Linux. 🤷🏻‍♂️

sourishkrout (approved):
👍 LGTM

internal/runner/service.go
adambabik: Definitely a good call to have it set even if didn't affect the tests.

Issue #600, opened 2024-06-03
Some languages require a specific file extension (e.g. typescript + tsc)


sourishkrout:
For reasons I didn't further research the Typescript Compiler requires temp-filenames to end with `.ts*`. Otherwise it will not treat the contents as Typescript.

Example of error: https://gist.github.com/sourishkrout/4c497f62636d390ca3ad4aa7e1dac0f3#file-ts-01hzfc6yz1sc34rzfmd7xw3k99-md

In runner v1 this is being handled here:
https://github.com/stateful/runme/blob/699148d37b82344397883fdd23936c6a34d5f83d/internal/runner/command.go#L213-L215

PR #601, opened 2024-06-03
Add extension to script file name for file commands

adambabik:
Closes https://github.com/stateful/runme/issues/600

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,8 +21,8 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version: ${{ matrix.go }}
-      - name: Setup deno
       - uses: denoland/setup-deno@v1
+      - name: Setup deno
         with:
           deno-version: v1.x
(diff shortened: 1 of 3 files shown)


sourishkrout (approved):
:+1: works!

sourishkrout:
I suspect the test's failing in C/I because `ts-node` is not installed by default. Perhaps this test case is a candidate to run inside a container. Added an e2e test case on the extension side https://github.com/stateful/vscode-runme/pull/1403.

Using `deno` might get around the problem. Going to try a few things.

sourishkrout:
> I suspect the test's failing in C/I because `ts-node` is not installed by default. Perhaps this test case is a candidate to run inside a container. Added an e2e test case on the extension side [stateful/vscode-runme#1403](https://github.com/stateful/vscode-runme/pull/1403).
> 
> Using `deno` might get around the problem. Going to try a few things.

Adding `deno` as a GHA worked. Rolling with it.

PR #602, opened 2024-06-03
Simplify `commandFactory.buildVirtual`

adambabik:
In the case of `virtualCommand`, `stdin` is wrapped in `readCloser` and it should be returned to any other struct that depends on `virtualCommand`. `base.stdin` should stay unchanged.


sourishkrout (approved):
Makes sense 👍 . LGTM.

PR #607, opened 2024-06-05
Fail gracefully if remote runner connection info is unreliable

sourishkrout:
Unless any `--insecure`/`--insecure=false`/`--insecure=true` is explicitly set.

There is an edge case when VS Code resumes previously persisted terminal sessions upon loading which shows an ugly error if the server is now no longer running under the previously specified socket. In this case, it's better UX to gracefully fall back to a local runner.

PS: Stopped short of addressing gRPC upstream deprecations. Would like to tackle them as part of moving to `runme beta` cmds.

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -22,6 +22,7 @@ var (
 	fProject               string
 	fProjectIgnorePatterns []string
 	fRespectGitignore      bool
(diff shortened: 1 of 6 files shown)


pastuxso (approved):
LGTM!

PR #609, opened 2024-06-06
Run `shellscript` as such

sourishkrout:
In the extension when the `languageID` set to `shellscript`, it's being rewritten to `sh`. However, only after serialization following the initial deserialization. To deal with this edge-case let's be sure we run `shellscript` as such. Prior to this fix it would default to `cat` (shebang++) which is wrong.

PS: Interestingly this is already fixed in `v2alpha1` and the code in this PR is exclusively for `v1`.


adambabik:
> PS: Interestingly this is already fixed in `v2alpha1` and the code in this PR is exclusively for `v1`.

It's because `shellscript` was added to the `interpreterByLanguageID` map. I think it's still worth to add a test in `internal/command/command_file_test.go`:

```diff
--- a/internal/command/command_file_test.go
+++ b/internal/command/command_file_test.go
@@ -27,6 +27,20 @@ func TestFileCommand(t *testing.T) {
                testExecuteCommand(t, cfg, nil, "test", "")
        })
 
+       t.Run("Shellscript", func(t *testing.T) {
+               t.Parallel()
+
+               cfg := &ProgramConfig{
+                       LanguageId: "shellscript",
+                       Source: &runnerv2alpha1.ProgramConfig_Script{
+                               Script: `echo "run this as shell script"`,
+                       },
+                       Mode: runnerv2alpha1.CommandMode_COMMAND_MODE_FILE,
+               }
+
+               testExecuteCommand(t, cfg, nil, "run this as shell scripta\n", "")
+       })
+
        t.Run("Python", func(t *testing.T) {
                t.Parallel()
```

sourishkrout:
Done in https://github.com/stateful/runme/pull/609/commits/945dd14a51b4d010f81cb71d40a72444425f53b3 @adambabik.

PR #617, opened 2024-06-14
Implement fifo based shell env collector

adambabik:
Replace the shell env collector from a file-based to named pipe-based (fifo). For Windows, it will be a named pipe, but it will be done in a follow-up PR. Also, the env dump regardless of the destination is encrypted.

The encryption is mostly to avoid leaking secrets by accidents. Our threat model assumes execution of runme by the same user without any additional isolation.

Enablement of this feature is controlled by two vars:
```go
var (
	envCollectorEnableEncryption = true
	envCollectorUseFifo          = true
)
```


adambabik:
@sourishkrout as this is a big change to a crucial component, I would recommend a three-step release:
1. Release with disabled encryption and fifo. There should be no breaking changes.
2. Enable fifo.
3. Enable encryption.

sourishkrout:
> @sourishkrout as this is a big change to a crucial component, I would recommend a three-step release:
> 
> 1. Release with disabled encryption and fifo. There should be no breaking changes.
> 2. Enable fifo.
> 3. Enable encryption.

A release in stages makes sense. These changes are not affecting runner v1, right? If so, we can skip on step ahead and as long as beta commands v2 in the extension is still opt-in, we can test-drive `fifo` ahead of turning on encryption.

sourishkrout (approved):
✅ LGTM. Let's merge and then figure out the release strategy.

adambabik:
> A release in stages makes sense. These changes are not affecting runner v1, right? If so, we can skip on step ahead and as long as beta commands v2 in the extension is still opt-in, we can test-drive `fifo` ahead of turning on encryption.

It's a good point. It does not affect v1.

PR #618, opened 2024-06-24
Delete `make docker` target

pastuxso:
During a setup on a new machine, I noticed the command `make docker` was not working due to a path issue.

@adambabik I was wondering if it's still needed. is it?.

Trying to solve this issue on Linux:

image

### DIFF
--- a/internal/project/teststub/teststub.go
+++ b/internal/project/teststub/teststub.go
@@ -15,6 +15,7 @@ func Setup(t *testing.T, temp string) Path {
 
 	testDataSrc := originalTestDataPath()
 	require.NoError(t, copy.Copy(testDataSrc, temp))
+	Cleanup(t, temp)
 
 	err := os.Rename(
 		filepath.Join(temp, "git-project", ".git.bkp"),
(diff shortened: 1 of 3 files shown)


sourishkrout:
@adambabik can you take this review please?

adambabik (approved):
Good catch! I don't see `make docker` being used anywhere. GoReleaser uses the image but not the Makefile. It's here for the sake of completeness, I guess. Deleting the target sounds like a good option too.

pastuxso:
Let's remove the "make docker" instead. I Agree with you @adambabik.

internal/project/teststub/teststub.go
```diff
@@ -15,6 +15,7 @@ func Setup(t *testing.T, temp string) Path {
 
 	testDataSrc := originalTestDataPath()
 	require.NoError(t, copy.Copy(testDataSrc, temp))
+	Cleanup(t, temp)
```
adambabik, line 18: @pastuxso did you encounter any problem that led to creating this helper? Overall, when using `T.TempDir()` it is cleared automatically when the test case finishes.

pastuxso: @adambabik When I run the tests on Linux, it's failing, even individually performed.

This is an example:

image

PR #621, opened 2024-06-27
Increase max msg size to 100 MiB

sourishkrout:
Sessions Output documents can get quite large. 4 MiB isn't enough.


adambabik (approved):
I noticed in other work that tests `Test_runnerService/*` can be flaky, unfortunately...

Issue #624, opened 2024-07-02
"bufio.Scanner: token too long" after large binary stream


sourishkrout:
Works in runnerv1 but errors in runnerv2. Running this notebook's cells sequentially throws an error. I suspect it's related to saving the output into `$__` since the error only happens if the cell with curl'd image ran right before.

https://gist.github.com/sourishkrout/d75c3a782374d64ffcb1c07b395a8aa5

```
2024-07-02T15:53:14.013-0700	INFO	InlineShellCommand	command/command_inline_shell.go:48	failed to collect the environment	{"id": "01J1TTDBBY3TNMJBYD3VTGDASD", "instanceID": "01J1TTDBC2KM20HKBC3D3FAPHN", "error": "failed to scan the env file \".env_start\": bufio.Scanner: token too long", "errorVerbose": "bufio.Scanner: token too long\nfailed to scan the env file \".env_start\""}
```

PR #626, opened 2024-07-04
Fix scanner max buffer size in env collector

adambabik:
The solution is to provide a custom buffer to the scanner in the env collector. Effectively, it is a port from runnerv1 with a better initial buffer size.

Closes https://github.com/stateful/runme/issues/624


sourishkrout (approved):
✅ LGTM

PR #627, opened 2024-07-05
Fix 1455 and retain invalid frontmatter over failing/overwriting

sourishkrout:
Turns out people use templating languages and interpolation in frontmatter which makes it invalid unless rendered first. The idea of this PR is to pass-thru frontmatter when attempts to parsing it fail.

Fixes https://github.com/stateful/vscode-runme/issues/1455

### DIFF
--- a/pkg/document/document.go
+++ b/pkg/document/document.go
@@ -116,6 +116,10 @@ func (d *Document) splitAndParse() error {
 	return nil
 }
 
+func (d *Document) FrontMatterRaw() []byte {
+	return d.frontmatterRaw
+}
+
 // splitSource splits source into FrontMatter and content.
 // TODO(adamb): replace it with an extension to goldmark.
 // Example: https://github.com/abhinav/goldmark-frontmatter
--- a/pkg/document/document.go
+++ b/pkg/document/document.go
@@ -116,7 +116,7 @@ func (d *Document) splitAndParse() error {
 	return nil
 }
 
-func (d *Document) FrontMatterRaw() []byte {
+func (d *Document) FrontmatterRaw() []byte {
 	return d.frontmatterRaw
 }
 
--- a/pkg/document/editor/editor.go
+++ b/pkg/document/editor/editor.go
@@ -2,11 +2,11 @@ package editor
 
 import (
 	"bytes"
-	"fmt"
 	"strconv"
 	"time"
 
 	"github.com/pkg/errors"
+	"go.uber.org/zap"
 
 	"github.com/stateful/runme/v3/pkg/document"
 	"github.com/stateful/runme/v3/pkg/document/constants"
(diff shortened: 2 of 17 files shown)


sourishkrout:
Lemme know what you think @adambabik. Making checking for errors optional seemed like the best way forward to solve for the PR's description/linked issue.

pkg/document/editor/editor.go
```diff
@@ -17,17 +18,19 @@ const (
 	DocumentID     = "id"
 )
 
-func Deserialize(data []byte, identityResolver *identity.IdentityResolver) (*Notebook, error) {
+func Deserialize(log *zap.Logger, data []byte, identityResolver *identity.IdentityResolver) (*Notebook, error) {
```
adambabik, line 21: Suggestion: as this interface will likely get extended, I suggest this:

```suggestion
type Options struct {
    IdentityResolver *identity.IdentityResolver
    Logger *zap.Logger
}

func Deserialize(data []byte, opts Options) (*Notebook, error) {
```

pkg/document/document.go
```diff
@@ -116,6 +116,10 @@ func (d *Document) splitAndParse() error {
 	return nil
 }
 
+func (d *Document) FrontMatterRaw() []byte {
```
adambabik, line 119: Suggestion: Let's stick with `Frontmatter` for consistency.

```suggestion
func (d *Document) FrontmatterRaw() []byte {
```

pkg/document/frontmatter.go
```diff
@@ -182,20 +182,24 @@ func (f *Frontmatter) ensureID() {
...
-func (d *Document) Frontmatter() (*Frontmatter, error) {
+func (d *Document) Frontmatter() *Frontmatter {
 	d.splitSource()
 
+	d.parseFrontmatter()
+
+	return d.frontmatter
+}
+
+func (d *Document) FrontmatterError() error {
```
adambabik, line 193: Suggestion: We can have two methods: `Frontmatter() *Frontmatter` and `FrontmatterWithError() (*Frontmatter, error)`. I think it will be a bit more clear and the caller will only need to use one method when it wants to handle frontmatter's error.

```suggestion
func (d *Document) FrontmatterWithError() (*Frontmatter, error) {
```

pkg/project/formatter.go
```diff
@@ -30,7 +33,7 @@ func Format(files []string, basePath string, flatten bool, formatJSON bool, writ
 		identityResolver := identity.NewResolver(identity.DefaultLifecycleIdentity)
 
 		if flatten {
-			notebook, err := editor.Deserialize(data, identityResolver)
+			notebook, err := editor.Deserialize(logger, data, identityResolver)
```
adambabik, line 36: Remark: With my earlier suggestion, this will become:
```suggestion
			notebook, err := editor.Deserialize(data, editor.Options{IdentityResolver: identityResolver})
```

The logger can be figure out by `editor.Deserialize` and default to noop.

adambabik (commented):
I think the solution makes sense. I added a few comments on how the public APIs can be improved.

sourishkrout:
@adambabik feedback makes perfect sense 👌. Thank you.

PR #629, opened 2024-07-06
We don't want two Project APIs

sourishkrout:
Internal tools (cross-repo) depend on this which why it was originally moved under `pkg`.

If anybody else depends on the old Project APIs, they can file a ticket since we don't make general stability guarantees (yet).


sourishkrout:
@adambabik, can you please take a look? We should have nuked the old Project API the first time around (irrespective that it lives under `pkg`).

PS: I'm starting to wonder if we'd be better of moving things out under `internal` since `internal` vs `pkg` sends mixed signals. If we did't have `internal` we could move a subset of what's under `pkg` right now that's truly "stable".

adambabik (approved):

Issue #636, opened 2024-07-23
Panic when reading/writing to closed pipe


sourishkrout:
This is super easy to reproduce from the extension with a non-interactive cell. Once started, you click the stop button (see video), the runme kernel server dies due to a panic with the stack:

```
io: read/write on closed pipe
github.com/stateful/runme/v3/internal/runnerv2service.(*execution).Write
	/Users/sourishkrout/Projects/stateful/oss/runme/internal/runnerv2service/execution.go:215
github.com/stateful/runme/v3/internal/runnerv2service.(*runnerService).Execute.func1
	/Users/sourishkrout/Projects/stateful/oss/runme/internal/runnerv2service/service_execute.go:90
runtime.goexit
	/opt/homebrew/Cellar/go/1.22.5/libexec/src/runtime/asm_arm64.s:1222
```

Here's a log: https://gist.github.com/sourishkrout/ba1d6fac2fcdde464ec5a0bf4cbcb5e3

https://github.com/user-attachments/assets/dc670039-a4a8-4bc9-a2cc-aa8990e8aa4a

adambabik:
@sourishkrout how does the extension implement the "stop" button? We don't have a separate call so I assume that the gRPC stream does something to the existing client->server stream.

adambabik:
From the logs it's clear that something like this happens:
1. The client sends a `ExecuteRequest` message with the `stop` field.
2. The client kills the existing client->server stream immediately.
3. Server receives the request and attempts to stop the command but the server exits.

I was able to reproduce it and it seems that the problem is with the fact that the signal is sent to the whole process group. The server process and its subprocesses (commands) are in the same process group. I will send a PR to disable sending a signal to the group. The drawback is that when the command has its own subprocesses, they may leak and become zombie.

Also, we likely should fix the communication between the client and the server. The client should await a confirmation that the stop request was handled and close the connection only after that, or after a timeout. Otherwise, some command output might be missed.

adambabik:
Consider the following solutions:
- Create a new process group for the command 
- Handle `SIGCHLD` using `SIG_IGN`
- Set the `SA_NOCLDWAIT` flag

sourishkrout:
> @sourishkrout how does the extension implement the "stop" button? We don't have a separate call so I assume that the gRPC stream does something to the existing client->server stream.

Here's how. Btw, this the same for Runner v1 vs v2 (not saying it's correct):

[...]

Issue #635, opened 2024-07-23
FiFo issue: CTRL+C runs indefinitely when signals are being caught by process


sourishkrout:
This is no longer an issue once FiFo & encryption is disabled:

https://github.com/stateful/runme/blob/main/internal/command/factory.go#L14-L17

https://github.com/user-attachments/assets/950978b6-c11a-41de-8a59-552f7cad4ba9

The problem can be observed with the [Foyle](https://foyle.io/docs/getting-started/) binary running the `foyle serve` command or `terraform apply` in both cases using SIGINT (`Ctrl+C`).

This is the error Runme throws:
```
signal: interrupt,errorVerbose:signal: interrupt
github.com/stateful/runme/v3/internal/command.(*virtualCommand).Wait
	/Users/sourishkrout/Projects/stateful/oss/runme/internal/command/command_virtual.go:153
github.com/stateful/runme/v3/internal/command.(*inlineShellCommand).Wait
	/Users/sourishkrout/Projects/stateful/oss/runme/internal/command/command_inline_shell.go:49
github.com/stateful/runme/v3/internal/runnerv2service.(*execution).Wait
	/Users/sourishkrout/Projects/stateful/oss/runme/internal/runnerv2service/execution.go:158
github.com/stateful/runme/v3/internal/runnerv2service.(*runnerService).Execute
	/Users/sourishkrout/Projects/stateful/oss/runme/internal/runnerv2service/service_execute.go:124
github.com/stateful/runme/v3/pkg/api/gen/proto/go/runme/runner/v2alpha1._RunnerService_Execute_Handler
	/Users/sourishkrout/Projects/stateful/oss/runme/pkg/api/gen/proto/go/runme/runner/v2alpha1/runner_grpc.pb.go:369
google.golang.org/grpc.(*Server).processStreamingRPC
	/Users/sourishkrout/go/pkg/mod/google.golang.org/grpc@v1.64.0/server.go:1673
google.golang.org/grpc.(*Server).handleStream
	/Users/sourishkrout/go/pkg/mod/google.golang.org/grpc@v1.64.0/server.go:1794
google.golang.org/grpc.(*Server).serveStreams.func2.1
	/Users/sourishkrout/go/pkg/mod/google.golang.org/grpc@v1.64.0/server.go:1029
runtime.goexit
	/opt/homebrew/Cellar/go/1.22.5/libexec/src/runtime/asm_arm64.s:1222
```

sourishkrout:
[...]

PR #637, opened 2024-07-24
command: fix signalling to command process group

adambabik:
This change also fixes a regression that broken running interactive commands, like runme in runme, run from the CLI in https://github.com/stateful/runme/pull/637/commits/da2c38c928edec6858b1833bcfb3e2db638d2090.

It also disables FIFO which should mitigate https://github.com/stateful/runme/issues/635 but a proper fix is needed.

Relates to https://github.com/stateful/runme/issues/636


adambabik:
@sourishkrout I was able to reproduce the issue and verify this fix using grpcurl.

adambabik:
@sourishkrout there is a genuine problem with `TestRunnerServiceServerExecute_WithStop` so getting this PR back to draft.

I dug into this and when doing CTRL-C in the terminal, you actually send a SIGINT to the whole process group ([a great thread](https://superuser.com/a/1830133) that explains it thoroughly). Now I remember why I implemented this :) However, when running a server and executing commands by it, the server and the commands share the same process group and we kill the server when we send a signal to the process group; the effect is the reported bug.

It seems like the ideal solution is to make the command its own process group, just as it is done by bash. I will try that. In fact, there is code doing that but it's not used in v2.

sourishkrout:
> I dug into this and when doing CTRL-C in the terminal, you actually send a SIGINT to the whole process group ([a great thread](https://superuser.com/a/1830133) that explains it thoroughly). Now I remember why I implemented this :) However, when running a server and executing commands by it, the server and the commands share the same process group and we kill the server when we send a signal to the process group; the effect is the reported bug.

That explains why the kernel server suddenly dies 😆.

adambabik:
@sourishkrout ready to test and review.

sourishkrout (approved):
👍 LGTM

PR #638, opened 2024-07-24
APIs to support Terminal integration

sourishkrout:
First stab at providing a session fork/rollup terminal. To make this work following provisional changes were necessary:

1. Beta subcommand to source the current state of the environment, i.e. `eval $(runme beta env source...)` - similar to how `eval $(ssh-agent)` works.
2. Subcommand hits gRPC directly, will likely want to move this to an abstraction later.
3. Had to silent some of the `autoconf` errors since it currently relies on the presence of `runme.yaml` - will walk this back as part of building out `runme beta` cmds & autoconf. However, don't want to hold up Runme Terminal progress.
4. Inserted leading spaces into the trap mechanism to skip shell history.

### DIFF
--- a/internal/cmd/beta/beta_cmd.go
+++ b/internal/cmd/beta/beta_cmd.go
@@ -1,6 +1,8 @@
 package beta
 
 import (
+	"fmt"
+
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
@@ -10,8 +12,10 @@ import (
 )
 
 type commonFlags struct {
+	silent     bool
 	categories []string
 	filename   string
+	subtle     bool
 }
 
 func BetaCmd() *cobra.Command {
@@ -27,7 +31,7 @@ All commands are experimental and not yet ready for production use.
 All commands use the runme.yaml configuration file.`,
 		Hidden: true,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-			return autoconfig.InvokeForCommand(func(cfg *config.Config) error {
+			err := autoconfig.InvokeForCommand(func(cfg *config.Config) error {
 				// Override the filename if provided.
 				if cFlags.filename != "" {
 					cfg.ProjectFilename = cFlags.filename
@@ -44,6 +48,10 @@ All commands use the runme.yaml configuration file.`,
 
 				return nil
 			})
+			if !cFlags.silent && err != nil {
+				_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "%s\n", err)
+			}
+			return nil
 		},
 	}
 
@@ -54,6 +62,8 @@ All commands use the runme.yaml configuration file.`,
 	pFlags := cmd.PersistentFlags()
 	pFlags.StringVar(&cFlags.filename, "filename", "", "Name of the Markdown file to run blocks from.")
 	pFlags.StringSliceVar(&cFlags.categories, "category", nil, "Run blocks only from listed categories.")
+	pFlags.BoolVar(&cFlags.silent, "silent", false, "Silent mode. Do not error messages.")
+	pFlags.BoolVar(&cFlags.subtle, "subtle", false, "Explicitly allow delicate operations to prevent misuse")
 
 	// Hide all persistent flags from the root command.
 	// "beta" is a completely different set of commands and
@@ -73,6 +83,7 @@ All commands use the runme.yaml configuration file.`,
 	cmd.AddCommand(printCmd(cFlags))
 	cmd.AddCommand(server.Cmd())
 	cmd.AddCommand(runCmd(cFlags))
+	cmd.AddCommand(envCmd(cFlags))
 
 	return &cmd
 }
--- a/internal/cmd/beta/beta_cmd.go
+++ b/internal/cmd/beta/beta_cmd.go
@@ -2,6 +2,7 @@ package beta
 
 import (
 	"fmt"
+	"io"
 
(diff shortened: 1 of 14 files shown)


sourishkrout:
Please review @adambabik. The `beta` sub cmds are provisional to enable the integration. We'll clean this up as part of a consolidated effort to build out the remaining CLI gap.

sourishkrout:
Also, any idea why `TestTerminalCommand_EnvPropagation` is so flakey?

internal/cmd/beta/beta_cmd.go
```diff
@@ -10,8 +13,10 @@ import (
 )
 
 type commonFlags struct {
+	silent     bool
```
adambabik, line 16: Nit: I started organizing fields alphabetically when there is no apparent ordering dictated by business logic.

internal/cmd/beta/beta_cmd.go
```diff
@@ -44,6 +53,13 @@ All commands use the runme.yaml configuration file.`,
 
 				return nil
 			})
+
+			// print the error to stderr but don't return it
```
adambabik, line 57: Nit: let's explain why so. We don't want to return it because?

internal/cmd/beta/beta_cmd.go
```diff
@@ -54,6 +70,8 @@ All commands use the runme.yaml configuration file.`,
 	pFlags := cmd.PersistentFlags()
 	pFlags.StringVar(&cFlags.filename, "filename", "", "Name of the Markdown file to run blocks from.")
 	pFlags.StringSliceVar(&cFlags.categories, "category", nil, "Run blocks only from listed categories.")
+	pFlags.BoolVar(&cFlags.silent, "silent", false, "Silent mode. Do not error messages.")
```
adambabik, line 73:
```suggestion
	pFlags.BoolVar(&cFlags.silent, "silent", false, "Silent mode. Do not print error messages.")
```

internal/cmd/beta/beta_cmd.go
```diff
@@ -54,6 +70,8 @@ All commands use the runme.yaml configuration file.`,
...
+	pFlags.BoolVar(&cFlags.subtle, "subtle", false, "Explicitly allow delicate operations to prevent misuse")
```
adambabik, line 74: Nit: it's first time I see "subtle" in such context. Typically it's more like `--unsafe`.

internal/command/command_terminal.go
```diff
@@ -43,9 +43,21 @@ func (c *terminalCommand) Start(ctx context.Context) (err error) {
 	c.logger.Info("a terminal command started")
 
 	if c.envCollector != nil {
-		return c.envCollector.SetOnShell(c.stdinWriter)
+		if err = c.envCollector.SetOnShell(c.stdinWriter); err != nil {
```
adambabik, line 46:
```suggestion
		if err := c.envCollector.SetOnShell(c.stdinWriter); err != nil {
```

internal/command/env_shell.go
```diff
@@ -7,19 +7,24 @@ import (
 
 func setOnShell(shell io.Writer, prePath, postPath string) error {
 	var err error
+
+	// Prefix commands with a space to avoid polluting the shell history.
```
adambabik, line 11: +1

adambabik (approved):
I added a few comments but no blockers.

adambabik:
> Also, any idea why `TestTerminalCommand_EnvPropagation` is so flakey?

It relies on the output of bash, but looking at the stack trace in the CI, it seems that `_, err = stdinW.Write([]byte{0x04}) // EOT` does not necessarily works as expected. Maybe you can try to replace it with `_, err = stdinW.Write([]byte("exit\n"))`. This might be more reliable than sending a signal via stdin.

On macOS it seems fine:
```
go test -count=5 -timeout 10s -run ^TestTerminalCommand_EnvPropagation$ github.com/stateful/runme/v3/internal/command
```

internal/cmd/beta/beta_cmd.go
sourishkrout: :+1:

internal/cmd/beta/beta_cmd.go
sourishkrout: Ha! I'm pretty sure I stuck with the copilot suggestion :-/.

internal/cmd/beta/beta_cmd.go
sourishkrout: I was between subtle and insecure. I think I'll use the latter since we already made use in non-beta cmds.

sourishkrout:
> On macOS it seems fine:
> 
> ```
> go test -count=5 -timeout 10s -run ^TestTerminalCommand_EnvPropagation$ github.com/stateful/runme/v3/internal/command
> ```

Yep, it runs pretty reliable on macOS but does the inverse on Linux/CI

adambabik:
> > On macOS it seems fine:
> > ```
> > go test -count=5 -timeout 10s -run ^TestTerminalCommand_EnvPropagation$ github.com/stateful/runme/v3/internal/command
> > ```
> 
> Yep, it runs pretty reliable on macOS but does the inverse on Linux/CI

I tried to replicate it in a Docker container and it seems much more stable after [this change](https://github.com/stateful/runme/blame/b73a0f5052b0c92489a2caeacd13dceeaf8b3fb9/internal/command/command_terminal_test.go#L58). Usually it's a good indicator for running in the CI. Let's keep an eye on it.

PR #640, opened 2024-07-29
Generate GQL schema

sourishkrout:
Having GQL types for the protos is required for downstream usage. Making them available for now.


sourishkrout:
When you get a chance, @pastuxso, could use see if the protos generate cleanly for you locally using `runme run proto-dev`? You may have to do a build first to pull in the latest dependencies.

sourishkrout:
Having trouble making `protoc-gen-gql` installable because it's kind of old.

pastuxso (approved):
It works fine now! Tested on Mac and Linux. Requires a new `make install/dev`. It doesn't have a name in the `CONTRIBUTING.md`. It would be fine to have it to perform something like this:

`runme run install-dev proto-dev` or add it as a dependency directly in the Makefile

LGTM 👍

PR #643, opened 2024-08-02
command: fix blocking fifo env collector

adambabik:
Fixes https://github.com/stateful/runme/issues/635

@sourishkrout I was able to reproduce it but only using the `kill` signal. Using `interrupt` was ok as the envs are properly collected by the trap in bash. It definitely depends on the behaviour of the actual program being executed, though. If you want to reproduce or test with other commands, just port `TestInlineShellCommand_CollectEnv/KillCommandWhileUsingFifo` to main.


sourishkrout (approved):
This PR does fix the error I originally reported. ✅ LGTM.

PR #644, opened 2024-08-02
Clean up s/errC/errc and rm expect pkg

adambabik:
This change makes the naming of error channels consistent and removes the obsolete `expect' package.


sourishkrout:
This is just a capitalization change, right @adambabik?

adambabik:
@sourishkrout and removal of the vendored `expect` package which is not used anymore.

sourishkrout (approved):
👍 LGTM

Issue #645, opened 2024-08-05
runme attempts to access unreadable files or directories


pastuxso:
I encountered an issue where the **runme** does not skip unreadable files or directories during processing, leading to a **permission denied** error. This behavior occurs even when the files or directories should be ignored using the **--ignore-pattern** flag.

Issue reported on [Discord](https://discord.com/channels/1102639988832735374/1102639989700968583/1269389054399479880) 

PR to replicate the issue: https://github.com/stateful/runme/pull/646 https://github.com/stateful/runme/actions/runs/10252402970/job/28362697530?pr=646#step:7:33

adambabik:
It looks like we don't handle `os.ErrPermission` https://github.com/stateful/runme/blob/0cf2e624ba20f8b6cf91eb586f75148e9b682cf2/pkg/project/project.go#L367-L370.

This type of error should be handled separately, and if the path is ignored, it should be skipped. @pastuxso would you like to tackle this?

pastuxso:
@adambabik, I could fix it. Thank you for your review. 👍🏻

PR #646, opened 2024-08-05
Fixes runme when it attempts to access unreadable files or directories

pastuxso:
Handling the os.ErrPermission when an error occurs while trying to read files without proper permissions.

Closes: #645

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -3,7 +3,6 @@
 package main
 
 import (
-	"bufio"
 	"bytes"
 	"os"
 	"os/exec"
@@ -16,42 +15,6 @@ import (
 	"github.com/rogpeppe/go-internal/testscript"
 	"github.com/stretchr/testify/require"
 )
-
-func isDocker() bool {
-	if _, err := os.Stat("/.dockerenv"); err == nil {
-		return true
-	}
-
-	paths := []string{"/proc/1/cgroup", "/proc/self/cgroup"}
-	for _, path := range paths {
-		file, err := os.Open(path)
-		if err != nil {
-			continue
-		}
-
-		scanner := bufio.NewScanner(file)
-		isDocker := false
-		for scanner.Scan() {
-			if strings.Contains(scanner.Text(), "docker") || strings.Contains(scanner.Text(), "kubepods") {
-				isDocker = true
-				break
-			}
-		}
-
-		if err := scanner.Err(); err != nil {
-			_ = file.Close()
-			return false
-		}
-
-		_ = file.Close()
-
-		if isDocker {
-			return true
-		}
-	}
-
-	return false
-}
 
 func TestMain(m *testing.M) {
 	os.Exit(testscript.RunMain(m, map[string]func() int{
@@ -71,10 +34,6 @@ func TestRunme(t *testing.T) {
 }
 
 func TestRunmeFilePermissions(t *testing.T) {
-	if isDocker() {
-		return
-	}
-
 	testscript.Run(t, testscript.Params{
 		Dir:             "testdata/permissions",
 		ContinueOnError: true,
--- a/main_test.go
+++ b/main_test.go
@@ -3,6 +3,7 @@
 package main
 
 import (
+	"bufio"
 	"bytes"
 	"os"
 	"os/exec"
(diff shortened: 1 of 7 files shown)


pkg/project/project.go
```diff
@@ -366,6 +366,12 @@ func (p *Project) loadFromDirectory(
 
 	err := util.Walk(p.fs, ".", func(path string, info fs.FileInfo, err error) error {
 		if err != nil {
+			if errors.Is(err, os.ErrPermission) {
```
sourishkrout, line 369: Let's be sure to log these instances `p.logger` should be available here

pkg/project/project.go
```diff
@@ -366,6 +366,13 @@ func (p *Project) loadFromDirectory(
...
```
sourishkrout, line 369: you should be able to unnest this if since it early returns and a `nil` value for error should result in a false, no?

pkg/project/project.go
```diff
@@ -455,6 +462,11 @@ func (p *Project) extractTasksFromFile(
 	})
 
 	if err != nil {
+		if errors.Is(err, os.ErrPermission) {
```
sourishkrout, line 465: same here, can we unnest for the same reasons as above?

pkg/project/project.go
sourishkrout: also, if we have more than a single condition to check the `switch err.(type)` pattern is likely more idiomatic in Golang

main_test.go
```diff
@@ -33,6 +70,17 @@ func TestRunme(t *testing.T) {
 	})
 }
 
+func TestRunmeFilePermissions(t *testing.T) {
+	if isDocker() {
```
sourishkrout, line 74: Consider moving this test case to a new file and use tags to toggle inclusion/exclusion from docker? perhaps i'm overlooking something.

sourishkrout: lemme ask a question: why is this `isDocker` check required? do these tests ever run in docker?

sourishkrout (changes requested):
Some questions.

main_test.go
pastuxso: @sourishkrout, For some reason, I don't know why the **chmod** command is not working. It is related to the volume. So I have to check if it's running inside a container to skip it. It is passing successfully locally and inside Github Actions Runner.

pkg/project/project.go
```diff
@@ -366,6 +366,13 @@ func (p *Project) loadFromDirectory(
 
 	err := util.Walk(p.fs, ".", func(path string, info fs.FileInfo, err error) error {
 		if err != nil {
+			if errors.Is(err, os.ErrPermission) {
+				p.logger.Warn("permission denied", zap.String("path", path), zap.Error(err))
```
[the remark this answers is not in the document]
sourishkrout: that's a good point, i think we should go with better UX

main_test.go
adambabik: I can replicate different results when running this test in Docker and locally. I suspect that the Docker image does not create a user and runs everything as root which results in no permission denied error even for the `000` permissions.

Either try that or control this with build tags. If the latter, it's a bit complicated... We already have a tag `test_with_docker`. It means that the file, like `command_docker_test.go` should be run only in the environment that provides Docker. We could introduce another tag which would indicate that the test should not in a Docker environment. It will require setting it and passing properly. I would try the option with a non-root user first.

sourishkrout:
@pastuxso is this still in draft since it's gotten two reviews already?

pastuxso:
@sourishkrout ready for review. I'll push in a separate branch an attempt to improve test running using a non-root user within Docker.

pkg/project/project.go
```diff
@@ -365,14 +365,22 @@ func (p *Project) loadFromDirectory(
...
+
+		switch err.(type) {
+		case nil:
+		case *os.PathError:
+			if !ignored {
+				p.logger.Warn("path error", zap.String("path", path), zap.Error(err))
+			}
+			ignored = true
+		default:
+			return err
```
sourishkrout, line 381: we should probably return the error to the client to be consistent with https://github.com/stateful/runme/blob/645-runme-attempts-to-access-unreadable-files-or-directories/pkg/project/project.go#L470-L475

sourishkrout (changes requested):
almost there

pkg/project/project.go
pastuxso: @sourishkrout If we return the error, it stops the execution. In this case, the approach ignores the permission error to continue walking through the dirs. Let me check if there is a way to return it and continue without halting.

sourishkrout: i think what you had before was right for the "handled error". i meant, the `default` case which is unhandled errors. when the error is return here will it be returned to the client?

pastuxso: @sourishkrout Yes, it will continue working as it did before. It will be handled by the following statement: 

```go
  if err != nil {
		p.send(ctx, eventc, LoadEvent{
			Type: LoadEventError,
			Data: LoadEventErrorData{Err: err},
		})
	}
```	

I'll revert the latest commit.

PR #650, opened 2024-08-08
Temp provide autoconf defaults

sourishkrout:
TBD if we automatically use defaults transparently or prompt the user to write a `runme.yaml` to disk. However, for the time being this does not hard block `runme beta` commands without much explanation about the absence of `runme.yaml`.

If we stick with transparent defaults we should probably rename `RootConfig` to `RunmeConfig` or some such.

Changes to the config protos were necessary to distinguish between null bools versus false.


sourishkrout:
I'm now realizing I could have done this with dependency injection 🤔 

https://github.com/stateful/runme/blob/seb/autoconf-defaults/internal/runnerv2service/service_execute_test.go#L35-L45

internal/config/autoconfig/autoconfig.go
```diff
@@ -210,22 +210,26 @@ func getProjectFilters(c *config.Config) ([]project.Filter, error) {
...
+
 	content, err := cfgLoader.RootConfig()
-	if err != nil {
+	switch err {
+	case nil:
+		if cfg, err = config.ParseYAML(content); err != nil {
+			return nil, err
+		}
+	case config.ErrRootConfigNotFound:
+		cfg = &config.ConfigDefaults
```
adambabik, line 222: Let's make sure we return a copy and not a pointer to a default object.

```suggestion
		temp := config.ConfigDefaults
		cfg = &temp
```

internal/config/config.go
```diff
@@ -42,6 +43,10 @@ type Config struct {
 	LogVerbose bool
 }
 
+func Defaults() Config {
+	return ConfigDefaults
```
adambabik, line 47: If you provide a function to get the default config, I think you can hide `ConfigDefaults`. Otherwise, it might be confusing and `ConfigDefaults` might be used incorrectly.

internal/config/config_defaults.go
```diff
@@ -0,0 +1,77 @@
+package config
+
+var ConfigDefaults Config
```
adambabik, line 3: Nit: No need to prefix it with `Config` because it's already in the `config` package. In my other comment, I suggested to hide it as you have `Defaults()` function.

```suggestion
var defaults Config
```

internal/tls/tls.go
```diff
@@ -169,6 +169,12 @@ func generateCertificate(certFile, keyFile string) (*tls.Config, error) {
 		return nil, errors.WithStack(err)
 	}
 
+	if _, err := os.Stat(filepath.Dir(certFile)); errors.Is(err, fs.ErrNotExist) {
```
adambabik, line 172: In theory, the key file might be in a different directory so this check should be repeated.

internal/config/config.go
```diff
@@ -125,32 +130,39 @@ func configV1alpha1ToConfig(c *configv1alpha1.Config) (*Config, error) {
...
+	cfg.ServerTLSKeyFile = server.GetTls().GetKeyFile()
+
+	cfg.LogEnabled = unwrapBool(log.GetEnabled(), cfg.LogEnabled)
+	cfg.LogPath = log.GetPath()
+	cfg.LogVerbose = unwrapBool(log.GetVerbose(), cfg.LogVerbose)
+
+	return &cfg, nil
+}
+
+func unwrapBool(wrappedVal *wrapperspb.BoolValue, defaultVal bool) bool {
```
adambabik, line 160: Nit: I have a generic alternative:

```go
func setIfHasValue[T any](prop *T, val interface{ GetValue() T }) {
	if val != nil {
		*prop = val.GetValue()
	}
}
```

And usage is as follows:
```go
setIfHasValue(&cfg.ServerTLSEnabled, server.GetTls().GetEnabled())
```

internal/tls/tls.go
sourishkrout: totally. overlooked that in v2 we could have separate paths as opposed to the v1 base dir for TLS.

internal/config/config.go
sourishkrout: love it!

internal/config/autoconfig/autoconfig.go
sourishkrout: Made it so that `config.Defaults()` always returns a deep copy

PR #662, opened 2024-09-02
Remove AI Service Protocol Definition

jlewi:
* Foyle is now publishing its protos to the buf registry and the client libraries pulled from the registry are what's used by the vscode extension.
* The AI service defined in the runme repository is no longer needed.
* Related to jlewi/foyle#173


sourishkrout (approved):
✅ LGTM

PR #667, opened 2024-09-13
Substitute category (deprecate) with tag

sourishkrout:
Users have reported that `tags` are self-explanatory, whereas `categories` are not.

1. In the stable CLI interface, let's move off `category` without interrupting folks.
2. In `beta`, there will only be a `tag`.


sourishkrout:
Another review, please, @pastuxso.

internal/config/config_filter.go
```diff
@@ -34,7 +34,7 @@ type FilterDocumentEnv struct {
 // Without it, all variables start with capitalized letters.
 type FilterBlockEnv struct {
 	Background             bool     `expr:"background"`
-	Categories             []string `expr:"categories"`
+	Tags                   []string `expr:"tags"`
```
adambabik, line 37: Suggestion: fields are sorted alphabetically in this struct.

internal/config/autoconfig/autoconfig.go
```diff
@@ -171,7 +171,7 @@ func getProjectFilters(c *config.Config) ([]project.Filter, error) {
 			filters = append(filters, project.Filter(func(t project.Task) (bool, error) {
 				env := config.FilterBlockEnv{
 					Background: t.CodeBlock.Background(),
-					Categories: t.CodeBlock.Categories(),
+					Tags:       t.CodeBlock.Tags(),
```
adambabik, line 174: Same suggestion about sorting fields.

pkg/document/block.go
```diff
@@ -102,12 +102,16 @@ func (b *CodeBlock) Background() bool {
 	return val
 }
 
-func (b *CodeBlock) Categories() []string {
-	cat, ok := b.Attributes()["category"]
+func (b *CodeBlock) Tags() []string {
+	tag, ok := b.Attributes()["category"]
+	if !ok {
+		tag, ok = b.Attributes()["tag"]
```
adambabik, line 108: We don't care about a case where both are set?

pkg/document/frontmatter.go
```diff
@@ -58,6 +58,7 @@ type Frontmatter struct {
 	Shell        string         `yaml:"shell"`
 	Cwd          string         `yaml:"cwd"`
 	Category     string         `yaml:"category"`
```
adambabik, line 60: Should `Category` be marked as deprecated?

pkg/api/proto/runme/parser/v1/parser.proto
```diff
@@ -95,6 +95,7 @@ message Frontmatter {
   FrontmatterRunme runme = 4;
   string category = 5;
```
adambabik, line 96: Should we comment `category` as deprecated?

internal/cmd/run.go
```diff
@@ -87,35 +92,35 @@ func runCmd() *cobra.Command {
...
+							fmTags := resolveFrontmatterTags(fm)
 							match := false
-							if len(fmCategories) > 0 && containsCategories(fmCategories, cmdCategories) {
-								if len(blockCategories) == 0 {
+							if len(fmTags) > 0 && containsCategories(fmTags, cmdTags) {
+								if len(blockTags) == 0 {
 									match = true
 								} else {
-									match = containsCategories(fmCategories, blockCategories)
+									match = containsCategories(fmTags, blockTags)
```
pastuxso, line 116: What do you think on renaming the function to `containsTags`?

pastuxso (changes requested):
Works fine, I left you a minor comment.

sourishkrout:
Whoops, I somehow overlooked @adambabik's comments. I am addressing them now.

pkg/document/block.go
sourishkrout: Right, I am combining them instead.

PR #668, opened 2024-09-13
Only resolve programs that are in shell

sourishkrout:
The CLI is broken when any non-shell tasks/cells are being run. The problem is that everything's treated as Shell and goes through the program resolver (env var extraction/prompting etc).

This PR passes the `language id` into the resolution to allow for an early return. Down the road, the resolve program could return language-specific resolutions. Once the protos are published and the extension passes `language id` the field will be required.

Example of the issue: https://gist.github.com/sourishkrout/fb7c54fb77b2a1a5d773f33072b8816d#file-01j7ps7tehd767qy13zz0qcysj-md


pastuxso (approved):
LGTM 🚀

PR #669, opened 2024-09-17
Allow multiple args for interpreters

sourishkrout:
This is a bug where we only expect one argument. Downstream, the OS libraries expect arguments to be passed as an array.

This PR fixes this. Please see the test case.


pastuxso (approved):
LGTM 🚀

PR #671, opened 2024-09-19
Telemetry (turn off with DO_NOT_TRACK env var)

sourishkrout:
Introducing anonymous telemetry that respects common ENV var `DO_NOT_TRACK=true` or `DO_NOT_TRACK=1`.

### DIFF
--- a/internal/telemetry/scarf.go
+++ b/internal/telemetry/scarf.go
@@ -1,0 +1,110 @@
+package telemetry
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+)
+
+const (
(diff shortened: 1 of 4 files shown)


internal/telemetry/scarf.go
```diff
@@ -0,0 +1,110 @@
...
+const (
+	base   = "https://home.runme.dev/"
+	client = "Kernel"
+)
+
+type LookupEnv func(key string) (string, bool)
+
+// Returns true if telemetry reporting is enabled, false otherwise.
+func ReportUnlessNoTracking(logger *zap.Logger) bool {
+	if v := os.Getenv("DO_NOT_TRACK"); v != "" && v != "0" && v != "false" {
```
adambabik, line 25: I really like `strconv.ParseBool`'s semantic which takes into account sane values and converts them to boolean.

sourishkrout: Good point 👌.

PR #672, opened 2024-09-23
runnerv2client: add client and "beta run --server"

adambabik:
With `beta run --server` it's possible to execute a Markdown code block on the server from the CLI.

This is a prerequisite for supporting the execution of code blocks in the remote locations using the CLI, even if the remote location is a well-defined sandbox Docker container running locally.

## Testing

First, start the server:

```
go run . beta server start
```

Next, execute a code block on the server using the CLI:

```
go run . beta run -s=local print-name
```


sourishkrout (approved):
I need to spend more time with this after KubeCon. However, let's merge since it's part of the `beta` command group and subject to change.

PR #677, opened 2024-10-01
Clear relationship between ephemeral and persistent lifecycle IDs

sourishkrout:
When we first introduced lifecycle identity, it was designed to turn persistence on or off at build time. The following changes will ideally allow clients to toggle persistence dynamically.

1. Always use ephemeral IDs if there are no persistent IDs in the document/node models.
2. Decide persistence exclusively at deserialization time in the Parser.

I have cleaned up `runme fmt` as part of these changes, changed to run `code-server` in docker-only (avoid rate limiting locally), and made runme terminal stop flaking when run inside of Docker.

### DIFF
--- a/cmd/gqltool/main.go
+++ b/cmd/gqltool/main.go
@@ -11,10 +11,8 @@ import (
 	"github.com/stateful/runme/v3/internal/client/graphql"
 )
 
-var (
-	apiURL = flag.String("api-url", "http://localhost:4000", "The API base address")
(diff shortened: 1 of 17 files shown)


sourishkrout:
The next thing to figure out is if it makes sense to harvest some of https://github.com/stateful/runme/pull/558 into this PR.

sourishkrout:
> The next thing to figure out is if it makes sense to harvest some of #558 into this PR.

Based on my findings, I think we should forgo #558 because it does not solve the problem on the extension side.

PR #678, opened 2024-10-01
internal/config: use JSON Schema

adambabik:
This change switches the `runme.yaml` schema definition from protobuf to JSON Schema. This results in fewer conversions and more natural representation. It also adds support for parsing and merging multiple configuration files. Default config was moved to [a file](internal/config/runme.default.yaml).

Go's struct representing the config is generated from the schema. When you change `internal/config/config.schema.json`, remember to call `make schema/generate`.


sourishkrout (approved):
LGTM ✅

PR #679, opened 2024-10-02
Cleanup: Remove the ai-logs flag.

jlewi:
* This flag is deprecated and no longer used.
* The frontend should have stopped using this flag starting in 3.7.5
* The frontend is now up to 3.8.5
* So we should be good to remove this flag because the frontend should no longer be setting it.

* Related to https://github.com/jlewi/foyle/issues/211

### DIFF
--- a/internal/cmd/server.go
+++ b/internal/cmd/server.go
@@ -36,7 +36,6 @@ func serverCmd() *cobra.Command {
 		devMode      bool
 		enableRunner bool
 		tlsDir       string
-		enableAILogs bool
 	)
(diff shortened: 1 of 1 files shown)


sourishkrout:
You doing push & pray over there 😆, @jlewi?

jlewi:
You caught me.

sourishkrout (approved):
✅ LGTM. I have an eye on the flakes. I made them more stable on Linux but they still only seem to flake in GHA :-/.

PR #683, opened 2024-10-13
Implement "runme beta session"

adambabik:
It implements a new command that allows starting a session and collecting all exported environment variables throughout its lifetime.

## Testing

As a prerequisite, add the following to your shell's startup script (`.bashrc` or `.zshrc`):

```
# runme
source <(runme beta session setup)
```

Then, you can start a new session in the terminal:

```
runme beta session
```

The session is a regular shell of your choice, inferred from `$SHELL`. You can use it like a normal shell, but any exported environment variables will be returned upon exiting the session via `exit`.

Currently, the collected variables are printed out, but in the future, they will be routed to a proper store.


adambabik:
https://github.com/stateful/runme/pull/683/commits/7351fbf56ded155bad4583187ee5c1b0d64886ca can be cherry-picked to main safely. After one of the GitHub Actions update, the CI workflow does not work.

internal/cmd/beta/session_cmd.go
```diff
@@ -0,0 +1,139 @@
...
+		Hidden: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return autoconfig.InvokeForCommand(
+				func(
+					cmdFactory command.Factory,
+					logger *zap.Logger,
+				) error {
+					defer logger.Sync()
+
+					if val, err := strconv.ParseBool(os.Getenv(command.EnvCollectorSessionEnvName)); err != nil || !val {
```
sourishkrout, line 107: This actually does not return `false` when the ENV is unset. Looks like an empty string `""` isn't converted into `false`? That's odd.

sourishkrout:
I'm getting this locally on `macOS` with `zsh`:

```
# runme
source <(runme beta session setup)
/dev/fd/14:3: parse error near `\n'
```

internal/cmd/beta/session_cmd.go
adambabik: It does return `false` for an empty string, [example](https://go.dev/play/p/TTvuomAM2vf), but because it's a zero/default value for boolean type. It actually returns an error, so the result is correct but the log is confusing. I will fix.

adambabik:
> I'm getting this locally on `macOS` with `zsh`:
> 
> ```
> # runme
> source <(runme beta session setup)
> /dev/fd/14:3: parse error near `\n'
> ```
> 
> Same happens when I source it inside of `.zshrc`.

@sourishkrout can you paste the output of executing `runme beta session setup`, please?

sourishkrout:
> @sourishkrout can you paste the output of executing `runme beta session setup`, please?

here you go:

```sh
$ runme beta session setup
#!/bin/sh
/Users/sourishkrout/Projects/stateful/oss/runme/runme env dump --insecure > 
__cleanup() {
rv=$?
/Users/sourishkrout/Projects/stateful/oss/runme/runme env dump --insecure > 
exit $rv
}
trap -- "__cleanup" EXIT```

**Comment by adambabik** [comment]
> oh, I see the problem now... no destinations for the redirects

Thanks! Yeah, I will add a check to protect from this... Overall, `RUNME_SESSION=1` is required to activate the session setup. Then, two envs from the env collector that define those redirects are used. This is all hidden from the user when using `runme beta session`, but these edge cases should be handled.

**Comment by sourishkrout** [comment]
It does not work for me. Can you provide step-by-step instructions on how to start with a clean slate? How are you running this locally?

I mainly don't understand how the `setup` command will work without `RUNME_SESSION_PREPATH` and `RUNME_SESSION_POSTPATH` present or how `RUNME_SESSION=1` need to be `1`?

Isn't this a chicken/egg problem @adambabik?

**Comment by adambabik** [comment]
@sourishkrout sorry, I missed your comment. It should work exactly as it's described in the PR description. In my case: 
1. I put `source <(/Users/myuser/path/to/runme/runme beta session setup)` where `runme` is a binary build in the project root directory on this branch. 
2. Do `./runme beta session` and it should display `Runme session active. When you're done, execute "exit".`.
3. Export some env like `export TEST=1`
4. Type `exit`. It should print:
```
Collected env during the session:
TEST=1
```

All the shenanigans with the `RUNME_` envs are handled by `beta session setup` and `beta session`.

**Comment by sourishkrout** [comment]
> Collected env during the session:
> TEST=1

It shows empty for me following the exact instructions 🤔. Not sure why, @adambabik.

File: internal/cmd/beta/session_cmd.go [hunk 2]
```diff
@@ -0,0 +1,197 @@
...
+					); err != nil {
+						logger.Info("session setup is skipped because the environment variable is not set", zap.Error(err))
+						return writeNoopShellCommand(out)
+					}
+
+					sessionSetupEnabled := os.Getenv(command.EnvCollectorSessionEnvName)
+					if val, err := strconv.ParseBool(sessionSetupEnabled); err != nil || !val {
+						logger.Debug("session setup is skipped", zap.Error(err), zap.Bool("value", val))
+						return writeNoopShellCommand(out)
+					}
```
sourishkrout, line 158: Also, is `RUNME_SESSION` expected to be a bool (serialized)? We are using the env var to hold the session ID all over the place so a semantic clash here is bad if bool <> SessionID (both string).

internal/cmd/beta/session_cmd.go
```diff
@@ -0,0 +1,197 @@
...
+					out := cmd.OutOrStdout()
+
+					if err := requireEnvs(
+						command.EnvCollectorSessionEnvName,
+						command.EnvCollectorSessionPrePathEnvName,
+						command.EnvCollectorSessionPostPathEnvName,
+					); err != nil {
+						logger.Info("session setup is skipped because the environment variable is not set", zap.Error(err))
+						return writeNoopShellCommand(out)
+					}
```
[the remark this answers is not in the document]
adambabik: The order is that `beta session` starts a shell process with those envs provided first. This shell process starts and executes, in the case of zsh, the `.zshrc` file inside which `beta session setup` is called. 

It's perfectly normal that it typically will error and output no-op shell command in `.zshrc`, but from the user point of view it should be invisible. We can recommend protecting `source <(/path/to/runme beta session setup)` with a conditional checking `command.EnvCollectorSessionEnvName` to avoid that.

internal/cmd/beta/session_cmd.go
adambabik: Oh yeah, that's a good point. I didn't notice the clash. I will change it to something like `RUNME_TERM_SESSION_ENABLED`.

adambabik:
> > Collected env during the session:
> > TEST=1
> 
> It shows empty for me following the exact instructions 🤔. Not sure why, @adambabik.

In order to debug it, try `tail -f /tmp/runme.log` in another terminal window and then e.g. `source ~/.zshrc` or opening a new terminal window. The path is taken from `experimental/runme.yaml`. I noticed that it's too chatty for some logs but overall should contain an answer.

internal/cmd/beta/session_cmd.go
adambabik: Fixed with `_RUNME_TERMINAL_SESSION_ENABLED`. Check out [`terminal_session.go`](internal/command/terminal_session.go).

sourishkrout:
@adambabik 🤔 I'm not sure what I'm doing wrong, but this does not work for me:

image

sourishkrout (approved):
✅ LGTM. There is more work to do, but it is mergeable.

PR #694, opened 2024-11-01
Fix data races

adambabik:
This PR adds `go vet` call, including [`checklocks`](https://pkg.go.dev/gvisor.dev/gvisor/tools/checklocks), in `make lint`. It also fixes data races found using `RACE=true make test/execute`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,6 +106,8 @@ lint:
 		./...
 	@staticcheck ./...
 	@gosec -quiet -exclude=G110,G204,G304,G404 -exclude-generated ./...
+	@go vet -stdmethods=false ./...
+	@go vet -vettool=$(shell go env GOPATH)/bin/checklocks ./...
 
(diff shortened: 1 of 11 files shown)


sourishkrout:
wtf, how did this merge go bad

sourishkrout:
@adambabik I'm puzzled about what went wrong when I used GitHub's `Update branch` button before the merge. However, I had to revert https://github.com/stateful/runme/commit/cf68547268c55466a9ef3a860f5bf09d9527baaf.

Could you please take a look? I can't get the docker tests to pass at all. I suspect some of your changes got lost?

adambabik:
@sourishkrout yeah, something went off. I didn't dig into that, but I created a new branch and rebased instead. I also dug a bit more into flaky tests in Docker and updated a couple. Let's review https://github.com/stateful/runme/pull/697 and try to merge again.

PR #697, opened 2024-11-08
Fix data races

adambabik:
This PR adds `go vet` call, including [`checklocks`](https://pkg.go.dev/gvisor.dev/gvisor/tools/checklocks), in `make lint`. It also fixes data races found using `RACE=true make test/execute`.

Retrying reverted https://github.com/stateful/runme/pull/694.


sourishkrout (approved):
✅ LGTM

PR #698, opened 2024-11-09
Clean up test code and increase parallelization

adambabik:
Nothing fancy, just cleaning up some tests and adding `t.Parallel()`.


sourishkrout (approved):
✅ LGTM

PR #702, opened 2024-11-21
Allow code blocks to be selectively rendered as markdown

sourishkrout:
This PR allows more control over how code blocks are being rendered.

1. Introduce new annotation `transform` which defaults to `true`. If set to `false` code blocks will be deserialized as markdown.
2. Make exception for `mermaid` in `languageID` which can be overwritten with `transform: true`

This unlocks Mermaid diagram support in the notebooks.


pastuxso (approved):
LGTM 👍

PR #717, opened 2024-12-17
Unescape shell literal if it's not a placeholder

### LINKED ISSUE #709: Env prompt: Escape sequences
Hello there!

Just an optical inconvenience...

Given following snippet:

```bash
export TYPE=[Guest type \(hyperv,proxmox,openstack\)]
export TARGET=[IP of waiting server]
```

Result:
image

The backslash-escape in front of the parentheses is required in bash. Runme should probably parse, sanitize and display these properly.

sourishkrout:
Prompt messages are unquoted which requires them to be valid shell. Any values not quoted become prompt messages and are exclusively for human consumption. This is why we're removing backslashes from any form of parentheses.

Fixes #709.

### DIFF
--- a/internal/command/program_resolver.go
+++ b/internal/command/program_resolver.go
@@ -254,7 +254,12 @@ func (r *ProgramResolver) findOriginalValue(decl *syntax.DeclClause) (string, bo
 		return true
 	})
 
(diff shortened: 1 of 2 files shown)


pastuxso (approved):
LGTM 🚀

PR #729, opened 2025-01-18
fix: shell dry-run output should be runable

lalyos:
I've redirected the output of `runme run --dry-run` into a file.
The code block was of bash type and the file nicely starts with shebang:

```
#!/usr/local/bin/bash

// run in "/Users/lalyos/somedir"

set -e -o pipefail
echo hello
```
but the `run id "dir"` line supposed to be a comment, but the syntaxt doesn't fit shell syntax.


adambabik (approved):
Good catch! Thanks for the contribution, LGTM!

adambabik:
@lalyos it seems that this PR is out of date. Could you rebase it on the main branch?

lalyos:
rebased

sourishkrout:
This change was included in https://github.com/stateful/runme/releases, @lalyos.

PR #735, opened 2025-01-30
Retain markdown's fenced code block encoding format

sourishkrout:
Not all tools processing markdowns equally accommodate attributes/annotations [inside fenced code blocks](https://docs.runme.dev/configuration/cell-level#configure-using-markdown). This is especially true if they aren't encoded in HTML attribute format. A notable example is https://gohugo.io/.

Instead of always preferring `JSON` encoding. This PR changes the default behavior of the attribute parser to retain the original encoding format.

### DIFF
--- a/internal/cmd/beta/beta_cmd.go
+++ b/internal/cmd/beta/beta_cmd.go
@@ -59,7 +59,6 @@ All commands use the runme.yaml configuration file.`,
 
 				return nil
 			})
-
 			// Print the error to stderr but don't return it because error modes
 			// are neither fully baked yet nor ready for users to consume.
 			if err != nil {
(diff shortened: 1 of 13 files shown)


adambabik:
@sourishkrout Initially I wanted to comment with suggestions, but as the number of comments was growing, I submitted https://github.com/stateful/runme/pull/735/commits/5d0f8286265c722787592470ee2628bd84ff6743 instead. The main change is removal of the `AttributeStore` interface as it is not needed; `Attributes` is just fine.

adambabik (approved):

sourishkrout:
> @sourishkrout Initially I wanted to comment with suggestions, but as the number of comments was growing, I submitted [5d0f828](https://github.com/stateful/runme/commit/5d0f8286265c722787592470ee2628bd84ff6743) instead. The main change is removal of the `AttributeStore` interface as it is not needed; `Attributes` is just fine.

Sounds good 👌. Thank you!

PR #736, opened 2025-02-05
First class dagger shell

sourishkrout:
In order to natively support Dagger's shell Runme needs to take the entire notebook as opposed to an atomic cell into account. This PR introduces the APIs necessary to transform a notebook into a Dagger Shell script using some basic AST transformation.

Integration with extension in https://github.com/stateful/vscode-runme/pull/1947.

PS: The protos are not published yet and need local-gen to work.

The new service will turn a deserialized version of the following markdown into a runnable Dagger shell script.

## Simple example notebook

Unamed cell:

```sh {"closeTerminalOnSuccess":"false","interactive":"true","openTerminalOnError":"true","terminalRows":"23"}
git github.com/stateful/runme |
    head |
    tree |
    file examples/README.md
```

And the same as a named cell:

```sh {"name":"README"}
# Named in README in cell attributes
git github.com/stateful/runme |
    head |
    tree |
    file examples/README.md
```

## Resulting Dagger Shell Script

Running the unamed cell:

```sh
DAGGER_01JKE40B6A0F5NBESE49X122RD()
{
  git github.com/stateful/runme \
    | head \
    | tree \
    | file examples/README.md
}
DAGGER_01JKE40B6A0F5NBESE49X122RD
```

Running the named cell:

```sh
DAGGER_01JKE4K781F87XRK9SQZ9T1GQD()
{
  git github.com/stateful/runme \
    | head \
    | tree \
    | file examples/README.md
}
README()
{
  git github.com/stateful/runme | head | tree | file examples/README.md
}
README
```


pkg/api/proto/runme/notebook/v1alpha1/notebook.proto
```diff
@@ -0,0 +1,27 @@
...
+}
+
+message ResolveNotebookResponse {
+  string script = 1;
+}
+
+service NotebookService {
+  // ResolveNotebook that takes a notebook and a cellIndex and returns a script
+  // that takes the whole notebook into account to run as cell.
+  rpc ResolveNotebook(ResolveNotebookRequest) returns (ResolveNotebookResponse) {}
```
adambabik, line 26: Question: Is there any reason this method couldn't be in `RunnerService`?

internal/notebook/service.go
```diff
@@ -0,0 +1,47 @@
...
+	if notebook == nil {
+		return nil, status.Error(codes.InvalidArgument, "notebook is required")
+	}
+
+	cellIndex := req.GetCellIndex()
+	if cellIndex == nil {
+		return nil, status.Error(codes.InvalidArgument, "cell index is required")
+	}
+
+	commandMode := req.GetCommandMode()
```
adambabik, line 40: Suggestion: Let's bring it to the top. No need to check anything else if the command mode is not supported.

internal/notebook/resolver.go
```diff
@@ -0,0 +1,132 @@
...
+			return nil, err
+		}
+	}
+
+	var rendered bytes.Buffer
+	if err := script.RenderWithCall(&rendered, targetName); err != nil {
+		return nil, err
+	}
+
+	return &notebookv1alpha1.ResolveNotebookResponse{
```
adambabik, line 129: Suggestion: Returning `renderer.String()` should be easier to reason about and tests. The service can wrap it as it pleases without polluting other files with `notebookv1alpha1`.

internal/notebook/resolver.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func NewNotebookResolver(notebook *parserv1.Notebook) *NotebookResolver {
+	return &NotebookResolver{
+		notebook: notebook,
+		editor:   editorservice.NewParserServiceServer(zap.NewNop()),
+	}
+}
+
+func (r *NotebookResolver) ParseNotebook(context context.Context) (*parserv1.Notebook, error) {
```
adambabik, line 29: Suggestion: Can be unexported.

internal/notebook/resolver.go
```diff
@@ -0,0 +1,132 @@
...
+
+func (r *NotebookResolver) ResolveDaggerShell(context context.Context, cellIndex uint32) (*notebookv1alpha1.ResolveNotebookResponse, error) {
+	notebook, err := r.ParseNotebook(context)
+	if err != nil {
+		return nil, err
+	}
+
+	var targetCell *parserv1.Cell
+	targetName := ""
+	for idx, cell := range notebook.Cells {
```
adambabik, line 69: Suggestion: I think you can skip `for`: 
```suggestion
    // TODO: check if not out of bounds
	cell := notebook.Cells[cellIndex]
```

adambabik (approved):
A couple of suggestions, but overall LGTM. 

One more asks I have is to add an example in the PR description of the input (exemplary Markdown) and the expected output (a script) and how it will be executed.

pkg/api/proto/runme/notebook/v1alpha1/notebook.proto
sourishkrout: That's where I had it first. I decided to make it a separate service because the `Runner` only knows about atomic cells, and this service needs the entire notebook. Perhaps that's not "different enough". We could consider bringing it back into the Runner as part of runnerv2.

internal/notebook/service.go
sourishkrout: Totally.

internal/notebook/resolver.go
sourishkrout: 👍

internal/notebook/resolver.go
sourishkrout: Good point!

PR #738, opened 2025-02-10
Make cell names more obvious in markdown viewers

sourishkrout:
Cells with names will contain a leading comment with the name of the cell. This is helpful if you are new to Runme. Let's start with `dagger shell` first because call names are used to string together functions/pipelines.

Example:

image


pastuxso (approved):
LGTM 🚀

PR #745, opened 2025-02-14
Dagger Shell natively in CLI

sourishkrout:
Natively run Dagger Shell when frontmatter (`shell: dagger shell`) requires it. An example is available [here](https://github.com/stateful/runme/pull/745/commits/7c1b4b5462c0751e03c465b71cbb9965a9a09b84).

Since the Notebook Resolver is entirely stateless we skip the GRPC service layer. Please note that `txtar` e2e test coverage of the CLI is excluded from SonarCloud.

https://github.com/user-attachments/assets/f7c273bd-a674-41cd-aa03-0e3cf648ef13

PS: The way Dagger's TUI renders in Runme's TUI isn't fantastic. However, since no low-hanging fix was available this will have to wait for a separate effort/PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,7 @@ jobs:
           version: "latest"
           verb: core
           args: "engine local-cache"
+        if: ${{ matrix.os == 'ubuntu-latest' }}
       - name: Setup go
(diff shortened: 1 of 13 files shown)


pastuxso (approved):
LGTM 🚀

Issue #744, opened 2025-02-13
Switch to Go 1.24


adambabik:
[Go 1.24 is out](https://go.dev/blog/go1.24).

Changes to do:
- Bump Go version in `.github/workflows/ci.yml`
- Try to remove the Makefile target `install/dev` and replace installing tools with the `-tool` flag in `go get`
- There are likely a few places where `rand.Text()` could be used
- As well as [iterators](https://antonz.org/go-1-24/#more-string-and-byte-iterators) on `string` and `[]byte`

PR #746, opened 2025-02-18
go: upgrade to 1.24

adambabik:
Closes https://github.com/stateful/runme/issues/744


Makefile
```diff
@@ -76,10 +76,6 @@ test-docker/run:
 		-v dev.runme.test-env-gocache:/root/.cache/go-build \
 		runme-test-env:latest
 
-.PHONY: test/update-snapshots
-test/update-snapshots:
-	@TZ=UTC UPDATE_SNAPSHOTS=true go test ./...
```
adambabik, line 81: Deleted, because I could not find any reference to it.

internal/cmd/tui.go
```diff
@@ -335,9 +335,9 @@ func (m tuiModel) View() string {
 			intro := block.Intro()
 			words := strings.Split(intro, " ")
 			// todo(sebastian): this likely only works well for English
-			max := 9 - len(strings.Split(relFilename, string(filepath.Separator)))
-			if len(words) > max {
-				intro = strings.Join(words[:max], " ") + "..."
+			maxComponents := 9 - len(strings.Split(relFilename, string(filepath.Separator)))
```
adambabik, line 338: Changed a variable name, because `gosec` complained that it was overwriting a global symbol. AFAIK from Go 1.23 there are `max` and `min` functions.

internal/runnerv2service/execution.go
```diff
@@ -184,6 +184,10 @@ func (e *execution) Wait(ctx context.Context, sender runnerv2.RunnerService_Exec
 	e.closeIO()
 
 	if waitErr != nil {
+		// Drain the readSendDone channel to avoid goroutine leaks.
```
adambabik, line 187: Found when running e2e tests thanks to the logger reporting calls after the test case finished.

sourishkrout (commented):
✅ LGTM

PR #755, opened 2025-03-03
fix: fix the wrong error return value

rustco:
In fact, err is incorrect and should be `c.init()` .


sourishkrout:
Hey @rustco, any chance you could share the test case / repro that lead you to this issue?

I want to use it to add a regression test.

rustco:
> Hey @rustco, any chance you could share the test case / repro that lead you to this issue?
> 
> I want to use it to add a regression test.

Sorry, I found the problem directly through the code.

internal/command/env_collector_fifo_unix.go
```diff
@@ -47,8 +47,8 @@ func newEnvCollectorFifo(
 		temp:     temp,
 	}
 
-	if c.init() != nil {
-		return nil, err
+	if errC := c.init(); errC != nil {
```
adambabik, line 50:
```suggestion
	if err := c.init(); err != nil {
```

internal/command/env_collector_fifo_unix.go
```diff
@@ -47,8 +47,8 @@ func newEnvCollectorFifo(
...
+		return nil, errC
```
adambabik, line 51:
```suggestion
		return nil, err
```

adambabik (approved):
This line is suspicious. It looks like a typo to me and it very likely should be fixed as proposed. Let's make sure the tests pass. It might have happened that some of the tests work, because the error was not propagated...

PR #759, opened 2025-03-09
Replace `golang.org/x/exp` with stdlib

Juneezee:
These experimental packages are now available in the Go standard library.

1. `golang.org/x/exp/slices` -> `slices` (https://go.dev/doc/go1.21#slices)
2. `golang.org/x/exp/rand` -> `math/rand/v2` (https://go.dev/doc/go1.22#math_rand_v2)

`golang.org/x/exp/rand` has been deprecated and scheduled to be deleted.

Reference: https://github.com/golang/exp/commit/f9890c6ad9f380fd3cdb66a33843f522d4790e03

### DIFF
--- a/internal/dockerexec/docker.go
+++ b/internal/dockerexec/docker.go
@@ -2,10 +2,9 @@ package dockerexec
 
 import (
 	"context"
-	"encoding/binary"
 	"encoding/hex"
 	"io"
-	"math/rand/v2"
+	"math/rand"
 	"time"
 
 	"github.com/docker/docker/api/types/filters"
@@ -35,9 +34,7 @@ func New(opts *Options) (*Docker, error) {
 		logger = zap.NewNop()
 	}
 
-	var seed [32]byte
-	binary.LittleEndian.PutUint64(seed[:], uint64(time.Now().UnixNano()))
-	rnd := rand.NewChaCha8(seed)
+	rnd := rand.New(rand.NewSource(time.Now().UnixNano()))
 
 	d := &Docker{
 		client:       c,
@@ -64,7 +61,7 @@ type Docker struct {
 	image        string
 
 	logger *zap.Logger
-	rnd    *rand.ChaCha8
+	rnd    *rand.Rand
 }
 
 func (d *Docker) CommandContext(ctx context.Context, program string, args ...string) *Cmd {
--- a/internal/dockerexec/docker.go
+++ b/internal/dockerexec/docker.go
@@ -2,8 +2,10 @@ package dockerexec
 
 import (
 	"context"
+	"encoding/binary"
 	"encoding/hex"
 	"io"
+	"math/rand/v2"
 	"time"
 
 	"github.com/docker/docker/api/types/filters"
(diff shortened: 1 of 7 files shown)


internal/dockerexec/docker.go
```diff
@@ -34,7 +35,9 @@ func New(opts *Options) (*Docker, error) {
 		logger = zap.NewNop()
 	}
 
-	rnd := rand.New(rand.NewSource(uint64(time.Now().UnixNano())))
+	var seed [32]byte
```
adambabik, line 38: There is no need to use `math/rand/v2`. `math/rand` will be fine and the code can stay simple:

```go
rnd := rand.New(rand.NewSource(time.Now().UnixNano()))
```

internal/ulid/generator.go
```diff
@@ -18,12 +19,12 @@ var (
...
 // DefaultEntropy returns a reader that generates ULID entropy.
 // The default entropy function utilizes math/rand.Rand, which is not safe for concurrent use by multiple goroutines.
-// Therefore, this function employs x/exp/rand, as recommended by the authors of the library.
+// Therefore, this function employs math/rand/v2 (supersedes x/exp/rand), as recommended by the authors of the library.
 func DefaultEntropy() io.Reader {
 	entropyOnce.Do(func() {
-		seed := uint64(time.Now().UnixNano())
-		source := rand.NewSource(seed)
-		rng := rand.New(source)
+		var seed [32]byte
```
adambabik, line 25: The same here, `math/rand` is fine and then have: `rng := rand.New(rand.NewSource(time.Now().UnixNano()))`.

adambabik (changes requested):

adambabik:
Thanks for the contribution! I requested to stay with `math/rand`, because it will result in simpler code. The rest looks fine.

internal/ulid/generator.go
```diff
@@ -60,7 +37,7 @@ func GenerateID() string {
 }
 
 func DefaultGenerator() string {
-	entropy := DefaultEntropy()
+	entropy := ulid.DefaultEntropy()
```
Juneezee, line 40: After applying the code suggestion https://github.com/stateful/runme/pull/759#discussion_r1986251111, I realize that we can use the `DefaultEntropy()` function from `github.com/oklog/ulid/v2` since it also uses `math/rand` and `LockedMonotonicReader` internally.

https://github.com/oklog/ulid/blob/v2.1.0/ulid.go#L132-L140

Juneezee:
@adambabik @sourishkrout friendly ping for review :blush:

adambabik (approved):
Thanks, LGTM!

sourishkrout:
> @sourishkrout something odd happened and the old checks for Go 1.22 are required now. Did anything change with the configuration?

Not quite sure. However, we did move the repos out under the `stateful` into the `runmedev` org which might have something to do with it. I'll look into it.

sourishkrout:
Thanks for the contribution, @Juneezee!

PR #762, opened 2025-03-13
Move from stateful to runmedev org as part of CNCF sandbox admission

sourishkrout:
Move the new package namespace (`runmedev/runme`) explicitly instead of relying on GitHub's redirect from the old `stateful/runme`.


adambabik (approved):
LGTM

Two other places I can think of:
- Docker Hub
- Buf Schema Registry

Not sure whether they need to migrated as well.

PR #765, opened 2025-03-17
Generate es versions of proto libraries.

jlewi:
Rerun the generator.

@sourishkrout Does the generated code look right to you in terms of .js and .d.ts? I think I might be doing something wrong as I need to have two build rules in buf.gen.yaml


sourishkrout:
> @sourishkrout Does the generated code look right to you in terms of .js and .d.ts? I think I might be doing something wrong as I need to have two build rules in buf.gen.yaml

You can combine the two directives as per my latest commit. Btw, once I release/publish the latest protos tomorrow, this shouldn't be needed because the registry provides all SDKs automatically.

sourishkrout:
@jlewi could you please fix the DCO on your commits? Feel free to force-push; then I'll merge.

jlewi:
Sorry about the DCO.
Please note I added some logging in tls.go to print out the filename.

>  Btw, once I release/publish the latest protos tomorrow, this shouldn't be needed because the registry provides all SDKs automatically

Ack. I'm good with not merging this PR if you don't want to check in es versions of the protos. I'm not really blocked on it being merged. I'm copying the generated es libraries to my repo anyway and will start pulling them from the registry once they are published. So do whatever makes sense for the Runme repo.

PR #767, opened 2025-03-19
Support embedded Runme servers and the connect protocol

jlewi:
I'd like to be able to link Runme into an existing CLI. This way users don't have to download a new binary in order to be able to run runme as a server; I can just distribute as part of my existing binary server.

The blocker is that most of the Runme libraries are hidden inside the internal package which prevents other packages from importing them as a library.

This creates a thin non-hidden wrapper for the stuff I need. Arguably, refactoring the code out of the internal package is a better solution but that seems like it could be a bit of a lift.

This PR also updates `buf.gen.yaml` so we can generate connectrpc golibs. It turns out that trying to link golibraries from the buf registry and simulatenously link Runme as a library leads to name packages.

@sourishkrout I'm not really blocked on this because I can just use a rewrite declaration in go.mod to pull in my fork. So no rush.


jlewi:
There are some lint failures but I'm going to punt on resolving them until we figure out what to do about this PR.
I

sourishkrout:
> There are some lint failures but I'm going to punt on resolving them until we figure out what to do about this PR. I

That's fair. As long as this isn't blocking you.

I have yet to take a closer look and make up my mind on how to support linking.

sourishkrout:
> This PR also updates `buf.gen.yaml` so we can generate connectrpc golibs. It turns out that trying to link golibraries from the buf registry and simulatenously link Runme as a library leads to name packages.

🤔 `connectrpc` seems superfluous to me since `protobuf-es` is just as good and doesn't require a facade layer. Both `protobuf-es` and `connectrpc` are both blocked by a lack of a WebTransport grpc-transport (client for bidi) in the browser. I trust you have a plan.

jlewi:
So originally I needed connectrpc because I want to use the Connect GoLang Server(https://github.com/jlewi/cloud-assistant/pull/4). Which requires generating the connectrpc golang generated libraries. I was then connecting to it using grpc-web as the transport. Originally I was only planning on using server side streaming as that is what is supported.

If we are using websockets, then the Runme service no longer needs to be exposed as a GRPC/Connect service in my client in which case generating golang connectrpc clients should no longer be necessary. So we could drop that part.

I think I still want to be able to link in Runme to an existing binary which requires finding someway to relax the internal visibility restriction. 

I'm fine waiting to ensure there is sufficient value in linking in Runme before modifying code on main to support it. I'm fine with closing this PR for now or marking it as draft.

sourishkrout:
> So originally I needed connectrpc because I want to use the Connect GoLang Server([jlewi/cloud-assistant#4](https://github.com/jlewi/cloud-assistant/pull/4)). Which requires generating the connectrpc golang generated libraries. I was then connecting to it using grpc-web as the transport. Originally I was only planning on using server side streaming as that is what is supported.
> 
> If we are using websockets, then the Runme service no longer needs to be exposed as a GRPC/Connect service in my client in which case generating golang connectrpc clients should no longer be necessary. So we could drop that part.
> 
> I think I still want to be able to link in Runme to an existing binary which requires finding someway to relax the internal visibility restriction.
> 
> I'm fine waiting to ensure there is sufficient value in linking in Runme before modifying code on main to support it. I'm fine with closing this PR for now or marking it as draft.

Aye on both. I'll look into the best way to remove the `internal` obstacle.

Moving the PR into draft for now.

jlewi:
> However, it should be easy to replicate the <binary> env dump ... cmd which mimics more or less what Unix's env does
Seems reasonable. 

Note to self: We also need to figure out how to properly deal with it being different subcommands when linking Runme into unrelated binaries.

@sourishkrout now that we are using websockets; I don't think we need ExecuteOneshot (#764) . Want me to send you a revert PR?

sourishkrout:
> @sourishkrout now that we are using websockets; I don't think we need ExecuteOneshot (#764) . Want me to send you a revert PR?

No need, thank you. I'll remove it.

jlewi:
I think this is superceded by the branch
https://github.com/runmedev/runme/tree/seb/env-collector

PR #804, opened 2025-05-16
fix #763, disable color output by reading NO_COLOR

### LINKED ISSUE #763: Cannot read at all on light theme
I'm using light theme, but runme shows everything in this color, so I can't read anything at all.

Image

Baoqi:
By reading NO_COLOR, disable color output to avoid color issue in light theme. 

This solution is a quick fix, may not be a good long term solution.  

Feel free to close this PR if not suitable. 

Thanks


sourishkrout (approved):
✅ LGTM. Thank you, @Baoqi!

PR #847, opened 2025-07-09
Migrate from Block to Runme's Cell type

sourishkrout:
Todo:
- [x] Extend Cell to be on par with Block type
- [x] Port backend using Block to use Cell
- [x] Fix backend tests
- [x] Verify evals work
- [x] Port frontend
- [x] Remove obsolete Block-related types
- [x] Final pass to make sure all mentions of block are edited

Tracking frontend changes here: https://github.com/runmedev/web/pull/3.

Related to https://github.com/jlewi/cloud-assistant/issues/69.


api/proto/runme/parser/v1/parser.proto
```diff
@@ -54,18 +55,53 @@ enum CellKind {
...
+  // TextRange is the range of the cell covering in the markdown source.
   TextRange text_range = 5;
+
+  // Outputs are the outputs of the cell, e.g. stdout, stderr, or other media types.
   repeated CellOutput outputs = 6;
+
+  // ExecutionSummary is the summary of the cell execution, e.g. success/failure and execution time.
   CellExecutionSummary execution_summary = 7;
+
+  /* AGENT SPECIFIC FIELDS (porting from Block) */
```
sourishkrout, line 89: @jlewi, I've managed to get the first full test suite passing with Runme-ported protos.

Please check it out and let me know your thoughts on the modeling. The most notable difference is that I renamed `file_search_results` into `doc_results`. If we want to retain that it's the result of a file search, we could add a property to every result.

We'll probably keep the 100 offset and "settle" the sequence in a future `v2` of the parser.

jlewi:
Amazing.

Issue #873, opened 2025-07-29
Web: Custom Actions


jlewi:
We'd like vendors to be able to create Action components specific to their systems. We'd also like companies to be able to create components specific to their internal systems. 

Examples:
A Grafana component(https://github.com/grafana/grafana/issues/82586)
A component for KQL and Azure Logs.

I know there's some prior art in terms of the vscode extension handles custom executors e.g. [GCP](https://github.com/runmedev/vscode-runme/tree/main/src/extension/executors).

I think we can use [language_id](https://github.com/runmedev/runme/blob/1ca44ac46352733085672361cc3c856114b4425b/api/proto/runme/parser/v1/parser.proto#L75) in cell to indicate the content of a cell and condition the React component that gets loaded.

The Actions component loops over cells [here](https://github.com/runmedev/web/blob/c693b8c7d24434d47118e30cb00c1c3bf2405297/packages/react-components/src/components/Actions/Actions.tsx#L333) and creates an Action component for each one. 

Do we just need to inject a map from languages to component creator and then create the component registered for a particular language type?

@sourishkrout

sourishkrout:
> I think we can use [language_id](https://github.com/runmedev/runme/blob/1ca44ac46352733085672361cc3c856114b4425b/api/proto/runme/parser/v1/parser.proto#L75) in a cell to indicate the content of a cell and condition the React component that gets loaded.

While that seems the obvious choice, it's a bit of a fallacy. `language_id` is associated with the input, not the outputs; however, what's being rendered are output items, which individually could consist of different media types.

E.g. if you `curl` a text file versus a PNG image; both are `shell`, however, one is `text/plain` the other `image/png`. In the Grafana case, this could be CSV (table), an SVG (plot), or a PNG (bitmap plot).

I will refactor the existing `Action` to follow a repeatable pattern.

jlewi:
Looks like runmedev/web#14 landed a renderer registry.

PR #874, opened 2025-07-29
Make extra handlers injectable.

jlewi:
* Add a function so that extra http or proto handlers can be added to the server
* Makes it easy for services to add additional services e.g. as backends for custom components

Related to #873


sourishkrout (approved):
✅ LGTM

PR #875, opened 2025-07-29
Add proto definition for log service

zizheng-openai:


jlewi:
Proto service LGTM.

@sourishkrout Any concern with changes to the generated typescript code? There are certain places that look funky which make me wonder if there is an issue with using different versions of proto compiler then used previously.

sourishkrout:
> @sourishkrout Any concern with changes to the generated typescript code? There are certain places that look funky which make me wonder if there is an issue with using different versions of proto compiler then used previously.

Yes, there's something not quite right. What puzzles me how this could be different because buf doesn't use local protoc's.

Effective this is `make proto/generate` is available for, e.g. [outputs](https://gist.github.com/sourishkrout/328dd10bc2fdd9e37423e5b20c1a43c2#file-01k1esccqer1hbkvwrtnh8c9nr-md).

In any case, I suggest you, @jlewi, run, commit, and push a corrected version on top of this branch. Or, we merge and I'll correct it in `main` line.

zizheng-openai:
Hi @sourishkrout , I did use `make proto/generate`. Do you think it could because we are not pining version for the plugins in  [buf.gen.yaml](https://github.com/runmedev/runme/blob/main/buf.gen.yaml) ?

zizheng-openai:
Hi @sourishkrout, I did some debugging and probably know why there is a discrepancy in our generated protos. I noticed most of the discrepancies are under `api/gen/proto/es` and `api/gen/proto/ts`.

### For the discrepancy under `api/gen/proto/es`

It is true that the ES plugin has the version pinned to `buf.build/bufbuild/es:v2.6.0` in `buf.gen.yaml`.  
However, it seems like the protos in the current repo are not aligned with the `buf.lock` file in the current repo. Specifically, in the current repo at `api/gen/proto/es/buf/validate`, you will see a comment like:

> "`disabled` is a boolean flag that, when set to true, nullifies any validation"

I checked the source code of `buf.build/bufbuild/protovalidate`, and that specific comment only appears in versions **< 0.14** — for example, [protovalidate 0.13.4 source code](https://buf.build/bufbuild/protovalidate/file/v0.13.4:buf/validate/validate.proto).

This means that some protos in the repo were generated using `protovalidate` < 0.14.  
However, in the `buf.lock` file in the current repo, the `protovalidate` version is locked to `6c6e0d3c608e4549802254a2eee81bc8`.  
After checking [this commit](https://buf.build/bufbuild/protovalidate/commits/commit/6c6e0d3c608e4549802254a2eee81bc8), it is actually linked to versions 0.14.1 and 0.14.2.

So I think you might have upgraded the `protovalidate` version but forgotten to regenerate the protos.

### For the discrepancy under `api/gen/proto/ts`

I think the TS plugin’s version is not pinned in `buf.gen.yaml` (`"buf.build/community/timostamm-protobuf-ts"`), which could cause the problem.

PR #884, opened 2025-08-07
Pinned protobuf-ts version and re-generated protos

zizheng-openai:
Mentioned in #875

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -390,6 +390,7 @@ type AssistantServerConfig struct {
 	// CorsOrigins is a list of allowed origins for CORS requests
 	CorsOrigins []string `json:"corsOrigins" yaml:"corsOrigins"`
 
+	// StaticAssets is the path to the static assets to serve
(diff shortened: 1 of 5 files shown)


jlewi:
Is it intended that there are code changes here?
I think if we want to be systematic here the thing to do would be to first pin to versions that match the code that's in the repo. We can prove this by running the generator and showing there is no diff in the code.

Then if/when we want to update the versions we can update the pinned versions and regenerate the protos.

zizheng-openai:
> Is it intended that there are code changes here? I think if we want to be systematic here the thing to do would be to first pin to versions that match the code that's in the repo. We can prove this by running the generator and showing there is no diff in the code.
> 
> Then if/when we want to update the versions we can update the pinned versions and regenerate the protos.
> 
> So per [#875 (comment)](https://github.com/runmedev/runme/pull/875#issuecomment-3157070606) I think you want to pin proto-validate to the version that matches the current version of the code.

Hi @jlewi, I've updated my PR. And currently there are only version changes in this PR.

PR #914, opened 2025-09-17
Support proper URL parsing for server addresses

twz123:
This allows for using proper URLs for UNIX paths, so both forms with the leading double slash and without the leading double slash are correctly decoded.

Also add support for tcp URLs.

Drop unused `internal/socket` package. Whenever Windows pipe support is considered again, it makes sense to implement it via npipe URLs, analogous to unix URLs.


sourishkrout:
Thanks for the contribution, @twz123! Much appreciated. Please let me know what motivated these changes. I do see a general simplification. However, changing the `unix:/` notation to no longer accept the triple slashes will break a feature in the VS Code extension; as per [UDS support](https://docs.runme.dev/resources/security#unix-domain-socket-uds-support-for-enhanced-security).

twz123:
> Thanks for the contribution, @twz123! Much appreciated. Please let me know what motivated these changes. I do see a general simplification.

I'm usually stumbling upon this as I try to construct valid URLs to be passed to args/fields that are supposed to accept unix://, file:// or similar URL schemes, just to realize that those are, in fact, not treated as URLs at all :see_no_evil:

I understand that this is primarily a documentation issue. Nothing in the code base / docs claims that the address flag for UNIX sockets is a URL. However, for me, it's surprising that it isn't. For example, nobody would expect something that starts with "https://" to be anything other than a URL.

> However, changing the `unix:/` notation to no longer accept the triple slashes will break a feature in the VS Code extension; as per [UDS support](https://docs.runme.dev/resources/security#unix-domain-socket-uds-support-for-enhanced-security).

I added a test case that verifies the behavior of `SplitAddress`. Should have done that from the beginning. Helped me to come up with an even cleaner function implementation. In fact, URLs accept both notations, single slash and triple slash. The only thing that's working differently than before is is relative paths with double slashes, as everything after :// until the next slash in a URL indicates the "host" or "authority" part. This is basically what unix:///foo means: Scheme is unix, host is empty (as we're interested in the local file system), the path is /foo.

I was wondering if it made sense to have `SplitAddress` return an error if something about the raw address looks wrong, but I stuck to the original behavior to not be fail-fast return everything that doesn't look like a URL as TCP.

twz123:
An adjacent observation: The VSCode extension [treats addresses as URLs](https://github.com/runmedev/vscode-runme/blob/v3.15.4/src/extension/server/kernelServer.ts#L261), already.

sourishkrout:
> An adjacent observation: The VSCode extension [treats addresses as URLs](https://github.com/runmedev/vscode-runme/blob/v3.15.4/src/extension/server/kernelServer.ts#L261), already.

Yes, I believe that's where the original "requirement" for the URL came from. I agree, however, that it wasn't implemented properly. Thanks for contributing this implementation, which makes it crystal clear.

sourishkrout (approved):
LGTM ✅

sourishkrout:
Will fix the `go.mod` linter issue in `main`.

PR #916, opened 2025-09-18
Make tests more portable

twz123:
* Fix tests that call commands without a proper `PATH`
  Prefix all calls to executables with `/usr/bin/env`, or replace them with shell builtins. Use an absolute path when dumping the env.
* Only source cargo env if the file exists
  This allows for the tests to succeed, even if there's no such file on the local system.
* Remove the hardcoded SHELL variables
  It's usually not required to set this at all, except for rare occasions. There are systems out there which don't have a shell/bash in `/bin/bash`.
* Don't rely on /usr/local
  Use standard paths, instead.

### DIFF
--- a/command/command_unix_test.go
+++ b/command/command_unix_test.go
@@ -55,9 +55,9 @@ func TestCommand(t *testing.T) {
 		{
 			name: "ShellScript",
 			cfg: &ProgramConfig{
-				ProgramName: "bash",
+				ProgramName: "sh",
 				Source: &runnerv2.ProgramConfig_Script{
-					Script: "#!/usr/local/bin/bash\n\nset -x -e -o pipefail\n\necho -n test\n",
+					Script: "#!/usr/bin/env sh\n\nset -x -e\n\necho -n test\n",
 				},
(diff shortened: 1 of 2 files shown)


sourishkrout:
@twz123 thank you! Please see my extra commit as part of https://github.com/runmedev/runme/pull/919. We need to support both BSD and GNU utilities in the test suite.

sourishkrout (changes requested):
Pending BSD/GNU compatibility fix [here](https://github.com/runmedev/runme/pull/919) we can merge this.

docker/runme-build-env.Dockerfile
```diff
@@ -7,7 +7,6 @@ LABEL org.opencontainers.image.title="Runme build environment"
 LABEL org.opencontainers.image.description="An image to build and test runme."
 
 ENV HOME=/root
-ENV SHELL=/bin/bash
```
sourishkrout, line 10: Without this Dagger builds will fail. This might very well be an issue in the underyling container image. However, it deserves it's own PR to address this because without it will fail.

PR #1013, opened 2026-02-01
[agent] changes to support colab like notebook

jlewi:
Server changes to go with 
https://github.com/runmedev/web/pull/33

* Add a chatkit handler to handle chatkit responses
* Most of the code generated in this PR is generated proto files.
* This PR introduces the pattern of using protos to define toolcall JSON schemas and then using a proto plugin to generate the relevant JSON schemas.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,4 +65,3 @@ api/gen/proto/es/package*.json
 
 # TODO(jlewi): Do we want to ignore this
 .npmrc
-
--- a/api/gen/proto/es/agent/tools/v1/notebooks_pb.d.ts
+++ b/api/gen/proto/es/agent/tools/v1/notebooks_pb.d.ts
@@ -784,4 +784,3 @@ export declare const NotebookService: GenService<{
     output: typeof SendSlackMessageResponseSchema;
   },
 }>;
-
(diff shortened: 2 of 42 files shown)


sourishkrout:
Sorry, I've merged some other PRs into `main`. However, they shouldn't conflict and rebase/merge clean.

api/proto/agent/tools/v1/notebooks.proto
```diff
@@ -0,0 +1,181 @@
...
+  // request, you may omit this property which will start a new thread.`,
+  //
+  // text is the message that the user will see in slack.
+  //  A subset of markdown is supported: `-` for bullets, *bold*, _italic_, ~strike~, `inline code`,
+  //  ```multiline\ncode```, > blockquote, `<http://www.example.com|hyperlink>`. DO NOT tag @here or
+  //  @channel in this message.",
+  //
+  // fileIDs optionally send files generated as part of the investigation to the user in slack
+  // (often images). e.g. ["call_ABC", "call_DEF"]`,
+  rpc SendSlackMessage(SendSlackMessageRequest) returns (SendSlackMessageResponse) {}
```
sourishkrout: I agree, if it's got no use, we should probably remove it. Can do it post-merge though.

.gitignore
```diff
@@ -62,3 +62,6 @@ internal/project/testdata/git-project/**/.gitignore
 experimental/runme.*.yaml
 
 api/gen/proto/es/package*.json
+
+# TODO(jlewi): Do we want to ignore this
+.npmrc
```
sourishkrout: Seems optionally because there' shouldn't be any npm mgmt stuff in this repo. Probably won't hurt.

PR #1041, opened 2026-02-12
Improve runner launching - IAM Bindings, Disable Agent Service, Fix Generated Certificate

jlewi:
* Create a command to add IAM policy bindings

This can be used in launcher scripts to create suitable IAM bindings e.g.

```
"${RUNME_BIN}" agent add-iam-policy-binding --member="user:${USER}@acme.com" --role="role/runner.user" --config "${RUNME_CONFIG}"
"${RUNME_BIN}" agent add-iam-policy-binding --member="user:${USER}@acme.com" --role="role/parser.user" --config "${RUNME_CONFIG}"
```

* Fix the common name on the self generated certificate; it should be Runme not Cloud Assistant

* Add the option to disable the agent service when launching a runner as we don't need it.


jlewi:
We should merge 
https://github.com/runmedev/runme/pull/1042

PR #1075, opened 2026-03-16
agent/server: gate SPA serving on configured assets

jlewi:
## Summary
- only initialize SPA assets when static assets are configured (or a custom assets provider is supplied)
- gate SPA route registration on assets FS availability instead of agent presence
- fix misleading "Single page app is disabled" logging in the SPA-disabled path

## Why test stabilization was needed
After rebasing this branch onto latest `main`, this PR started running against newer server wiring that now initializes Jupyter proxy handlers during service registration.

That exposed two test issues that were not about the SPA behavior itself:
- `pkg/agent/codex/ws_handler_test.go` had a websocket race (`CallRoundTrip*`) where `bridge.Call` could execute before the websocket connection was installed.
- `pkg/agent/server/agent_service_test.go` had a brittle assertion that called full `registerServices()` from an SPA-assets-focused unit test; after rebase, that path now depends on unrelated config for Jupyter proxy initialization.

I stabilized tests by:
- waiting for bridge connection readiness in websocket round-trip tests and using channel-based goroutine error reporting
- removing the unrelated `registerServices()` assertion from the SPA-assets unit test

## Testing
- `go test ./pkg/agent/codex -run 'TestToolBridge_(CallRoundTrip|CallRoundTripBinary|IgnoresUnsupportedPayloads)' -count=80`
- `go test ./pkg/agent/server -run Test_NewServer_AgentEnabled_WithoutStaticAssets_DoesNotInitializeAssets -count=200`
- CI run `23411370618` passed all required checks (Lint, Docker, Ubuntu, Windows, Parser, DCO)

### DIFF
--- a/pkg/agent/codex/ws_handler_test.go
+++ b/pkg/agent/codex/ws_handler_test.go
@@ -2,6 +2,7 @@ package codex
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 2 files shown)


sourishkrout (approved):
Approve with one fix needed:

**Line 430 bug:** The log statement uses `"assetsFSConfigured", s.assetsFS != nil` inside the else block where `s.assetsFS` is guaranteed to be `nil`. This will always log `false`, making the key name misleading.

Suggest either:
- Remove this log (redundant with line 178's disabled message)
- Or fix to: `log.Info("Single page app is disabled")`

Everything else looks good — logic is correct, tests cover the main paths, and the SPA/assets decoupling makes sense.

PR #1076, opened 2026-03-16
Add Jupyter proxy routes and websocket auth support

jlewi:
Related to runmedev/web#63

## What this PR does
- Add Jupyter proxy routes for listing servers, managing kernels, and bridging kernel channels WebSockets.
- Accept bearer auth from either `Authorization: Bearer ...` or `?authorization=Bearer ...` at Runme ingress.
- Strip the client `authorization` query param before forwarding the channels WebSocket upstream.
- Continue using the configured Jupyter server token for upstream auth (`token=` query and `Authorization: token ...` header).

## Why the query-param fallback exists
Browser WebSocket clients cannot set custom `Authorization` headers during the handshake. The web client sends `?authorization=Bearer ...` to Runme for the initial auth check, then Runme removes that param before proxying upstream to avoid leaking the bearer token.

## Tests
- `go test ./pkg/agent/server ./pkg/agent/iam`

### DIFF
--- a/pkg/agent/server/jupyter_proxy.go
+++ b/pkg/agent/server/jupyter_proxy.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


pkg/agent/iam/oidc.go
```diff
@@ -381,8 +381,16 @@ func NewAuthMiddlewareForOIDC(oidc *OIDC) (func(http.Handler) http.Handler, erro
 			var token *jwt.Token
 			var err error
 
-			// Prefer bearer token over session cookie
+			// Prefer bearer token over session cookie.
+			// For browser WebSockets, custom Authorization headers are unavailable,
+			// so we also accept `?authorization=Bearer ...` as a fallback.
```
jlewi: For ws for Runme cell execution I think we put the authorization in the payload.

I think ?authorization=Bearer got added to support the proxying of Jupyter ws protocol
But I think we should follow the same pattern and put it in the payload.

sourishkrout (approved):
✅ Approved

Solid implementation. Clean auth flow, good test coverage, CI green.

Nice-to-haves for future:
- WebSocket bridges could respect r.Context() cancellation for graceful shutdown
- Consider connection limits/throttling on concurrent bridges per server
- Integration test with mock Jupyter server would be helpful

Ready to merge.

PR #1142, opened 2026-05-28
Wait for stream tap finalization on shutdown

sourishkrout:
## Summary
- add a waitable shutdown path for stream websocket handlers
- track multiplexer completion through tap RunEnd/Close finalization
- remove the unused run-level exit code parameter from StreamTap.RunEnd
- add coverage for ShutdownAndWait waiting and timeout behavior

## Test Plan
- go test ./pkg/agent/runme/stream
- runme run lint test

### DIFF
--- a/pkg/agent/runme/stream/handler.go
+++ b/pkg/agent/runme/stream/handler.go
@@ -38,20 +38,32 @@ type WebSocketHandler struct {
 	// preprocessor transforms initial ExecuteRequests before execution. May be nil.
 	preprocessor RequestPreprocessor
 
-	// clientGracePeriod, when > 0, overrides the default multiplexer client
-	// close grace period.
-	clientGracePeriod time.Duration
+	// clientGracePeriod overrides the default multiplexer client close grace
+	// period when set. A zero value disables the grace period.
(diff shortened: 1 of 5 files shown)


pkg/agent/runme/stream/handler.go
```diff
@@ -171,6 +172,25 @@ func (h *WebSocketHandler) Shutdown() {
 	}
 }
 
+// ShutdownAndWait cancels all active multiplexers and waits until each one
+// completes its close path, including tap RunEnd/Close delivery.
+func (h *WebSocketHandler) ShutdownAndWait(ctx context.Context) error {
```
sourishkrout: Does it make sense to have both `ShutdownAndWait` and `Shutdown`? One is a fast path without grace period?

sourishkrout: Good point. Looking at the consumers, `WebSocketHandler.Shutdown()` only appears to be used by `Server.shutdown()`, and that path already applies grace periods to the other resources.

So I don’t think we need both methods. Better to make the existing shutdown path wait for websocket/multiplexer finalization as part of graceful server shutdown.

PR #1167, opened 2026-06-16
fix: serialize multiclient stream processing

sourishkrout:
## Summary
- add a ping/pong readiness barrier before starting `TestRunmeHandler_MutliClient` execution
- only let the websocket handler that creates a multiplexer own `process()` for that run
- serialize processor startup so only one owner can create the processor
- make multiplexer close idempotent
- remove a newly-created multiplexer from `h.runs` if its initial `acceptConnection` fails

## Why
The original flake showed one client missing the first broadcast because the test started execution before every websocket server-side stream loop was ready. After adding the readiness barrier, the test exposed a deeper race: multiple websocket handlers can enter `process()` concurrently for the same run. Because the old check/set of `m.p` was not atomic, more than one handler could create/close processor state.

The fix keeps the multiplexer tied to the original request context so upstream cancellation behavior remains intact, but prevents non-creating handlers from racing to own the run. It also cleans up the just-created `h.runs` entry if the initial connection cannot be accepted.

## Verification
- `go test ./pkg/agent/runme/stream -run TestRunmeHandler_MutliClient -count=1000 -race`
- `go test ./pkg/agent/runme/stream`

### DIFF
--- a/pkg/agent/runme/stream/handler.go
+++ b/pkg/agent/runme/stream/handler.go
@@ -119,16 +119,16 @@ func (h *WebSocketHandler) Handler(w http.ResponseWriter, r *http.Request) {
 	}
 	sc := NewConnection(conn)
 
-	multiplex, err := h.handleConnection(ctx, runID, streamID, sc)
+	multiplex, created, err := h.handleConnection(ctx, runID, streamID, sc)
 	if err != nil {
 		log.Error(err, "Could not handle websocket connection")
 		_ = sc.Error("Could not handle websocket connection")
 		return
 	}
 
-	// If the processor was celling, we remove the run from the handler.
-	wait := multiplex.process()
-	if wait {
+	wait := false
+	if created {
+		wait = multiplex.process()
 		h.removeRun(ctx, runID)
 	}
 
@@ -136,7 +136,7 @@ func (h *WebSocketHandler) Handler(w http.ResponseWriter, r *http.Request) {
 }
 
 // handleConnection accepts a websocket connection as a stream into a multiplexer.
-func (h *WebSocketHandler) handleConnection(ctx context.Context, runID string, streamID string, sc *Connection) (*Multiplexer, error) {
+func (h *WebSocketHandler) handleConnection(ctx context.Context, runID string, streamID string, sc *Connection) (*Multiplexer, bool, error) {
 	log := logs.FromContextWithTrace(ctx)
 	log.Info("WebSocketHandler.handleConnection", "runID", runID, "streamID", streamID)
 
@@ -144,6 +144,7 @@ func (h *WebSocketHandler) handleConnection(ctx context.Context, runID string, s
 	defer h.mu.Unlock()
 
 	// If we already have a run, accept the connection on the existing multiplexer.
+	created := false
 	multiplex, ok := h.runs[runID]
 	if !ok {
 		var tap StreamTap
@@ -157,13 +158,17 @@ func (h *WebSocketHandler) handleConnection(ctx context.Context, runID string, s
 
 		multiplex = NewMultiplexer(ctx, runID, h.auth, h.runner, tap, h.preprocessor, options)
 		h.runs[runID] = multiplex
+		created = true
 	}
 
 	if err := multiplex.acceptConnection(streamID, sc); err != nil {
-		return nil, errors.Wrap(err, "could not accept connection")
+		if created {
+			delete(h.runs, runID)
+		}
+		return nil, false, errors.Wrap(err, "could not accept connection")
 	}
 
-	return multiplex, nil
+	return multiplex, created, nil
 }
 
 // Shutdown cancels all active multiplexers and waits until each one completes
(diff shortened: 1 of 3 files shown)


pkg/agent/runme/stream/handler.go
```diff
@@ -155,7 +155,9 @@ func (h *WebSocketHandler) handleConnection(ctx context.Context, runID string, s
 			options = &MultiplexerOptions{ClientGracePeriod: *h.clientGracePeriod}
 		}
 
-		multiplex = NewMultiplexer(ctx, runID, h.auth, h.runner, tap, h.preprocessor, options)
+		// The multiplexer outlives the HTTP handler that happened to create it.
+		// Otherwise a non-owning handler returning can cancel the whole run.
+		multiplex = NewMultiplexer(context.WithoutCancel(ctx), runID, h.auth, h.runner, tap, h.preprocessor, options)
```
sourishkrout: This seems dangerous... being able to cancel the context is a feature to allow control of response time latency upstream. Isn't this a bit heavy-handed to fix a test?

sourishkrout: Agreed. `context.WithoutCancel` was too broad here because upstream cancellation is intentional.

I revised the fix so the multiplexer still uses the original request context. Instead of detaching cancellation, `handleConnection` now returns whether it created the multiplexer, and only that creating handler calls `process()`/owns cleanup. Existing-run handlers just attach their websocket and return.

I kept the atomic processor ownership + idempotent close protections, and the stress test still passes:

- `go test ./pkg/agent/runme/stream -run TestRunmeHandler_MutliClient -count=1000 -race`
- `go test ./pkg/agent/runme/stream`

sourishkrout (commented):
## Code Review — PR #1167: serialize multiclient stream processing

The fix is well-reasoned and the layered approach is sound. The package builds and vets clean. One minor edge-case regression worth addressing.

### What's correct

- **The readiness barrier is the right mechanism.** The `Pong` is written from inside `streams.receive` (`acceptConnection` → `go receiveRequests` → `s.receive`), so receiving a pong genuinely proves that client's receive loop is registered and pumping. That directly closes the original flake (broadcast-before-ready). `streams.go:114` correctly skips runID/knownID validation for ping-only requests, so the barrier works before any execute request establishes `knownID`.
- **`created` gating is the primary, real fix.** Multiplexer creation happens under `h.mu`, so exactly one handler per `runID` sees `created=true`. Routing `process()` to that single owner eliminates the concurrent-`process()` race at its source.
- **`startInflight` correctly collapses** the previously non-atomic `getInflight()`/`setInflight()` pair into one critical section.
- **`closeOnce` is safe** — the early `!ok` return in `process()` happens *before* `defer m.close()` is registered, so the early-return path never schedules a close.

### Findings

**1. (Minor / edge case) Orphaned multiplexer if the creating handler's `acceptConnection` fails — a behavioral regression**

In `handleConnection`:

```go
multiplex = NewMultiplexer(...)
h.runs[runID] = multiplex     // added to map
created = true
...
if err := multiplex.acceptConnection(streamID, sc); err != nil {
    return nil, false, errors.Wrap(err, "could not accept connection")  // created discarded
}
```

If the creating handler's `acceptConnection` fails, the multiplexer is already in `h.runs` but we return `created=false` + error, so `process()` is never called. No future connection can recover it either: a subsequent client for the same `runID` finds the existing entry (`ok=true → created=false`) and also never calls `process()`. And because `acceptConnection` failed *before* `go startInactivityTimeout`, even the inactivity-cancel path never starts — the run is stuck in `h.runs` until `Shutdown`.

This is a regression from the old behavior, where a later connection *would* call `process()` and revive the orphaned multiplexer. Rare (requires `createStream` to fail), but worth a guard:

```go
if err := multiplex.acceptConnection(streamID, sc); err != nil {
    if created {
        delete(h.runs, runID)  // we still hold h.mu here
    }
    return nil, false, errors.Wrap(err, "could not accept connection")
}
```

**2. (Nit) `wait` is now effectively vestigial in the handler.** After the change, `process()`'s `wait` return is consumed only by the log line in `Handler` — `removeRun` is now unconditional inside `if created`. The existing TODO ("Can we get away without the wait flag?") is now answerable. Non-blocking follow-up cleanup.

**3. (Observation) `startInflight` atomicity and `closeOnce` are now partly defensive.** Given the `created` gating guarantees one `process()` call per multiplexer lifetime, they're belt-and-suspenders rather than strictly load-bearing. That's the right call for robustness against future callers — just noting the gating alone resolves the race.

### Tests

- Verification (`-count=1000 -race`) is appropriately aggressive for a flake fix.
- The barrier loop reuses `knownID` and per-socket timestamps correctly and asserts the echoed timestamp matches — good, that catches cross-talk.
- Consider a belt-and-suspenders test for the orphan path in #1 if you choose to fix it.

### Verdict

**Approve with one minor fix suggested.** The core race fix and readiness barrier are correct and well-justified. The only substantive item is the orphaned-multiplexer edge case (#1) — low-probability but a genuine regression in error handling that's a two-line fix.

sourishkrout:
Addressed the orphaned-multiplexer edge case from the review: if the creating handler fails `acceptConnection`, the just-created `h.runs[runID]` entry is deleted while `h.mu` is still held, so a later connection can create/process a fresh multiplexer.

Re-verified:

- `go test ./pkg/agent/runme/stream -run TestRunmeHandler_MutliClient -count=1000 -race`
- `go test ./pkg/agent/runme/stream`

PR #1169, opened 2026-06-17
feat: prepare runme-harbor for PyPI publishing

sourishkrout:
## Summary

- package Harbor's generated Python protobuf code inside `runme_harbor._proto`
- add package metadata, notice, README, and `runme run build-harbor` validation workflow
- add a PyPI publishing workflow for `runme-harbor-v*` tags using uv trusted publishing
- narrow the existing Go release workflow to `v*` tags so Harbor package tags do not trigger GoReleaser

## Verification

- `runme run build-harbor`
- `runme run test`

Note: the first two full `runme run test` attempts exposed local Docker/Dagger instability. After starting Docker and retrying, the full gate passed.


integrations/harbor/README.md
```diff
@@ -0,0 +1,52 @@
...
+cleanup() {
+  rm -rf "$tmp_venv"
+}
+trap cleanup EXIT
+
+python3 -m venv "$tmp_venv"
+"$tmp_venv/bin/pip" install -q dist/*.whl
+"$tmp_venv/bin/python" -c "import runme_harbor"
+"$tmp_venv/bin/runme-harbor" --help >/dev/null
+```
```
sourishkrout, line 42: this is a bit heavy-handed... can we just use some sort of dist folder instead of making a temp release?

sourishkrout: Done in 03d03f28. The build block now uses the wheel from dist directly via uv instead of creating a temp venv.

PR #1170, opened 2026-06-17
[codex] Sync Harbor job metadata from trials

sourishkrout:
## Summary

Sync local Harbor job config metadata after `runme-harbor run` completes so stock `harbor view` shows the observed agents and models from completed trial results.

The sync preserves the original planned `config.json` as `config.original.json` before the first update, skips jobs with no readable trial results, and can be disabled during eval runs with `RUNME_HARBOR_SKIP_METADATA_SYNC=1`.

This also adds `runme-harbor sync-metadata --jobs-dir ...` so metadata sync can run independently of a trial run. When Harbor trial metadata has no model info, sync falls back to ATIF `agent/trajectory.json` for the observed model name.

## Validation

- `uv run ruff check .` in `integrations/harbor`
- `uv run pytest` in `integrations/harbor`
- `runme run lint test`

### DIFF
--- a/integrations/harbor/src/runme_harbor/cli.py
+++ b/integrations/harbor/src/runme_harbor/cli.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import argparse
 import importlib
 import importlib.metadata
+import os
 import re
 import shutil
 import subprocess
@@ -17,6 +18,7 @@ CLAUDE_IMPORT_PATH = "runme_harbor.runme_agents:RunmeClaudeCode"
 OPENCLAW_IMPORT_PATH = "runme_harbor.runme_agents:RunmeOpenClaw"
 MIN_HARBOR_VERSION = (0, 13, 1)
 MAX_HARBOR_VERSION = (0, 14, 0)
+SKIP_METADATA_SYNC_ENV = "RUNME_HARBOR_SKIP_METADATA_SYNC"
 AGENT_ARGUMENTS = {
     "oracle": ("--agent", "oracle"),
     "codex": ("--agent-import-path", CODEX_IMPORT_PATH),
@@ -43,7 +45,15 @@ def run(args: argparse.Namespace) -> int:
     command = build_harbor_command(args)
     if args.debug:
         print(_command_string(command), file=sys.stderr)
-    return subprocess.call(command)
+    exit_code = subprocess.call(command)
+    if not _skip_metadata_sync():
+        from runme_harbor.metadata_sync import sync_jobs_metadata
+
+        try:
+            sync_jobs_metadata(args.jobs_dir)
+        except Exception as exc:
+            print(f"warning: failed to sync Harbor job metadata: {exc}", file=sys.stderr)
+    return exit_code
 
 
 def build_harbor_command(args: argparse.Namespace) -> list[str]:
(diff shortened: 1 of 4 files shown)


sourishkrout:
## Code Review

### Overview
Adds post-run synchronization of Harbor job metadata so stock `harbor view` reflects the *observed* agent/model from completed trial results rather than the *planned* `config.json`:
- New `metadata_sync.py` that reads trial results, backs up `config.json` → `config.original.json` (and `result.json` → `result.original.json`), and rewrites configs with observed agents/models.
- Auto-sync after `runme-harbor run` (toggleable via `RUNME_HARBOR_SKIP_METADATA_SYNC=1`), plus a standalone `runme-harbor sync-metadata --jobs-dir ...` subcommand.
- ATIF `agent/trajectory.json` fallback for model name when trial `model_info` is missing.
- Comprehensive test coverage.

### Code quality 👍
- **Idempotency** — `_same_json`/`_same_trial_result_json` compare parsed models (not raw text), so reruns don't churn files or clobber the original backup. Tests confirm a second sync returns `0`.
- **Lazy imports** in `cli.py` correctly defer the heavy `harbor.models` imports until after preflight. No real circular import despite `metadata_sync` importing from `cli`.
- Backup-before-mutate and "skip jobs with no readable trials" are sensibly conservative.

### Issues & risks

**1. Potential `TypeError` when sorting `ObservedAgent` (correctness, edge case)**
`ObservedAgent` is `@dataclass(frozen=True, order=True)` over `(name, model_name)` where `model_name: str | None`, and `_observed_agents` does `sorted(observed)`. If the *same* agent name appears across trials with both `model_name=None` and `model_name="x"` (e.g. one trial has model info / a trajectory and another has neither), the order comparison falls through to `None < "x"` → `TypeError`. In `run` this is swallowed into a warning, but in the `sync-metadata` subcommand it would escape `main` (which only catches `SystemExit`) and dump a traceback.
*Suggest a None-safe sort key, e.g. `sorted(observed, key=lambda a: (a.name, a.model_name or ""))`.*

**2. `sync-metadata` has no preflight (UX)**
`run` calls `_preflight` (friendly message if `harbor` isn't installed), but `sync_metadata` imports `metadata_sync` directly, which imports `harbor.models...` at module top level. Running `runme-harbor sync-metadata` without harbor installed yields a raw `ModuleNotFoundError` traceback instead of the usual `SystemExit` message. Consider a lightweight preflight for parity.

**3. Sync runs even on harbor failure (minor)**
In `run`, sync executes regardless of `exit_code`. Probably fine (it skips jobs without readable trials), but worth a comment documenting intent, or gating on `exit_code == 0` if partial syncs are undesirable.

### Minor / nits
- `metadata_sync.py`: `config.model_copy(update=updates)` is split across lines for a single arg — collapse to match surrounding style.
- `_observed_agents`' trajectory fallback is now partly redundant since `_sync_trial_results` runs first and populates `model_info`, but it's harmless defensive code.

### Test coverage
Strong — covers import-path resolution, dedup+sort, missing/builtin model info, ATIF fallback and precedence, unreadable trials/trajectories, backup preservation, and the prior-name-correction migration. **Gap:** no test for the mixed `None`/`str` model_name same-agent case from issue #1, which would catch the sort `TypeError`.

### Verdict
Solid, well-tested change that's safe by construction (backups + idempotency). Recommend addressing **#1** (real crash potential, even if narrow) before merge, and consider **#2** for CLI consistency.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1174, opened 2026-06-18
fix: use bundled Harbor CLI for runme-harbor

sourishkrout:
## Summary

- re-export Harbor's CLI as `runme-harbor-harbor` from the `runme-harbor` package
- make `runme-harbor run` resolve the bundled sibling executable instead of ambient `harbor` on `PATH`
- add tests for the script metadata, strict preflight, global `harbor` rejection, and debug command output
- keep `runme run build-harbor` scoped to building `dist`
- add CI-only wheel smoke coverage that installs the wheel into isolated uv tool dirs and verifies both executables

## Test plan

- `uv run pytest tests/test_cli.py -q`
- `uv run ruff check .`
- `runme run build-harbor`
- `runme run test`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,18 +53,9 @@ jobs:
         working-directory: integrations/harbor
         run: uv run pytest
       - name: Build package
-        working-directory: integrations/harbor
-        run: |
-          set -euo pipefail
-
-          rm -rf dist
-
-          proto_src="../../api/gen/proto/python/runme/harbor/v1"
-          proto_dst="src/runme_harbor/_proto/runme/harbor/v1"
-          mkdir -p "$proto_dst"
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -52,6 +52,31 @@ jobs:
       - name: Run tests
         working-directory: integrations/harbor
         run: uv run pytest
+      - name: Build package
+        working-directory: integrations/harbor
+        run: |
```
sourishkrout: why not do `runme run build-harbor` or use the runme GHA like we do in other workflows already.

sourishkrout: Fixed in `d35d5087`: the Harbor CI job now uses `stateful/runme-action@v2` with `workflows: build-harbor` instead of duplicating the build shell. The wheel install smoke remains CI-only after that build step.

PR #1194, opened 2026-06-23
fix: serialize codex bridge websocket connects

sourishkrout:
## Summary

- reserve the Codex bridge websocket singleton slot before upgrade/auth completes
- use connection sequence tokens so superseded in-flight handlers cannot install stale connections
- make the second-connection test deterministic and add concurrent/force-replace coverage

## Why

The Ubuntu CI flake in job `83013277329` failed in `TestToolBridge_RejectSecondConnection` because the test could dial a second websocket before the first handler had installed `bridge.conn`.

That exposed a real server-side race: two non-force websocket requests could both pass the `b.conn == nil` check before either one installed its connection.

## Test Plan

- `go test ./pkg/agent/codex -run 'TestToolBridge'`
- `go test ./pkg/agent/codex -run 'TestToolBridge_(RejectSecondConnection|RejectsConcurrentSecondConnection|ForceReplaceConnection|ForceReplaceSupersedesConnectingConnection)$' -count=50`
- `runme run test`

### DIFF
--- a/pkg/agent/codex/ws_handler.go
+++ b/pkg/agent/codex/ws_handler.go
@@ -46,10 +46,11 @@ type ToolBridge struct {
 	upgrader websocket.Upgrader
 	auth     *iam.AuthContext
 
-	mu      sync.Mutex
-	conn    *websocket.Conn
-	connSeq uint64
-	pending map[string]pendingBridgeCall
+	mu         sync.Mutex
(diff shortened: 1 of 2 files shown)


sourishkrout:
@jlewi please review.

jlewi:
LGTM

PR #1197, opened 2026-06-23
feat: add eval view command

sourishkrout:
## Summary

- add `runme eval view [jobs-dir]` for launching Harbor eval job dashboards
- delegate directly to bundled `runme-harbor-harbor view --jobs` with Harbor's `8080-8180` port range
- open the local dashboard by default, with `--no-open` and `--port` escape hatches
- preflight `runme-harbor` before eval dataset validation so missing optional dependency errors are consistent

## Tests

- `go test ./cmd -run 'TestEval'`
- `go test ./internal/harbor -run 'TestRunEvalView|TestFirstOpenLocalPort|TestRunEvalMissingHarborBeforeMissingDefaultDataset|TestRunEvalMissing|TestRunEvalDelegates'`
- `go test ./cmd`
- `go test ./internal/harbor`
- `runme run test`

### DIFF
--- a/internal/harbor/eval_view.go
+++ b/internal/harbor/eval_view.go
@@ -1,0 +1,265 @@
+package harbor
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"time"
+
+	"github.com/pkg/browser"
+)
+
+const (
+	defaultEvalViewPort      = 8080
+	defaultEvalViewPortLimit = 8180
+)
+
+type EvalViewOptions struct {
+	JobsDir        string
+	Port           int
+	Open           bool
+	RunmeHarborBin string
+	Debug          bool
+	CommandStart   CommandStartFunc
+	BrowserOpen    BrowserOpenFunc
+	DashboardReady DashboardReadyFunc
+	LookPath       func(string) (string, error)
+	Stdin          io.Reader
+	Stdout         io.Writer
+	Stderr         io.Writer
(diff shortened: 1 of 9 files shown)


internal/harbor/eval_view.go
```diff
@@ -0,0 +1,259 @@
...
+	opts EvalViewOptions
+}
+
+type evalViewPaths struct {
+	jobsDir         string
+	delegateJobsDir string
+	workingDir      string
+}
+
+func NewEvalViewRunner(opts EvalViewOptions) *EvalViewRunner {
```
sourishkrout: I don't love that it's called `EvalViewRunner`. We do have `runners` in Runme which are user facing. Why not call it `EvalViewer`?

sourishkrout: Renamed the new view-side types to `EvalViewer` / `NewEvalViewer` and updated the cmd test seam to match.

internal/harbor/eval_view.go
```diff
@@ -0,0 +1,280 @@
...
+		args = []string{url}
+	case "windows":
+		name = "rundll32"
+		args = []string{"url.dll,FileProtocolHandler", url}
+	default:
+		name = "xdg-open"
+		args = []string{url}
+	}
+	return exec.Command(name, args...).Start()
+}
```
sourishkrout: is this true to DRY in the code base? we don't have this yet or should refactor existing code for a shared utility?

sourishkrout: Good catch. `runme open` delegates browser opening to `code-server --open`, but Runme already has `github.com/pkg/browser` via auth. Swapped the local OS-specific opener to `browser.OpenURL` and removed the duplicate switch.

PR #1211, opened 2026-06-30
feat: gate eval promote with compare

sourishkrout:
## Summary

- Refactors eval job loading around shared `evalJobRef` / `evalJobStore` primitives used by both compare and promote.
- Moves comparison behavior onto the comparison model, including shared result rendering and promotion gate evaluation.
- Makes `runme eval promote` transparently compare the selected job against the tracked HEAD baseline.
- Blocks real promotion on compare gate failures unless `--promote-anyway` is passed.
- Updates `--dry-run` to show comparison/gate context and clearly label the proposed commit message.

## Behavior

- `runme eval promote --dry-run` exits successfully and reports whether real promotion would be blocked.
- `runme eval promote` blocks on regressions, no matching eval results, metadata mismatches, summary ambiguity, and job-level counter regressions.
- `runme eval promote --promote-anyway` bypasses the gate explicitly.
- `--evidence-only` uses the same gate as regular promotion.

## Tests

- `go test ./internal/harbor -run 'EvalPromote|EvalCompare'`
- `go test ./cmd -run Eval`
- `runme run test`

### DIFF
--- a/internal/harbor/eval_compare_render.go
+++ b/internal/harbor/eval_compare_render.go
@@ -71,9 +71,70 @@ type evalComparisonDiff struct {
 	Delta     interface{} `json:"delta,omitempty"`
 }
 
+type promoteCompareGateReason string
+
+const (
+	promoteCompareGateReasonRegressed promoteCompareGateReason = "regressed"
+	promoteCompareGateReasonNoResults promoteCompareGateReason = "no_results"
+	promoteCompareGateReasonMetadata  promoteCompareGateReason = "metadata"
+	promoteCompareGateReasonSummary   promoteCompareGateReason = "summary"
+)
+
+type promoteCompareGateReasonText struct {
+	blockMessage   string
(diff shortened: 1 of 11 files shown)


internal/harbor/eval_compare_render.go
```diff
@@ -80,28 +92,76 @@ func buildEvalComparison(base, candidate compareJob, baseRef string) evalCompari
...
+
+func (c evalComparison) recommendation() string {
+	gate := c.Gate()
+	if !gate.Blocking {
+		if sameComparisonJob(c.Base, c.Candidate) {
+			return "base and latest are the same eval job; nothing to compare."
+		}
+		return "candidate improved or held steady; promotion looks reasonable after normal review."
+	}
+	switch {
```
sourishkrout: this seems brittle. don't we produce the `Reason` and could we use a enum or fixed errors?

PR #1273, opened 2026-07-28
test: harden autoconfig server readiness

sourishkrout:
## Summary

- harden autoconfig server/client tests with per-test loopback addresses instead of the shared default `localhost:7998`
- keep config injection in the existing `fstest.MapFS` path
- replace the short fixed health retry loop with a 10s bounded exponential backoff readiness probe with jitter
- make TLS enabled/disabled explicit in the corresponding subtests

## Tests

```bash
go test ./internal/config/autoconfig -run '^TestInvokeForCommand_ServerClient$' -count=10
go test ./internal/config/autoconfig -run '^TestInvokeForCommand_ServerClient$' -count=50
go test ./internal/config/autoconfig
runme run test
```

### DIFF
--- a/internal/config/autoconfig/autoconfig_test.go
+++ b/internal/config/autoconfig/autoconfig_test.go
@@ -3,6 +3,8 @@ package autoconfig
 import (
 	"context"
 	"fmt"
+	"math/rand/v2"
+	"net"
 	"os"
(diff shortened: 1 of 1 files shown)


jtfacebokk-eng (approved):
Thank u

PR #1278, opened 2026-07-30
Restore static web app serving

sourishkrout:
## Summary

Restore `assistantServer.staticAssets` as an optional, supported way to host arbitrary static web applications from the agent server.

The server now:

- mounts a webapp only when the configured directory and root `index.html` are valid
- serves existing assets unchanged and falls back to `index.html` for client-side routes
- allows static hosting to be the server's only enabled capability
- keeps API, WebSocket, auth, health, and metrics routes ahead of the catch-all SPA handler
- applies configured static-asset CORS origins without credentials
- continues running without static routes when a configured asset directory is invalid

This deliberately does not restore the asset downloader, embedded assets, legacy `InitialConfigState` injection, provider API, or `webAppURL` redirects.

## Why

Static asset hosting was removed together with the legacy AI integration, but it is independently useful for serving arbitrary webapps and deployment-specific configuration files. The serving capability should remain optional and decoupled from the removed AI stack.

## User impact

Existing configurations using `assistantServer.staticAssets` become functional again and no longer produce a deprecation notice. Webapps can be hosted independently or directly by `runme agent serve`.

## Validation

- `go test ./pkg/agent/server ./pkg/agent/config`
- `runme run check test`


jlewi (approved):
LGTM

